PR #1, opened 2015-03-22
Better installation instructions.

dritter:
Hi there!

First off: Nice theme! I was about to write my own fork of the agnoster-theme. But yours is more sophisticated. So i'll try to push my thoughts in your theme. ;)

oh-my-zsh provides a way for 3rd party stuff: You can place it in `~/.oh-my-zsh/custom`. So you don't need to symlink your theme in the official themes-folder.


bhilburn:
Ah, excellent. Thanks for cleaning this up!

PR #3, opened 2015-03-22
Added mercurial support

dritter:
Now this theme also works with mercurial! :)

Only thing is:
I had to implement a very poor dirty check. Unfortunately I didn't find a better one that works with mercurial and git.. Now I just check, whether the main VCS_INFO-prompt contains a '●' or '✚'..

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -25,6 +25,8 @@ CURRENT_BG='NONE'
 # bizarre characters below, your fonts are not correctly installed.
 LEFT_SEGMENT_SEPARATOR=''
 RIGHT_SEGMENT_SEPARATOR=''
+VCS_UNSTAGED_ICON='●'
+VCS_STAGED_ICON='✚'
 
 ################################################################
 # vcs_info settings for git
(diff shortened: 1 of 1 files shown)


bhilburn:
Thanks for the pull-request! Support for other VCSes would be awesome.

As you said, I'm not a huge fan of checking status by the existence of that character. I'm going to dig around and see if I can sort out another way to do this. If not, I'll merge this changeset.

bhilburn:
Okay, I took a shot at doing this differently, and in the process changed the `git` printout somewhat. Have a look at my `mercurial_support` branch (bhilburn/powerlevel9k@3d4a1eba7da) and give it a shot. I don't really use Hg much, so I'm honestly not sure it's behaving correctly.

Feel free to send me feedback or pull-requests on that branch. If it looks good, I'll merge into master.

dritter:
What bothers me most in this approach is that we build a functionality for an information that the vcs_info-subsystem has already gathered. For me that was the point to use the existance of the staged/unstaged character, to transport the information from the vcs_info to our needs.

Secondly we have to implement that functionality for every vcs we want to support. So there is a lot of clutter in the template.

Vcs_info supports staged/unstaged only for git, mercurial and bazaar (as the zshcontrib manpage says).

I'll search for another solution at the weekend.

bhilburn:
Ah, I see what you are saying. You had changed the `prompt_vcs` function to simply detect staged / unstaged, and create the prompt status from that, rather than detecting a dirty local in that function and creating the prompt based on that. Yes, I agree that the method I implemented is not ideal, either. The other problem, as you mentioned, is that this is only supported for git, Hg, and bzr. What do we do about SVN or CVS?

That said, the more I think about it, the more I like your original solution to this more than the one I provided. If we can't come up with something else, I'll probably just merge this original pull request.

dritter:
Okay. I found a slightly better solution. Basically it is the same idea, but it uses a vcs_info-hook to check if one of `${hook_com[staged]}` or `${hook_com[unstaged]}` is empty. If not, the variable `VCS_WORKDIR_DIRTY` is set to true..

dritter:
And regarding the CVS/SVN (or: all which does not support staged/unstaged): I think it is possible to implement such a feature for them, but it might not be a good idea as we need to make an HTTP-call to query the server...
But maybe it is worth the thing (at least for SVN, CVS is IMHO pretty dead). A quick search found this (well in japanese, but google-translate for the rescue): http://www.techscore.com/blog/2013/08/02/zsh%E3%81%AEvcs_info%E3%81%AEhook%E3%82%92%E5%88%A9%E7%94%A8%E3%81%97%E3%81%A6svn-st%E3%82%92%E8%A1%A8%E7%A4%BA%E3%81%99%E3%82%8B/

For now, at least the theme won't be broken. The just see always a green background..

bhilburn:
Awesome! I like this method much better. Merged into master `=)`

Thanks for your awesome improvements, @dritter!

dritter:
You are welcome :) 

I just love this theme. Thanks for creating it!

PR #5, opened 2015-03-26
Remove the leading whitespace from the left prompt.

natemccurdy:
There is a leading whitespace on the left prompt. Seems like it shouldn't be there, so this fixes that.
Maybe this is how you like it though? Personally, it annoys me :)
image


bhilburn:
Nope! I actually fixed this already in a separate branch, but hadn't merged it yet. I'll accept your commit, though =)

PR #6, opened 2015-03-29
Display Changeset

dritter:
... And the next one ;)

I'd like to display the current changeset-ID in the prompt. Because this is IMHO quite controversial I've added a new switch `POWERLEVEL9K_SHOW_CHANGESET`, which must be enabled.

As an extra, `use-simple` is enabled, if we don't show the changeset. This boosts performance, at least when mercurial is used (as the zsh-documentation says, it affects `bzr` and `hg` only).

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,11 @@ To specify which segments you want, just add the following variables to your
     POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(context dir rbenv vcs)
     POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(status history time)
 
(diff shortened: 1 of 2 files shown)


bhilburn:
Awesome!

Merged into master. I did make one fix: changing the ($bool) to [$bool] in the conditional for `POWERLEVEL9K_SHOW_CHANGESET`.

Also, just so you know, none of your commits are being credited to your github account in the logs. I think you must be committing with an e-mail address that isn't associated with your github account? You can add multiple e-mails to your github account, if you care.

dritter:
Ha! Thanks for the hint! Just too much e-mail addresses.. :/

PR #10, opened 2015-03-29
Prompt on newline

dritter:
Added an option to set prompt on a new line.

This is just a port of https://github.com/robbyrussell/oh-my-zsh/pull/2984


bhilburn:
Nice work! Excellent customization option.

PR #11, opened 2015-03-29
A bit simpler installation instructions.

dritter:
The `themes`-folder will automatically created, when invoking `git clone .../powerlevel9k themes/powerlevel9k`..


bhilburn:
Great idea!

PR #14, opened 2015-03-30
Wording: VCS_INFO is a feature of ZSH itself..

dritter:
Can't stop harassing you.. :D

Here is just a little wording-fix: VCS_INFO is a feature of ZSH itself, not of oh-my-zsh.. ;)


bhilburn:
Hah, perfectly valid. Thanks =)

PR #19, opened 2015-04-02
Show a ratio of test classes vs real classes (Ruby and Symfony2)

dritter:
Here we show a little ratio of test vs real classes.
This is a loose port of https://gist.github.com/sent-hil/4641437 . I rather do symfony2 than ruby at work, so I extracted a little base method to show the actual prompt. Now the code is quite lean.
It may has some performance issues on large projects, but for most cases it will do nice.

Best regards
:)

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ currently available are:
 * **dir** - Your current working directory.
 * **history** - The command number for the current line.
 * **rbenv** - Ruby environment information (if one is active).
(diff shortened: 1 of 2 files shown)


bhilburn:
Are the docs accurate? Do they need to use `rspec` or `rspec_stats` to get that prompt?

dritter:
You are right! Just fixed it. ;)

Also improved the way how Classes and Tests are counted.

PR #21, opened 2015-04-04
Mercurial: show bookmark

dritter:
Here is a little port of https://github.com/robbyrussell/oh-my-zsh/pull/3365 . It shows the current active mercurial bookmark if any.

Also this patch contains a nice learning: If you want to override a substring of `hook_com[misc]` just set `ret=1` (from http://www.zsh.org/mla/workers/2010/msg00163.html).


bhilburn:
Awesome! Love the icon selection, too.

PR #28, opened 2015-05-20
Trim unwanted whitespaces.

dritter:
Long time no see ;)

On a colleagues machine we had some issues regarding whitespaces. I am not sure what the rootcause is, but trimming the whitespaces does the job.


bhilburn:
Good to hear from you, and thanks for the fix!

PR #30, opened 2015-05-24
To display the Segments correctly, we need a terminal capable of displaying 256 colors

dritter:
I had some Issues getting the right colors on my linux machine. After a while I figured out, that I was using `TERM=xterm`, wich can only display 8 colors. After changing that to `xterm-256colors`, the template worked as expected.
I am not 100% sure about this fix, as you can have `xterm-256colors` not installed because you use some other terminal editor (UXTerm, etc.). But it should do it in most cases. Unfortunately I didn't find a reliable way of determine whether xterm is installed or not..
What do you think?

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -43,6 +43,10 @@
 #   POWERLEVEL9K_COLOR_SCHEME='light'
 ################################################################
 
+# Fix: To display the correct colors, we need a terminal capable of displaying 
+# 256 colors.
+TERM=xterm-256color
+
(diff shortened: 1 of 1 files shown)


dritter:
Uhm. Maybe it's better to rely on [this plugin](https://github.com/chrissicool/zsh-256color)...
I'm still unsure. Think it's better to keep the Template clean from OMZ-Plugins..

bhilburn:
Yeah, I don't think we should automatically define that in the environment, as it will impact other terminal applications (ViM, emacs, etc.,), and may muck with users' applications. This is probably something we should instead make note of in the README.

PR #32, opened 2015-06-02
Modified Theme to support awesome-powerline-fonts as well as non-powerline-fonts

dritter:
By this pull request, you don't have to use a powerline-patched font, as long as the font can render a couple of glyphs like \uE0B0 (left segment separator) and \uE0B2 (right segment separator). It also has now support for [awesome-powerline](https://github.com/gabrielelana/awesome-terminal-fonts/tree/patching-strategy/patched)-fonts with a lot of new icons. Furthermore I introduced a new "flat" mode, without the segment separators.

To change the mode, set `POWERLEVEL9K_MODE` to
* `flat`: Renders without segment separators, but with `awesome` icons (awesome-powerline-font necessary)
* `compatible`: Renders without powerline-Icons. Still looks quite good
* `awesome-patched`: Renders with nice icons (awesome-powerline-font necessary)
* All other values (or not specifying it) default to the current status (powerline-font necessary)

The documentation is still untouched, because I have no good idea how to restructure it, to reflect this feature appropriately.


dritter:
That's how it looks like with `awesome-patched` mode.
bildschirmfoto 2015-06-03 um 12 39 37

dritter:
This is how it looks like with Helvetica as Terminal font (and 0 horizontal spacing in iTerm2) and `POWERLEVEL9K_MODE='compatible'`:
bildschirmfoto 2015-06-04 um 11 37 06

Well, not the prettiest, but hey! it renders the Segment separators..

Flat-Mode looks like this:
bildschirmfoto 2015-06-04 um 11 39 53

bhilburn:
I started merging this work in this branch: https://github.com/bhilburn/powerlevel9k/tree/dritter/font_mode

Couple of items:

1. I'm concerned about the theme code re-setting the symbols every single time a prompt is rendered. Is there a way we can make it execute that code only once? It seems like there may be other items that could make use of this same optimization.
2. I moved around some stuff in the README. I'm not 100% convinced I'm done with it, but the ToC is now inaccurate.

dritter:
The ToC is up to date again.

Hmm.. I don't think that setting these variables hits the performance very much. Other things are much more expensive. `vcs_info` has imho the biggest impact on performance. And every other filesystem-intense operation.
ZSH has built-in caching, but only for completion, afaik (http://zshwiki.org/home/examples/compsys/general). At least all examples are for completion...

What do you think about this PR in general? 
I'll add some current screenshots later..

bhilburn:
I actually like this PR a lot. I hadn't seen the Awesome-Patched fonts before, and they are pretty slick.

I definitely want to get this merged - I'm just trying to keep the theme code as clean & quick as possible in the process `=)`

If we can't find a way to only set those symbol names once, I'll merge the PR as-is.

dritter:
Glad to hear that. :)

Here are the Screenshots.
`awesome-patched`:
bildschirmfoto 2015-06-06 um 19 00 33

`powerline-patched`:
bildschirmfoto 2015-06-06 um 19 01 16

`compatible` with font `Menlo`
bildschirmfoto 2015-06-06 um 19 02 56

`flat`
bildschirmfoto 2015-06-06 um 19 04 40

dritter:
@bhilburn I think the variables are already cached. If I put an `echo "INITIALIZING"` at the top of the template and open a new shell, "INITIALIZING" gets printed only once..

bhilburn:
Closing this as it has been continued in #35

PR #35, opened 2015-06-07
Latest changes

dritter:
Here are my latest changes on the `font_mode`-branch.


bhilburn:
Merged as-is with the exception of disabling the VCS_BRAND_INDICATOR by default.

PR #38, opened 2015-06-11
Prezto & antigen support

### LINKED ISSUE #36: Prompt is empty - please help
I'm having trouble getting powerlevel9k to work for me, this is the output:
screen shot 2015-06-08 at 12 57 16 am

My .zshrc is:
```zsh
#!/usr/bin/env zsh

source ~/.dotfiles/non-packaged-repos/antigen/antigen.zsh

antigen use oh-my-zsh

export DEFAULT_USER=simonweil
POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(context dir vcs status)
POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(rbenv virtualenv history time)
antigen theme bhilburn/powerlevel9k powerlevel9k

# Tell antigen that you're done.
antigen apply
```

When I use `antigen theme funky` all is fine.

As I'm new to zsh I have no real idea what to do...

dritter:
- Added Support for antigen
- Added Support for prezto
- Added a virtual machine for test purposes

This should resolve #36


bhilburn:
Wow! Large changeset with what looks like some really amazing features. It'll take me a little bit to review. I'm traveling from this evening through the weekend, so I should be able to get to this early next week.

Thanks for the continuous stream of awesome new features, @dritter!

dritter:
@bhilburn Jep, it's quite big.. But no rush, from Saturday on I am three weeks on vacation. ;)
I hope it all still works like a charm.

dritter:
Hmm.. For me, it seems like prezto/antigen and OMZ interpret `local` variables differently. This must be an option in ZSH itself, that is set differently in that frameworks, but I can't figure it out..
antigen and prezto trip over top-level definitions of `local` variables, where OMZ has no problem at all.
It would be nice, if we could find that switch, because I think setting the variables in the theme to `local` is nicer, as they should never leave the scope of the theme itself.

bh:
:+1:

bhilburn:
@dritter - Working on merging this now. My primary concern is the loss of `local`. It seems like there must be a way to get this to work. Wondering if `local -x` might do the trick, or if there is something similar.

bhilburn:
Just gave the vagrant VM a test, and it fails during the Oh-My-Zsh install with:

```
==> default: Cloning Oh My Zsh...
==> default: Cloning into '/home/vagrant-omz/.oh-my-zsh'...
==> default: Looking for an existing zsh config...
==> default: Using the Oh My Zsh template file and adding it to ~/.zshrc
==> default: Copying your current PATH and adding it to the end of ~/.zshrc for you.
==> default: Time to change your default shell to zsh!
==> default: Password: 
==> default: chsh: PAM authentication failed
```

Looks like it's a common error with ZSH installs:
https://github.com/robbyrussell/oh-my-zsh/issues/1224

The test VM will be super useful for development - it's really a great idea. Thanks for getting this rolling!

When I do get this all merged, I'll want to break up the antigen & prezto support and the Vagrant support into separate merges. Since all of the Vagrant work is in a completely separate directory, this shouldn't be a problem.

dritter:
@bhilburn to answer your questions quickly:

1. Uhm. Not shure, if `local -x` would really fix that issue. As [the zsh docs](http://zsh.sourceforge.net/Guide/zshguide03.html) say, it just exports the variable to the environment, and unsets it, when the function exits. I am unsure about when - in context of themes - the function exits. Reads to me, as the variable is unsetted when the theme ends (and that's the end of the zsh lifecycle anyway). 

2. I added more than just this line. I analyzed some antigen themes to get a quick migration path. With the limited time, that's the reason why these lines are in there.

        setopt LOCAL_OPTIONS
        unsetopt XTRACE KSH_ARRAYS
        prompt_opts=(cr percent subst)

    As said, they come from several other themes/tutorials (e.g. [migrate-from-oh-my-zsh-to-prezto](http://jeromedalbert.com/migrate-from-oh-my-zsh-to-prezto/) [9-my-new-zsh-prompt](http://www.paradox.io/posts/9-my-new-zsh-prompt)). Unfortunately none of these explain this lines. I adopted them anyway, because I think that the theme should activate all needed options, to avoid any hassle (same counts for autoloading colors and zsh-hooks).

3. I did the migration to quoted vars in the odyssey for porting to antigen. I thought at that point, that the problems with antigen came from misinterpreted variables. Now I think this was unnecessary.

Great idea to split this pull request into two!

Regarding the OMZ-Error in vagrant: I'll have an eye on that in a week.

bhilburn:
1. Hm, okay. As you said in your commit message, it's really not ideal, but if we don't have another way around it for now then I suppose we'll just have to deal with it. Bummer.

2. Yeah, I saw those as well. The `setopt` and `unsetopt`s make sense to me, but the `prompt_opts` seemed out of place. That said, looking around at other prezto themes, I agree it seems global (e.g., https://github.com/davidjrice/prezto_powerline/blob/8e81152340c4beb2d941340d1feb2dc29bbcc309/prompt_powerline_setup). 

3. More details on how ZSH handles this relative to other shells. Based on this, I don't see any downsides to the migration: http://unix.stackexchange.com/a/171347/42642

I'll take care of the git foo later today to merge the prezto and antigen support, and I'll push a new dev branch including the Vagrant dev VM.

bhilburn:
Prezto and Antigen support has been merged. Thanks for the awesome changes, @dritter! Can you do me a favor and double-check that everything is working as expected?

Please continue development on Vagrant VM off of the bhilburn/vagrant-vm branch going forward =)

dritter:
@bhilburn awesome! Sorry for the messy pull request.

I'll double check the changes in a week. Last week without a Computer here. ;)

apjanke:
Hi folks!

The difference between OMZ and Prezto/Antigen and the interpretation of `local` variables is that OMZ loads themes by just `source`-ing the theme files from the base script, and Prezto/Antigen are sourcing it from inside a function. In `zsh`, the `local` statement is for dynamic scoping, and applies to the current function call. If you do a `local` and you are not inside any function call, it's degenerate and does nothing, and the variables end up being "global" anyway. That's what's happening in OMZ.

(Side note: if you do a `local FOO` at the top level without doing an assignment like `local FOO="something"`, then it will actually output the current value of the variable. People hack around this by doing `local FOO &> /dev/null`, so if you see that weird construction, that's what's going on.)

Because you need these variables to persist between calls to the prompt functions, they shouldn't be `local` anyway. (Zsh locals are dynamically, not lexically, scoped.) Their presence meant that your code could not be called correctly from within a function call, which was probably not what you intended.

OMZ may change in the future to load its themes from inside a function call as well. (In fact, PR robbyrussell/oh-my-zsh#3743 will do so, if it gets approved.) So removing the `local`s like this is good hygiene and future-proofing.

dritter:
@apjanke Thanks for clearing that up!

Glad, that removing the `local` is the way to go. :)

bhilburn:
@apjanke - Wow, thanks for the fantastic write-up! This was tremendously helpful.

Thanks for sharing =)

PR #40, opened 2015-06-29
updated info for antigen install option

dsager:
Thanks for getting the antigen support in! On a first try everything seems to work fine btw.

I just saw that the repository in the README example was not correct... I also mentioned (as it took me some time to figure out) that you need to define customizations before calling `antigen theme`.


bhilburn:
Thanks so much for verifying the antigen support, and for the quick fix! Very much appreciated `=)`

PR #43, opened 2015-07-07
Correctly set VCS dirty status on first invocation.

martinpelikan:
When using [last-working-dir](https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/last-working-dir) the problem becomes obvious. When first opening the terminal the prompt suggests the project is not dirty. Refreshing the prompt (simply hitting enter) updates the status accordingly. 

I took a stab at fixing the problem, and put an echo statement in [powerlevel9k.zsh-theme#L401](https://github.com/bhilburn/powerlevel9k/blob/558d137bfaf3b64c5e1b6b1091b0c066cb5f5cce/powerlevel9k.zsh-theme#L401), to see if it was getting called. The echo seemed to get compounded (0 calls on first prompt, 1 on second prompt, 2 on third prompt and so forth). It looks like on each invocation, an additional call was being added to the precmd via the hook. This might cause problems for zsh sessions spanning many commands, but I haven't experienced a problem like that. 

I would rate my zsh scripting experience as somewhere between -1 and 0, so I'm likely both wrong and doing this inefficiently, but here it is anyway.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -221,7 +221,8 @@ zstyle ':vcs_info:*' actionformats " %b %F{red}| %a%f"
 zstyle ':vcs_info:*' stagedstr " %F{$VCS_FOREGROUND_COLOR}$VCS_STAGED_ICON%f"
 zstyle ':vcs_info:*' unstagedstr " %F{$VCS_FOREGROUND_COLOR}$VCS_UNSTAGED_ICON%f"
 
-zstyle ':vcs_info:git*+set-message:*' hooks git-untracked git-aheadbehind git-stash git-remotebranch git-tagname
+zstyle ':vcs_info:git*+set-message:*' hooks vcs-detect-changes git-untracked git-aheadbehind git-stash git-remotebranch git-tagname
+zstyle ':vcs_info:hg*+set-message:*' hooks vcs-detect-changes
 
 # For Hg, only show the branch name
 zstyle ':vcs_info:hg*:*' branchformat "$VCS_BRANCH_ICON%b"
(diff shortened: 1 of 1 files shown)


bhilburn:
@martinpelikan - I was able to reproduce this issue.

@dritter - Your understanding of the problem is correct. When the prompt is first drawn, it shows the directory as clean. As soon as you press enter, though, and force a redraw, it properly shows that it is dirty. The problem only arises when you open a new tty session in `last-working-dir`.

@dritter - Any ideas on how we might address this while also not breaking mercurial? I wonder if `last-working-dir` is setting some global var that is hosing us.

martinpelikan:
@bhilburn You can also reproduce the issue by adding a 'cd <path_to_repo>' to your zshrc. I don't think it's related to the last-working-dir plugin, but more to do when (and how) that hook is added. Again, I'm a complete novice when it comes to shell/zsh scripting, but it seems odd to me that the hooked command gets run one additional time, each time the prompt is drawn...

dritter:
I think the best way to go is to add a hook for mercurial as well. @martinpelikan just add the following line underneath the git-hook registration:
`zstyle ':vcs_info:hg*+set-message:*' hooks vcs-detect-changes`

That should do the trick.

bhilburn:
@martinpelikan - Give @dritter's suggestion a shot. If it works for you, update your pull-request and let us know, and we can validate.

martinpelikan:
@dritter Please let me know if that works for you. I installed Mercurial to test it out, but I wasn't able to get it to show dirty status before or after the changes, so I suspect I'm missing something in my configuration.

dritter:
@martinpelikan works like a charm! :)

bhilburn:
@dritter Thanks for verifying. I will review, validate, and merge ASAP.

PR #44, opened 2015-07-07
Fixed oh-my-zsh test installation.

dritter:
Here we go! This fixes the `chsh`-error.

There is still one error during the installation of OMZ, but it works anyway:

    ==> default: sh: 52: /home/vagrant-omz/.zshrc:
    ==> default: Syntax error: "(" unexpected

Just log in with `vagrant ssh`, and you'll see a nice shell :).


dritter:
Are you testing this feature now? 
When can I expect that in master? ;) This PR aimed at your `vagrant-vm` branch..

bhilburn:
Now in `master` =)

PR #46, opened 2015-07-07
Make branch icon configurable

dritter:
Now the branch icon is configurable. If you want to display it, simply set `POWERLEVEL9K_SHOW_BRANCH_ICON` to `true`.

Defaults to false.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,13 +278,17 @@ By default, the `vcs` segment will provide quite a bit of information. If you
 would also like for it to display the current hash / changeset, simply define
 `POWERLEVEL9K_SHOW_CHANGESET` in your `~/.zshrc`. If activated, it will show
 the first 12 characters of the changeset id. To change the amount of characters,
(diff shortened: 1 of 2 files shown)


bhilburn:
I know you <3 the branch icon and have wanted to do this for ages, hah. Thanks for doing it in a flexible way `=)`

martinpelikan:
@dritter @bhilburn 
Just curious: is this the intended appearance with this enabled in powerline font (default) mode?

2015-07-08-200404_360x34_scrot

bhilburn:
@martinpelikan - The docs are currently wrong. I'm trying to address it in #47. If you have any thoughts / input, feel free to add them to that bug! `=)`

dritter:
@bhilburn Yeah. :) It took me some time to understand why you didn't want that icon.. :D 
I use the theme with activated commit hash, so for me it looked ugly without the icon (e.g. `772c9amaster`). You on the other hand, use it without the hash..

PR #48, opened 2015-07-09
Add a little developers guide

dritter:
This is just a little developers guide, as the theme grew quite big..


bhilburn:
Awesome! Thank you so much for adding documentation! I really, really appreciate it.

I would like to keep the README focused on end-users as much as possible, so let's put docs for developers in the Github wiki. I started it with your docs, here: https://github.com/bhilburn/powerlevel9k/wiki/Developers'-Guide

PR #53, opened 2015-07-11
Added node prompt

dritter:
Added a prompt that displays the node version.


bhilburn:
Fantastic! Thanks for another excellent contribution, @dritter!

PR #54, opened 2015-07-12
Added option to shorten dir path

dritter:
With this option you can shorten the output of the `dir` segment by specifying 

    POWERLEVEL9K_SHORTEN_DIR_LENGTH=2


bhilburn:
I love this feature! Really fantastic idea.

dritter:
:) Good to hear.
I am not 100% content with it, but it works with plain zsh.

PR #57, opened 2015-07-19
minimize dependencies

dritter:
`tr` shouldn't be necessary. `wc` output is imho without spaces.. Otherwise we should truncate the string by the means of zsh.


bhilburn:
Unfortunately it looks like this PR caused #60. Debugging with @natemccurdy, and we will revert to using `tr` for now.

PR #59, opened 2015-07-19
Multiline icons customizable

dritter:
Now the multiline-prompt is customizable. I see some folks forking the theme, just to set another multiline prompt icon. With this PR these forks become redundant. :)


bhilburn:
Awesome addition! And thanks for keeping an eye on the features people are trying to add themselves so that we can handle it in mainline =)

Issue #60, opened 2015-07-19
Question mark doesn't go away in git repository


natemccurdy:
Just updated to the latest version (c853be57e7efa79f9818394e3aec7900a6a35516) from a fairly old one, and now I can't seem to get rid of the question mark in the git `vcs` segment.

Here's a screenshot:
<img width="689" alt="screen shot 2015-07-19 at 4 06 03 pm" src="https://cloud.githubusercontent.com/assets/10200384/8768323/5bbda16e-2e30-11e5-85e1-4a866110721a.png">

natemccurdy:
The line that causes the problem is at https://github.com/bhilburn/powerlevel9k/blob/10c5b28859d372f480c1352d4b7a74672e77b430/powerlevel9k.zsh-theme#L183

A proposed fix is in PR #61 

The problem is that `wc -l` returns a string with whitespace characters in it:

     '       0'

This string is then compared to  the number `0` using `[[ $string != 0 ]]`.
That comparison will always return true unless the whitespace is stripped out somehow. In this case , by using `tr -d ' '`.

On a related note: What's with the pipe to `sed q` ? What does that do?

natemccurdy:
Fixed in commit 7f83bac947893175ef937e26f1a3bc8a389d1718 by reverting to using `tr -d '  '`

bhilburn:
@natemccurdy - Thank you so much for reporting this, finding the root cause, and submitting a pull request for it. This was the first time (that I'm aware of) that we broke `master`, which is a big deal, and so I super appreciate your time tracking this down and documenting your findings.

Strangely, the problem didn't exist for me - I tested it out before merging #57, and I imagine @dritter did, as well. I wonder if there is a difference in how `wc` behaves on different systems regarding whitespace in the output.

Regarding your question around the purpose of `sed -q`, after looking at it... I honestly don't know. I adapted the git hooks originally from @tupton, and didn't modify that. Looking through the docs, honestly, I can't figure out the reasoning for it. I might play with removing it and see what happens.

natemccurdy:
@bhilburn Happy to help! This is, by far, my favorite zsh prompt.

Strange that it worked for you. In a ZSH session, what does `true | wc -l` return for you?
Here's my output:
<img width="402" alt="screen shot 2015-07-19 at 6 28 26 pm" src="https://cloud.githubusercontent.com/assets/10200384/8768884/fb3e140e-2e43-11e5-99df-504c7e9efb65.png">

[...]

PR #61, opened 2015-07-19
Fix issue #60: Don't show as dirty in clean git repo

natemccurdy:
After commmit 10c5b28859d372f480c1352d4b7a74672e77b430, the git vcs segment always showed a question mark even if the repo wasn't dirty. This fixes that.

`wc -l` returns `\t0` which when compared to `0` in
`[[ "\t0" != 0 ]]` will always return true. This bug was causing the git repo to always show as dirty.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -337,7 +337,7 @@ prompt_vcs() {
 
 function +vi-git-untracked() {
     if [[ $(git rev-parse --is-inside-work-tree 2> /dev/null) == 'true' && \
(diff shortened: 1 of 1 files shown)


bhilburn:
@natemccurdy - Thanks so much for submitting this! To keep the git log clean, I used 'git revert' on the offending commit that you tracked down. Can you pull the latest `master` and confirm that it works as expected?

natemccurdy:
@bhilburn It works :+1: 

I'll close this PR and the issue I made.

bhilburn:
@natemccurdy - Awesome! Thanks so much for testing this out and confirming!

PR #62, opened 2015-07-20
Strip the whitespace of OSX `wc` with pure zsh. No need to install `coreutils`.

### LINKED ISSUE #60: Question mark doesn't go away in git repository

dritter:
At first, I havn't installed `coreutils`. It took me a while to figure out that this was the necessary package to install to get this Theme running. Without `tr`, there is no need to install `coreutils`.

This is the second approach to fix #60 .

@natemccurdy Can you verify my fix? 
Thx. :)

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -337,7 +337,7 @@ prompt_vcs() {
 
 function +vi-git-untracked() {
     if [[ $(git rev-parse --is-inside-work-tree 2> /dev/null) == 'true' && \
-            $(git ls-files --others --exclude-standard | sed q | wc -l | tr -d ' ') != 0 ]]; then
+            ${$(git ls-files --others --exclude-standard | sed q | wc -l)// /} != 0 ]]; then
         hook_com[unstaged]+=" %F{$VCS_FOREGROUND_COLOR}$VCS_UNTRACKED_ICON%f"
     fi
 }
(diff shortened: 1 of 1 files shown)


natemccurdy:
@dritter That fixed it! :+1:

bhilburn:
@dritter - We don't need the white-space stripping in `function +vi-git-aheadbehind()`, too?

dritter:
@bhilburn we strip the whitespace a line later.. But that one is just for optical reasons - no real functionality behind it. I put that line a while ago, when a colleague of mine complained about unwanted whitespace..

natemccurdy:
@bhilburn The variables `$behind` and `$ahead` are stripped of whitespace [here](https://github.com/bhilburn/powerlevel9k/blob/7f83bac947893175ef937e26f1a3bc8a389d1718/powerlevel9k.zsh-theme#L354) and [here](https://github.com/bhilburn/powerlevel9k/blob/7f83bac947893175ef937e26f1a3bc8a389d1718/powerlevel9k.zsh-theme#L359) using ZSH's `${variable// /}` syntax

Also, the `(( ))` test ignores whitespace before a number, so that works as intended.

bhilburn:
@dritter @natemccurdy - Ah, yes, I see, now. Thanks for pointing that out!

I'll test it out soon to confirm, and then merge. Thanks, guys!

bhilburn:
Awesome! Thanks, @dritter and @natemccurdy!

Issue #56, opened 2015-07-17
Theme breaks, if locale is "C"


dritter:
Recently I stumbled upon https://github.com/robbyrussell/oh-my-zsh/pull/4091 which affects powerlevel9k too. I started fixing this problem in my branch [dritter/character_protection](https://github.com/dritter/powerlevel9k/tree/dritter/character_protection), but it is harder than I thought..

The right prompt seems to complicate the problem. Maybe you have some good Ideas to fix that. ;)

Steps to reproduce:
Set `export LANG=C` in your `~/.zshrc`, open a new shell and voilá..

dritter:
Interestingly special characters seem to need to be wrapped in `%{ %}` to avoid unwanted spaces. This is just the case, if there is a right side prompt, which is why I set up a [gist](https://gist.github.com/dritter/ae0f0ebb2926f184838c) of the original agnoster theme with a time-segment on the right side (theme is based on PR 4091 in OMZ; RPROMPT based on powerlevel9k).

dritter:
The pure prompt-string looks quite good:

```
setopt xtrace
+omz_termsupport_precmd:1> [[ '' == true ]]
+omz_termsupport_precmd:5> title '%15<..<%~%<<' '%n@%m: %~'
+title:1> [[ '' == *term* ]]
+title:5> : '%n@%m: %~'
+title:7> [[ xterm-256color == screen* ]]
+title:9> [[ xterm-256color == xterm* ]]
+title:10> print -Pn '\e]2;%n@%m:\ %~\a'
+title:11> print -Pn '\e]1;%15\<..\<%~%\<\<\a'
+omz_termsupport_cwd:3> [[ iTerm.app == Apple_Terminal ]]
+vcs_info:1> emulate -L zsh
+vcs_info:2> setopt extendedglob NO_warn_create_global
+vcs_info:4> [[ -r . ]]
+vcs_info:6> local pat
+vcs_info:7> local -i found retval
+vcs_info:8> local -a enabled disabled dps
+vcs_info:9> local -x usercontext vcs rrn quiltmode LC_MESSAGES
+vcs_info:10> local -ix maxexports
+vcs_info:11> local -ax msgs
+vcs_info:12> local -Ax vcs_comm hook_com backend_misc user_data
+vcs_info:14> LC_MESSAGES=C
+vcs_info:15> [[ -n '' ]]
+vcs_info:20> vcs=-init-
+vcs_info:20> rrn=-all-
[...]

PR #63, opened 2015-07-24
OS Icon-prompt

dritter:
Added an icon that gives a little hint about the current operating system. It may sound stupid, but nowadays it is quite easy to work on different OSes in parallel (e.g. with a VM or via SSH).

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ like with a normal installation and default settings:
 - [Segment Customization](#segment-customization)
   - [The AWS Profile Segment](#the-aws-profile-segment)
   - [The 'context' Segment](#the-context-segment)
+  - [The 'dir' segment](#the-dir-segment)
   - [The 'time' segment](#the-time-segment)
   - [Unit Test Ratios](#unit-test-ratios)
   - [The 'vcs' Segment](#the-vcs-segment)
@@ -260,7 +261,7 @@ elements (it is by default), and define a `DEFAULT_USER` in your `~/.zshrc`:
 
     export DEFAULT_USER=<your username>
 
-#### The `dir` segment
+#### The 'dir' segment
 
 The `dir` segment shows the current working directory. You can limit the output
 to a certain length:
@@ -427,7 +428,7 @@ Before creating this theme, I also tried [jeremyFreeAgent's theme](https://githu
 #### Developing
 
 Documentation for developers is kept on the [Powerlevel9k Github
-wiki](https://github.com/bhilburn/powerlevel9k/wiki/Developers'-Guide).
+wiki](https://github.com/bhilburn/powerlevel9k/wiki/Developer's-Guide).
 
 #### Contributions / Bugs / Contact
 
--- a/README.md
+++ b/README.md
@@ -44,7 +44,8 @@ like with a normal installation and default settings:
 - [Styling](#styling)
   - [Double-Lined Prompt](#double-lined-prompt)
   - [Light Color Theme](#light-color-theme)
-  - [Further color customizations](#further-color-customizations)
+  - [Segment Color Customization](#segment-color-customization)
+  - [Special Segment Colors](#special-segment-colors)
 - [Troubleshooting](#troubleshooting)
   - [Gaps Between Segments](#gaps-between-segments)
 - [Meta](#meta)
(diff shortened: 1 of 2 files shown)


bhilburn:
Hey @dritter - I love this idea! Great work.

I've checked out the code and tested it , and have a few questions before calling it final:

1. What font do you have that provides the Linux icon? I'm using Powerline Deja Vu, and I don't have that Unicode character. How common is it?
2. Right now, the OS detection is done once (I think?), but we then go through the conditional string parsing to set the appropriate icon with every single prompt draw. Since the OS will never change out from under us, why not just set the icon once rather than parsing strings for each prompt draw? Thoughts?

dritter:
@bhilburn Good questions. :)

1. I am using `Droid Sans Mono Awesome`. The Icons should be quite common, as (except for the apple-icon) they are in the public unicode space:
[Penguin-Icon](http://unicode-search.net/unicode-namesearch.pl?term=1F427&.submit=Search)
[Sun-Icon](http://unicode-search.net/unicode-namesearch.pl?term=1F31E&.submit=Search)
[Beastie-Icon](http://unicode-search.net/unicode-namesearch.pl?term=1F608&.submit=Search)
[Apple-Icon](http://unicode-search.net/unicode-namesearch.pl?term=F8FF&.submit=Search), which is in private space, but should work on every OSX.

    Interestingly, as I retested it, I had some Issues with `echo $'\u1F427'`.. Instead echoing the character directly worked as expected: `echo 🐧`.

2. Yep, only once. Indeed, it may be better to figure out the right icon when the OS is detected. I just wanted to keep that together. But now I moved it..

dritter:
@bhilburn I found it! It seems there is a big difference whether you print unicode with an uppercase `\U`, or a lowercase `\u`. I always thought that doesn't matter, but it does. In my last commit, I changed all appearances from lowercase to uppercase..

bhilburn:
Hm, yeah, those characters don't exist in my system font. They don't even render on the referenced websites you gave, actually.

Also, `echo $'\u1F427' $'\U1F427'` shows gibberish for both: `ὂ7 🐧`

I think these glyphs are only in the `Awesome` version of Powerline fonts. 

Here's my suggestion: Let's treat these like we do for all of the other symbols. If someone has the 'awesome' fonts turned on, then great, let's use these glyphs. For non-awesome fonts, let's provide a fall-back symbol / character / string so that this segment can still be used.

Thoughts?

dritter:
Excellent idea! I've just done it so!

The difference between `\u` and `\U` still bothers me, but I think the uppercase one seems more stable. Could you double check that?

bhilburn:
Hey! I made some changes and pushed them to to a branch [dritter-os-icon](https://github.com/bhilburn/powerlevel9k/tree/dritter-os-icon).

The code was still evaluating a case statement to set an OS icon with every call to the prompt segment function for `os_icon`. I moved that statement out, combined it with the OS detection case statement, and stuck it at the top of the file. My goal is to evaluate as little as possible with each prompt draw to keep things fast `=)`

Two other minor changes:

1. You had OS detection for 'Dragonfly' and 'OpenBSD', but no icon for those OSes, so I removed them.
2. I changed the foreground color to something a bit brighter - on my screen, the previous color blended into the BG a bit too much.

Pull my branch, give it a shot, and let me know what your thoughts are!

dritter:
@bhilburn looks awesome!

1. For Completeness, I added them again. They are a part of the BSD family so the BSD icon should be fine..

2. Nice. looks good. :)

bhilburn:
Okay, so I was just doing some final testing on this branch before a merge, and I found an issue. In certain scenarios that I haven't quite figured out yet, I get a garbage UTF character in my VCS segment. But it doesn't always show up... example:

utf

It has to be a result of the change from `\u` to `\U`. Any thoughts, @dritter?

bhilburn:
Okay, update on my above post. Even when not using this branch, I'm now seeing garbage characters.

I just upgraded from Fedora20 to Fedora22, and I suspect something is hosed. Working on sorting it out, now.

dritter:
@bhilburn You could check my `character_protection`-Branch (see #56 ) even that this branch needs an update, there might be a fix for your problem..

dritter:
@bhilburn I think the biggest suspect is the uppercase `\U`. Could you change that back to `\u` and see if the error persists? `sed -i "s/\\\U/\\\u/g" powerlevel9k.zsh-theme`

bhilburn:
This work has been merged into `next` for testing.

https://github.com/bhilburn/powerlevel9k/tree/next

PR #64, opened 2015-07-24
Implemented different ways to truncate  the directory path.

dritter:
There are some caveats with this pull request. First, I use `sed` with the parameter `-E` which activates extended regular expressions under OSX's native sed. GNU's sed uses the parameter `-r`, but `-E` seems to work here as well - although this feature is not documented...
Second, the new default mode makes the directory longer if 2* `POWERLEVEL9K_SHORTEN_DIR_LENGTH` is as long as the directory.

This is the previous behavior, now only if `POWERLEVEL9K_SHORTEN_STRATEGY="complete_directories"` is set:
bildschirmfoto 2015-07-25 um 01 49 49

This is the behavior, if `POWERLEVEL9K_SHORTEN_STRATEGY=truncate_from_right` is set:
bildschirmfoto 2015-07-25 um 01 50 10

The new default behavior. Here the directory `123456` is pumped up to `123..456`, see edge case description above.
bildschirmfoto 2015-07-25 um 01 50 26

In all cases my `POWERLEVEL9K_SHORTEN_DIR_LENGTH` was set to 3.


bhilburn:
I think this is the first feature I'm honestly unsure about. It seems like a directory name should either be displayed or not be displayed, but not truncated or cut.

Also, it makes me *really* nervous to use an un-documented flag of `sed`, because we aren't guaranteed portability.

Is this particular functionality something you yourself want, something you know others want, or are you just trying to provide a more flexible configuration in general? I'm concerned we are over-shooting on this one, and in the process introducing unnecessary risk in the codebase.

dritter:
Okay. I'll do it more stable, once the `os_icon`-branch is merged (need the OS-Detection).
Unfortunately, `sed` in OSX dosn't provide very much Information.. Not even a version. So I'll have to check on the path and the current OS (and maybe the non-existance of a version string) to be sure that we are dealing with OSX-sed.

All of them. I want it myself, as I am not content with the current shorten dir-functionality, and I found the `truncate_from_right` one in the [cobalt2](https://github.com/wesbos/Cobalt2-iterm)-theme and I want to make it more flexible in general. I want the "truncate the middle part" in special, as I like this feature in the OSX file manager very much.

bhilburn:
@dritter - Okay, roger. If you think it's useful and important, let's get it done.

Let me test out the `os_icon` stuff and get it merged so that you can continue hacking =)

dritter:
@bhilburn cool!

Btw. just found a nice read about the `-E` option in sed: http://blog.dmitryleskov.com/small-hacks/mysterious-gnu-sed-option-e/

bhilburn:
@dritter - Interesting. So does `-r` work on BSD and OSX? If so, we should use that, since it is actually documented.

dritter:
@bhilburn No. Unfortunately not. `-r` works with GNU `sed` (which is the default on Linux) only. The `sed` on OSX derives from the BSD one and understands only `-E`. :(

So, either we use `-E`, or we need a os-detection (and some other checks for GNU sed) to set the correct flag.

bhilburn:
@dritter - Feel free to continue this work, branching from the HEAD of `next`!

dritter:
@bhilburn Unfortunately I have a lot of troubles with the check of the `sed`-version, as there is no good switch to ask `sed` which one it is (GNU or BSD). The problem seems to be the inability to get the correct shell ouput.

Here is a little test script I made, to test this. If I paste the line with the command into my shell it works perfectly fine (`sed --version &>> /dev/null && echo "--version Works"`, vs `sed -xx &>> /dev/null && echo "-xx Works"`). I don't know why this is so, maybe you have an idea? 

```
#!/bin/zsh

typeset -A programs
params=('--version' '-xx')

for (( i = 1; i <= $#params; i++ )) do
  param=$params[$i]
  print "Testing param: $param"
  RESULT=$(sed $param &>> /dev/null && echo "$param Works")
  #RESULT=$(sed $param 1>/dev/null 2>&1 && echo "$param Works")
  #RESULT=$(ls $param &>> /dev/null && echo "$param Works")
  print "Result: $RESULT"
done
```

dritter:
Ha! I sorted that one out! Now I can continue working on calling the correct `sed` params.

Explanation:
As I work mainly on mac, I have fairly old versions of some basic unix commands (BSD version of `sed`, `ls` and others). To use the newer GNU versions of these programs I installed a package `coreutils` via MacPorts and modified `PATH` to find these first in my `~/.zshrc`. The problem is now that `~/.zshrc` gets read in interactive shells only, so my script always used the BSD variants of `sed` and `ls`.
Solution was to modify the `PATH` in `~/.zshenv` which is read in all shell modes.

bhilburn:
@dritter - Awesome! I'm happy to hear it!

How is development on this work going? Since it's not in a merge-ready state, yet, should we close this PR and have you just submit a new one when your branch is ready?

dritter:
@bhilburn I would favor to leave this PR open, as I am not far from a solution.

dritter:
@bhilburn Here we go! Now on OSX with a BSD-sed `-E` is used and on all other OSes we use `-r`.

bhilburn:
@dritter - Your customizable icon work has been merged! You can now rebase and finish this one out =)

dritter:
@bhilburn This one is now ready to be merged into `next`. :)

bhilburn:
Merged into `next` in bf9ef6acf!

PR #66, opened 2015-07-25
Disable rprompt possible

dritter:
Some folks want no right prompt, so here is the possibility for that. I think it would be nicer, if it would be disabled by setting an empty array as `POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS`, but it seems there is no way in zsh to distinguish between an empty array and a not set variable (at least I didn't found one)..

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ this theme focus on three primary goals:
     - [Symbols](#symbols)
 - [Styling](#styling)
   - [Double-Lined Prompt](#double-lined-prompt)
+  - [Disable Right Prompt](#disable-right-prompt)
   - [Light Color Theme](#light-color-theme)
   - [Segment Color Customization](#segment-color-customization)
   - [Special Segment Colors](#special-segment-colors)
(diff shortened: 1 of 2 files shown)


bhilburn:
I agree... adding another conditional to each prompt draw isn't ideal. Merged as-is for now, but let's try to find an alternative.

maxwell-oroark:
would there be an easy way to conditionally set this variable in my .zshrc? I want to basically deactivate the RPROMPT when the tmux pane becomes a certain width.

dritter:
@maxwell-oroark The easiest way is to tell ZSH that it should automatically hide the RPROMPT if the PROMPT or the input touches it. The setting for that is [`transientrprompt`](https://fossies.org/linux/zsh-doc/Doc/Options.html#index-TRANSIENTRPROMPT).

The manual way of disabling the RPROMPT would be something like this: You can get the terminal width by checking `$COLUMNS`, and every time you open a new TMUX pane, it starts a new shell, right? So you could check for `((COLUMNS < 100 )) && POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=()` (untested pseudo code); this would disable the RPROMPT if the window is smaller than 100 columns. This is of course a very static way of doing so and won't work in all cases. For instance if you split the window vertically, the fist window would still have it's RPROMPT and it does not react to window resize.. To work around this you'd have to trap SIGWINCH, and so on.

PR #67, opened 2015-07-29
Even less dependencies

dritter:
If `sed q | wc -l` is just for testing if the string is empty or not, we can do that with just ZSH. Trusting that the implentation of `test -n` is intelligent enough to abort if a character is found. Even that `sed q` might be super fast, the big amount of indirections seem quite slow to me.

This is an addition to #60 

I tested this with OSX and the vagrant-ubuntu.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -337,7 +337,7 @@ prompt_vcs() {
 
 function +vi-git-untracked() {
     if [[ $(git rev-parse --is-inside-work-tree 2> /dev/null) == 'true' && \
-            -n $(git ls-files --others --exclude-standard) ]]; then
+            -n $(git ls-files --others --exclude-standard | sed q) ]]; then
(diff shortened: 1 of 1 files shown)


bhilburn:
This does appear to do the trick, and I like the cleaner code! Thanks, @dritter!

dritter:
Only thing is that if the `git` command returns an empty string as first line this might fail.. But it think this is an edge case we could ignore. If anyone has issues with this solution, we still could switch back. But IMHO this is not going to happen..

bhilburn:
@dritter - Tracking whether or not something causes a breakage for people is really hard to track, and might frustrate people to the point that they stop using the theme. I would rather make sure that this won't break anything for people, and then migrate to it knowing that it's stable.

Towards that, I'm staging it into an `unstable` branch. More, shortly `=)`

bhilburn:
This work has been merged into `next` for testing.

https://github.com/bhilburn/powerlevel9k/tree/next

PR #70, opened 2015-08-07
Customizable icons

dritter:
This is a quite big change, but adds a lot of flexibility on a new level.

## What users get
Every Icon is now customizable. :) Even combinations like a "flat"-styled theme with a normal powerline font are possible now.
The `POWERLEVEL9K_MODE` is just a shortcut for a set of icons to be used.

## What we get
We get the ability to test all icons at once, by adding `icons_test` to our prompts. I hope to avoid regressions like the one in #63 by that. I was even able to delete some lines of code.

## Drawbacks
Well, I had to add a new indirection layer.. This makes the Theme more complicated and some CPU cycles slower.


dritter:
I am getting crazy.. In antigen and prezto, the default mechanism seems not to work.

```
+left_prompt_segment:21> print_icon LEFT_SEGMENT_SEPARATOR
+print_icon:1> local 'icon_name=LEFT_SEGMENT_SEPARATOR'
+print_icon:2> local 'ICON_USER_VARIABLE=POWERLEVEL9K_LEFT_SEGMENT_SEPARATOR'
+print_icon:3> local 'USER_ICON='
+print_icon:4> [[ -n '' ]]
+print_icon:7> echo -n
```
Although no user icon for the left prompt separator is set, `[[ -n '' ]]` evaluates to true.. In my world this is a pure ZSH function and has nothing to do with prezto or antigen.. :(

bhilburn:
@dritter - Any updates on things being broken in prezto and Antigen? Since we formally support those installation methods, we can't merge this until it's resolved.

Unfortunately, I'm not of much use, since I have zero experience with both of those =(

dritter:
@bhilburn This is ready now. It works in antigen, omz, prezto and plain zsh.

bhilburn:
Hey @dritter - I pulled in these changes, but it looks like the issue that was resolved in #74 resurfaced. I tried to fix it by replicating your changes that migrated some of the `\U` characters  to `\u`, but it unfortunately doesn't seem to be doing the trick.

Can you have a look? I'm working off of my branch, `dritter-icon-custom`.

dritter:
Yep. I'll have a look. Should be easy. ;)

dritter:
Note to self: Did I say easy? Hah. I couldn't even reproduce it..

@bhilburn I wasn't able to reproduce this issue, but I stumbled over the incoming/outgoing piece of code, which I forgot to convert to the `print_icon`-function. Also I changed the uppercase `\U`s to lowercase. That _should_ be the fix.
Could you double check?

bhilburn:
@dritter - Your new work looks good! Merged!

PR #72, opened 2015-08-09
Added some hints about troubleshooting missing fonts

dritter:


bhilburn:
Great! I appreciate you responding to filed bugs - both in terms of the bug itself as well as submitting patches to fix them or prevent them in the future =)

PR #77, opened 2015-08-26
Set RETVAL for left prompt also.

jkoelndorfer:
Pull request per https://github.com/bhilburn/powerlevel9k/commit/d8ed0964866bb9e187b26a88fd19404f6be39262#commitcomment-12902303. :-)


bhilburn:
Thanks for the fix, @jkoelndorfer! Highly appreciated.

PR #79, opened 2015-08-29
Added a possibility to not show a specific icon

dritter:
A couple of forks use multiline-prompts and removed the prefix of the first line (e.g. @jkoelndorfer 's fork). This is the main motivation for this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,6 +414,12 @@ This special prompt does not work on the right side, as it would be too long,
 and ZSH hides it automatically. Also have in mind, that the output depends on
 your `POWERLEVEL9K_MODE` settings.
 
+You can prevent rendering of an icon by setting the according icon variable to
+`false`. E.g. if you use a multiline prompt and do not want a special icon
+before the first line, just set:
+
+    POWERLEVEL9K_MULTILINE_FIRST_PROMPT_PREFIX=false
+
 #### Segment Color Customization
 
 For each segment in your prompt, you can specify a foreground and background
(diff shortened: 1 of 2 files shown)


jkoelndorfer:
Did some reading on checking whether variables are defined or not in shell. Found http://mywiki.wooledge.org/BashFAQ/083#Testing_that_a_variable_has_been_declared, which gives a couple options to do that:

`typeset -p varname` or `[[ ${varname+1} ]]`

A quick test shows both of these methods seem to work for detecting whether or not a variable is defined. They even work if the variable contains an empty string.

My proposal would be that one of those two methods be used, then defaults can be set and features can be easily disabled without needing to check that some other variable is true or false. We could apply this method to a bunch of existing configuration options, though that would break backwards compatibility. If folks aren't opposed, I could open a separate issue for that and maybe even put some work in on it. :-)

Thoughts?

dritter:
Ha! I was looking for something like that for quite some time, when I tried to disable the `RPROMPT` if `POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS` is not set instead of defining a new variable.

Secondly, setting icon variables to false isn't ideal either. I would love to see that solution you suggested. 

Go for it! :)

bhilburn:
@jkoelndorfer - That sounds great! Please base your work off of the HEAD of `next`, and take a crack at it =)

jkoelndorfer:
@dritter, per your work in next, plus #83, I think this change is obsolete.

@bhilburn ^

dritter:
Closed in favor of #83

PR #81, opened 2015-08-29
New segments for system stats and current IP

dritter:
I was inspired by @Illizian blogpost [here](https://www.alexscotton.com/post/a-modular-responsive-zsh-theme-and-some-bonus-unix-commands) and thought that it would be nice to have such segments as well..
The new segments worked for me in OMZ, Antigen, Prezto, plain ZSH (all vagrant) and a native Linux system.

What comes in this pull request as well is that I changed the Icons for Apple and Linux to glyphs from font-awesome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ this theme focus on four primary goals:
   - [The AWS Profile Segment](#the-aws-profile-segment)
   - [The 'context' Segment](#the-context-segment)
   - [The 'dir' segment](#the-dir-segment)
+  - [The 'ip' segment](#the-ip-segment)
   - [The 'time' segment](#the-time-segment)
   - [Unit Test Ratios](#unit-test-ratios)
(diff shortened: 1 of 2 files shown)


dritter:
Btw. here is how it looks like: 
bildschirmfoto 2015-09-01 um 19 23 15

bhilburn:
@dritter - Awesome new segments! I love them.

This PR has some conflicts, though, with the head of `next`. Also, can you update it to use the new `defined` function instead of the `-n` check?

Let me know as soon as the PR is updated and I'll proceed with the merge =)

simonweil:
Looks cool!
May be allow to chose between ram and load.
I specifically would like to see only load and also that only when it reaches some threshold.

dritter:
@bhilburn Now it is ready. :)

@simonweil Thanks. I added an option to disable the free ram. But I am not sure about the threshold. For now I left that one out..

PR #82, opened 2015-09-01
Fix format for git-svn

labianchin:
When the repository backend is like git-svn or git-hg, there was a conflict with `zstyle` `formats` from `functions/VCS_INFO_formats`. Now it applies the format to all backends starting with `git` or `svn`.


bhilburn:
I agree that this is a bug-fix and belongs on master. Merged.

Thanks so much for the patch, @labianchin! I really appreciate it!

PR #83, opened 2015-09-02
Implement defined, set_default utility functions

jkoelndorfer:
Per my previous comment, whipped up a couple of utility functions to help set defaults for powerlevel9k configurables.

`defined` does what you think it does - except unlike `-n`, it also works if the variable is defined but empty. Very nifty!

`set_default` allows for a shorthand if your variable is a scalar, e.g: `set_default myvar hello` sets `myvar` to "hello" only if it has not been defined previously.

Unfortunately, `typeset` can't set a value for an array. For arrays, do something like `defined array || array=(1 2 3)`.

Also put in some work to get existing variable checks switched over to use `defined` or `set_default` as appropriate. I didn't go into great depth to see if we could do any refactoring, just grabbed the low hanging fruit for this PR. :-)

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -241,16 +241,8 @@ else
   DEFAULT_COLOR_DARK="236"
 fi
 
-VCS_FOREGROUND_COLOR=$DEFAULT_COLOR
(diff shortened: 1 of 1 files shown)


dritter:
Awesome! Thanks. :)

bhilburn:
Fantastic contribution, @jkoelndorfer! Thanks so much!

PR #87, opened 2015-09-04
Improve same color segments

dritter:
Now a small segment separator is shown, if two subsequent segments have the same background color.
bildschirmfoto 2015-09-04 um 18 35 02


bhilburn:
Great change! I love it.

The only thing I don't understand is why you got rid of `$DEFAULT_COLOR` and replaced it universally with `white` in `prompt_longstatus`?

dritter:
To have the separator printed with high contrast I use the foreground color of the current segment. 

`longstatus` is a segment with black background and the default color (with the default "dark" theme) is black also even if it prints just icons anyway. Because of that the separator was never visible.
This is the reason why I changed it to white..

PR #91, opened 2015-09-12
Updated few icons to use with git.

tusharsrivastava:
I have updated some icons' utf8 code to look better. Old icons were not matched to many icons.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -118,9 +118,9 @@ case $POWERLEVEL9K_MODE in
     OK_ICON="\u2713" # ✓
     FAIL_ICON="\u2718" # ✘
     SYMFONY_ICON="SF"
-    VCS_UNTRACKED_ICON="\uE16C" # 
-    VCS_UNSTAGED_ICON="\uE17C" # 
-    VCS_STAGED_ICON="\uE168" # 
+    VCS_UNTRACKED_ICON="\uF059" # 
+    VCS_UNSTAGED_ICON="\uF06A" # 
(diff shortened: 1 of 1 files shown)


dritter:
Thanks for your contribution.

Do you use the "fallback-strategy" of `awesome-terminal-fonts`?
Unfortunately this produces different unicode-codepoints than the "patching-strategy" which this theme is using.. I'll try to suggest a pull request to awesome-terminal-fonts in the next days..

bhilburn:
@tusharsrivastava - Thanks so much for getting involved and submitting a contribution!

@dritter - Are you saying that this PR needs an update to work in a fallback situation?

dritter:
@bhilburn No, I think this PR is meant for the fallback strategy. 

Maybe we should have a new section in the icons-variable to support fallback strategy (`awesome-fallback`). IMHO this should be https://github.com/bhilburn/powerlevel9k/compare/master...tcbbd:master as @tcbbd made a quite complete port.

dritter:
@bhilburn Uhm. Merging this means, that we now have glyphs for fallback- and patching-strategy all together. So Icons are now broken in `awesome-patched`-mode..

It would be better to revert this merge and make a new branch in the `icons`-definition (see my previous comment).

bhilburn:
@dritter - Hm, I'm really having trouble understanding how you would like to handle this, as our icon expert =)

Can you lay out a step-by-step plan for what needs to happen?

dritter:
@bhilburn yes. Now it looks good again on `master` and `next`.

bhilburn:
@dritter - Great! Thanks for confirming!

PR #99, opened 2015-09-19
Make docs reflect difference between status and longstatus

giladgo:
Also, the default for the right side is `longstatus`.


bhilburn:
Thanks for the fix, @giladgo! I appreciate it!

Issue #107, opened 2015-09-28
Add missing icons in `awesome-fontconfig`


dritter:
@V1rgul could you add the missing Icons in my [`awesome_fontconfig_cleanup`](https://github.com/dritter/powerlevel9k/tree/awesome_fontconfig_cleanup)-branch? That would be awesome. Unfortunately neither me nor @bhilburn have installed awesome-terminal-fonts the fontconfig way..
Here is a screenshot of the `awesome-patched` configuration:
![bildschirmfoto 2015-09-28 um 20 21 13](https://cloud.githubusercontent.com/assets/1544760/10144671/187340ae-661f-11e5-944c-9a62dccf48e4.png)

bhilburn:
Fixed by #108

Issue #97, opened 2015-09-18
Icons: The Masterplan


dritter:
As you requested in #91 , here is the masterplan to get to stable icons. I'll describe it here from the beginning, it may serve as possible explanation for other users.
#### Introduction

This theme uses special icons to look fancy. So we need special fonts to display these special icons correctly. There are two ways to use these fonts: 
1. We can use pre-patched fonts. This is the preferred way for all `POWERLEVEL9K_MODE`s: default, compatible and awesome-patched.
2. Every operating system has a mechanism to fallback to the correct fonts. This is a more complicated way to install the fonts, as usually extra software is necessary.

Here I focus on the first (pre-patched) way, because even this is the easier method, some parts must be understood.
#### Installation

Three parts play a role to display the theme correctly:
1. The font itself: It must provide the correct glyphs that the theme wants to display.
2. The terminal emulator: The font must be proper installed in the operating systems font manager _and_ set in your [terminal emulator](https://en.wikipedia.org/wiki/Terminal_emulator) (such as iTerm2, Konsole, Terminal, RxVT, etc.) as font.
3. The theme: It prints special characters, representing a glyph in your font. As different fonts use (at least for these special characters) different [code points](https://en.wikipedia.org/wiki/Code_point), the theme needs to know which set of code points it should use. These sets are represented as `POWERLEVEL9K_MODE`.
   For the `default`-mode, you just need a [powerline-patched](https://github.com/powerline/fonts) font. For `awesome-patched`, you need a [awesome-patched](https://github.com/gabrielelana/awesome-terminal-fonts/tree/patching-strategy/patched) font.
   Unfortunately, there is no way to detect the used font automatically because the shell doesn't know much about how to render its content.
#### Far-future plan

As the `awesome-terminal-fonts` are not the only project which offer a broad range of pre-patched fonts and besides look somewhat inactive, I think the best would be to switch to [nerd-fonts](https://github.com/ryanoasis/nerd-fonts). They seem to be more active, have stable codepoints over different stategies, more icons. On the downside, they have some serious issues which need to be fixed first.
The other long-term solution could be to unify the codepoints in the different strategies in the `awesome-terminal-fonts` project.
**Status: Open**
#### Interim solution

[...]

PR #108, opened 2015-09-28
Added awesome fontconfig for next icons

V1rgul:
see #107


bhilburn:
Great! Thank you, @V1rgul!!

dritter:
Just for the record: This is the interim-solution from #97 

Thank you very much @V1rgul

PR #109, opened 2015-10-06
Quote value of IS_BSD_SED.

jkoelndorfer:
If this isn't quoted, in can bleed through to stdout in certain circumstances.
This one is pretty weird, haven't dug in to figure out exactly why it is bleeding through. I know that removing "local" fixes it, but quoting the value also seems to fix it.

It also only occurs if I invoke oh-my-zsh from both my zshrc AND zprofile. Super weird. /shrug


jkoelndorfer:
OK - fixed it up so that it is working on my end. Putting the 'local' declaration on its own line is enough to break things, even.

PR #111, opened 2015-10-11
Removed `spectrum_ls` from the README

dritter:
`spectrum_ls` is specific to OMZ. Replaced it by a universal one-liner.


dritter:
The information is now in the wiki, so I edited it there.

bhilburn:
Makes sense. Thanks, @dritter !

PR #113, opened 2015-10-12
Add NVM prompt segment

alexlafroscia:
Prints out the Node version that is currently active if it is different than the default version specified by NVM

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ The segments that are currently available are:
 * [ip](#ip) - Shows the current IP address.
 * **load** - Your machines 5 minute load average and free RAM.
 * **node_version** - Show the version number of the installed Node.js.
+* **nvm** - Show the version of Node that is currently active, if it differs from the version used by NVM
 * **os_icon** - Display a nice little icon, depending on your operating system.
 * **php_version** - Show the current PHP version.
 * [rbenv](#rbenv) - Ruby environment information (if one is active).
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -752,6 +752,17 @@ prompt_php_version() {
   fi
 }
 
+# Node version from NVM
+# Only prints the segment if different than the default value
+prompt_nvm() {
+  local node_version=$(nvm current)
+  local nvm_default=$(cat $NVM_DIR/alias/default)
+  [[ -z "${node_version}" ]] && return
+  [[ "$node_version" =~ "$nvm_default" ]] && return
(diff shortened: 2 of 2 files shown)


bhilburn:
@alexlafroscia - Thanks for the contribution! A couple of items:

1. We *just* released a new version of powerlevel9k, and this now has conflicts. Can you rebase and re-submit your PR for the `next` branch?
2. Can you add this segment to the README segment list, and document how it's different from `node_version`?

Thanks again for contributing this! I'm excited to get it in.

bhilburn:
Just thought of something else -

Perhaps it would be best to integrate this into the existing `node_version` segment? Is that possible?

dritter:
@bhilburn about merging segments together: This is exactly what I was thinking about the last days. 
I would love that for the `rvm` and `rbenv` segments as well.

My general thoughts about that topic:

- Some segments seem to be helpful in a "server"-like context, and some in a "project"-like one. For example, I work as a web developer and sometimes need to know about the PHP-Version on the server, but on my local machine I just need that information in development project. So if I am outside of that development project (-folder) that segment is not shown.

- Maybe we need a variable `SHOW_CONDITIONAL_$segment_name`, so users can decide if they want to show that segment in a "project"-context or always ("server"-context). In extreme we could let the user decide how that condition looks like (e.g. `find -maxdepth 2 -type f -name '*.php' | head -n 1` for a quick lookup of PHP-files).

- The [bullet-train guys](https://github.com/caiogondim/bullet-train-oh-my-zsh-theme/blob/master/bullet-train.zsh-theme) have a similar "project-" approach in some of their segments (like the `prompt_virtualenv` or `prompt_ruby`).

alexlafroscia:
Yeah, absolutely! I'll add those things and submit a new PR for that other branch instead.

I actually wrote my own custom `rbenv` segment that uses a similar technique to only show the segment if it's not the default, but at the moment it doesn't work if the default is set to `system`, so I didn't PR that one yet.

I totally agree that a flag for allowing the conditional to be only shown sometimes makes sense, and if you want to hold off on the PR until a good pattern is arrived at, that totally makes sense to me!

alexlafroscia:
Closed in favor of #114

dritter:
@alexlafroscia I have just implemented a first draft in my [improve_rvm](https://github.com/dritter/powerlevel9k/tree/dritter/improve_rvm) branch (and only for the `ruby_version` segment). As I don't use ruby, could you have a look over it? And @natemccurdy you as well? That would be nice!
What do you think? Any suggestions?

/cc @bhilburn

PR #114, opened 2015-10-13
Add nvm prompt segment

alexlafroscia:
Prints out the Node version that is currently active if it is different than the default version specified by NVM.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ The segments that are currently available are:
 * [ip](#ip) - Shows the current IP address.
 * **load** - Your machines 5 minute load average and free RAM.
 * **node_version** - Show the version number of the installed Node.js.
+* **nvm** - Show the version of Node that is currently active, if it differs from the version used by NVM
 * **os_icon** - Display a nice little icon, depending on your operating system.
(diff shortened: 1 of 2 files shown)


alexlafroscia:
At the moment, this doesn't do anything to give the user the option to show it all the time, or show it conditionally, although I'm in favor of making that a feature.  Having the information in the prompt is super useful, but doing so only when it's necessary is even better.

I wonder what the best way to provide that configuration is, though.  Ideally, I really only want the version to show up if I'm working on a Node project, and that could be discovered easily enough by checking for a `package.json` file, which would most likely be present.  I'm just not sure how that should interact with the default/non-default version setting, or if it would be better to ditch the default/non-default option all together in favor of detect a Node project instead.

bhilburn:
@alexlafroscia - Great contribution! Thanks so much!

I agree with you regarding the conditional display. I would like to find a way to do this consistently for all segments that would benefit from this. @dritter has opened an issue, #115, where we can discuss this and figure out how to do it in a smart, clean, and efficient way.

Thanks again for the PR! Happy hacking!

PR #116, opened 2015-10-16
Vcs cleanup

dritter:
Now VCS_INFO is only invoked if VCS segment is activated.


bhilburn:
Smart improvement, @dritter!

PR #118, opened 2015-10-16
Made whitespaces between segment separators customizable

dritter:
This is the second customization to integration @mustig s changes.

I didn't include documentation about this feature in the README. I think this feature is not so important and better located in the wiki.


bhilburn:
Thanks, @dritter! And agreed - the should be documented in the [Stylizing Your Prompt wiki page](https://github.com/bhilburn/powerlevel9k/wiki/Stylizing-Your-Prompt).

PR #119, opened 2015-10-17
Improve ruby version

dritter:
Now the ruby_version segment tries to get the version by asking rvm, rbenv, chruby and has an option to be displayed always.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -554,8 +554,8 @@ prompt_nvm() {
   local nvm_default=$(cat $NVM_DIR/alias/default)
   [[ -z "${node_version}" ]] && return
   [[ "$node_version" =~ "$nvm_default" ]] && return
-  NODE_ICON=$'\u2B22' # ⬢
-  $1_prompt_segment "$0" "green" "011" "${node_version:1} $NODE_ICON"
+
+  $1_prompt_segment "$0" "green" "011" "${node_version:1} $(print_icon 'NODE_ICON')"
 }
 
 # rbenv information
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -472,10 +472,27 @@ prompt_load() {
 
 # Node version
 prompt_node_version() {
-  local node_version=$(node -v 2>/dev/null)
-  [[ -z "${node_version}" ]] && return
+  defined POWERLEVEL9K_NODE_VERSION_CONDITION || POWERLEVEL9K_NODE_VERSION_CONDITION=true
+  defined POWERELVEL9K_NODE_VERSION_CHECKERS || POWERLEVEL9K_NODE_VERSION_CHECKERS=('nvm' 'node')
 
-  "$1_prompt_segment" "$0" "green" "white" "${node_version:1} $(print_icon 'NODE_ICON')"
+  typeset -Ah segment_definition
+  segment_definition=(
+    'segment'               "$0"
+    'background_color'      'green'
+    'foreground_color'      'white'
+    'position'              $1
+    'icon'                  '$(print_icon "NODE_ICON")'
+    'condition'             $POWERLEVEL9K_NODE_VERSION_CONDITION
+    'checker_nvm'           '
+        local node_version=$(nvm current 2>/dev/null)
+        local nvm_default=$(cat $NVM_DIR/alias/default 2>/dev/null)
+        [[ -z "${node_version}" ]] && return
+        [[ "$node_version" =~ "$nvm_default" ]] && return
+        echo ${node_version:1}'
+    'checker_node'          'node -v 2>/dev/null'
+  )
+
+  conditional_segment segment_definition POWERLEVEL9K_NODE_VERSION_CHECKERS
 }
 
 # print a little OS icon
@@ -548,6 +565,7 @@ prompt_ram() {
 
 # Node version from NVM
 # Only prints the segment if different than the default value
+# DEPRECATED! Use `node_version` instead!
 prompt_nvm() {
   local node_version=$(nvm current)
   local nvm_default=$(cat $NVM_DIR/alias/default)
@@ -904,6 +922,7 @@ powerlevel9k_init() {
     'longstatus'      'status'
     'rvm'             'ruby_version'
     'rbenv'           'ruby_version'
+    'nvm'             'node_version'
   )
   print_deprecation_warning deprecated_segments
 
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -789,10 +789,20 @@ prompt_vi_mode() {
 # More information on virtualenv (Python):
 # https://virtualenv.pypa.io/en/latest/
 prompt_virtualenv() {
-  local virtualenv_path="$VIRTUAL_ENV"
-  if [[ -n "$virtualenv_path" && "$VIRTUAL_ENV_DISABLE_PROMPT" != true ]]; then
-    "$1_prompt_segment" "$0" "blue" "$DEFAULT_COLOR" "($(basename "$virtualenv_path"))"
-  fi
+  defined POWERLEVEL9K_VIRTUALENV_CONDITION || POWERLEVEL9K_VIRTUALENV_CONDITION='[[ -n "$VIRTUAL_ENV" && "$VIRTUAL_ENV_DISABLE_PROMPT" != true ]]'
+  defined POWERELVEL9K_VIRTUALENV_CHECKERS || POWERLEVEL9K_VIRTUALENV_CHECKERS=('default')
+
+  typeset -Ah segment_definition
+  segment_definition=(
+    'segment'               "$0"
+    'background_color'      'blue'
+    'foreground_color'      $DEFAULT_COLOR
+    'position'              $1
+    'condition'             $POWERLEVEL9K_VIRTUALENV_CONDITION
+    'checker_default'      'basename "$VIRTUAL_ENV"'
+  )
+
+  conditional_segment segment_definition POWERLEVEL9K_VIRTUALENV_CHECKERS
 }
 
 ################################################################
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -215,6 +215,17 @@ function set_default() {
   defined "$varname" || typeset -g "$varname"="$default_value"
 }
 
+# Checks if an array contains a specific value.
+# First parameter: the searched value
+# Second parameter: array
+function in_array() {
+  local element
+  for element in "${@:2}"; do
+    [[ "$element" == "$1" ]] && return 0;
+  done
+  return 1
+}
+
 # Safety function for printing icons
 # Prints the named icon, or if that icon is undefined, the string name.
 function print_icon() {
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -777,6 +777,7 @@ prompt_nvm() {
 }
 
 # rbenv information
+# DEPRECATED! Use `ruby_version` instead!
 prompt_rbenv() {
   if [[ -n "$RBENV_VERSION" ]]; then
     "$1_prompt_segment" "$0" "red" "$DEFAULT_COLOR" "$RBENV_VERSION"
@@ -987,6 +988,12 @@ function zle-keymap-select {
   zle reset-prompt
 }
 
+function showWarningIfDeprecated() {
+  if in_array "$1" "${POWERLEVEL9K_LEFT_PROMPT_ELEMENTS[@]}" || in_array "$1" "${POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS[@]}"; then
+    print -P "%F{yellow}DEPRECATED!%f You use a deprecated segment '$1'. Please use '$2' instead!"
+  fi
+}
+
 powerlevel9k_init() {
   # Display a warning if the terminal does not support 256 colors
   local term_colors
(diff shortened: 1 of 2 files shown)


bhilburn:
@dritter - My only issue with this PR is that it will break prompts for users who are currently using `rbenv`, as that segment just suddenly disappears.

What I would like to do is:

1. Have the `rbenv` segment invoke the new `ruby_version` segment.
2. If `rbenv` is used, print, only once, a warning that it is deprecated and tell users to migrate to the new segment.

We could then make release v0.3.0 safely, and then remove `rbenv` entirely in v0.4.0.

What do you think?

dritter:
Sounds perfect! I'll do.

dritter:
@bhilburn I implemented only the second point. The first one might be bigger, because what about the customizations users made? Like special colors? Should we try to apply them as well?
I think there is no good automagical way (except if we check the invoked segment, transform the variables to the successor-segments one, etc), so it makes IMHO no difference if we leave them untouched and display a warning. Or are there good reasons to do it anyway?

dritter:
I reworked it again, so that users can control the logic when the segment gets rendered. Unfortunately this is a quite complex operation (find files named X, but search not to deep and execute that statement as callback), but I think it is a flexpoint that makes sense.

What do the ruby guys think? @natemccurdy @alexlafroscia

natemccurdy:
I'd love to check this over when I have time (which unfortunately, is not right now)... but I'll test it soon and chime in.

powerlevel9k.zsh-theme
```diff
@@ -955,6 +1007,12 @@ function zle-keymap-select {
   zle reset-prompt
 }
 
+function showWarningIfDeprecated() {
```
alexlafroscia: Just for consistency's sake, maybe this should be named `show_warning_if_deprecated` to avoid mixing camel and snake casing?

alexlafroscia:
I'll try this out later today and see how it feels.

I think for `rbenv`, and I totally might be wrong since I don't do a ton of Ruby, that `$RUBY_VERSION` isn't always necessarily set when using it.  Coming up with smart default values is important IMO and I think it'll be tough to do that.

dritter:
@natemccurdy @alexlafroscia Thanks for your feedback.

I reworked it again.

To your comments:

1. Done. Just missed that.

2. That is the problem of my approach here. I am trying to get a kind of a "universal" structure that could be applied to any conditional segment. But some tools have their own way of "just show, if there is something", like `rbenv`. So in that case, it would be a bit verbose, yes.
It might get a bit better, when I managed to use just the return value, instead of an string. In that case you could just write `POWERLEVEL9K_RUBY_VERSION_CONDITION=true`, which is still a bit verbose, but more intuitive. But I am still working on this one.

3. I moved the whitespaces to the Icon-definition. So if there is no icon, no whitespaces will be printed.

4. That was not intentional. Thanks for that finding. I changed it, so the loop exits as soon there is a version found. The `POWERLEVEL9K_RUBY_VERSION_CONDITION` is needed anyway, as we don't know if you are in a project context, if you use another checker than `rbenv`.
E.g. in a scenario where we use `ruby` directly as our version checker, we don't know when to display the version number (only how).

dritter:
Nr. 2 is now done. The condition must now evaluate to `true` or `false`. :)

bhilburn:
@dritter - What's the status of this one? Do you think it's ready for merge?

@natemccurdy @alexlafroscia - Since you guys have been testing it, and would be using it, I'd like to have your "sign off" on the merge. Thoughts on current state?

alexlafroscia:
Sorry I disappeared on this.  Just installed it and am getting the following error:

```
build_right_prompt:4: command not found: prompt_longstatus
```

dritter:
@alexlafroscia the `longstatus`-segment got merged into the `status` segment.. Maybe we need a deprecation warning here as well.

@bhilburn Not completly. I am still working on extracting a general "conditional segment" function, so that other prompts can be easily conditional as well. I am just having trouble to transfer the segment configuration array correctly to that new general method..
But feedback is always good. :)

alexlafroscia:
@dritter if there's a convention for deprecation warnings, that would definitely be helpful.  I had to dig a little bit to figure out what the problem was.

alexlafroscia:
Excuse the poor quality, but here's a GIF of my prompt now.  It's a little choppy, but demonstrates well enough the performance difference between a directory with Ruby files and without them
nov 02 2015 18 39

dritter:
@alexlafroscia Is that a performance issue because of the ruby prompt, or because of vcs?
Could you deactivate vcs for that test?

Btw. would it be a good default condition to check for the existence of a gemfile?

alexlafroscia:
@dritter Definitely seems to be the ruby prompt. Disabling VCS does make it _slightly_ faster, but it's still significantly slower inside that Rails app than outside of it.

dritter:
@alexlafroscia Can you help me track that down?

1. Can you do an `setopt xtrace` in that directory, and paste the output in a gist? That would be helpful.

2. Is the performance significantly better if you deactivate `ruby_version` segment (but with `vcs`)?

3. And how is the performance compared if you use the `master` branch with the according ruby segment?

alexlafroscia:
@dritter 

1. [Here's a gist of that output](https://gist.github.com/alexlafroscia/0c882090d2791a0b7d28)
2. Yes; I opened two windows side by side, one with `ruby_version` disabled and one with it enabled, and the terminal with it disabled performed normally while the other was just as slow as before.
3. Switching to `master` doesn't seem to affect performance at all

dritter:
@alexlafroscia Thank you very much. It seems like you use the default set of ruby-checkers (`+prompt_ruby_version:1> POWERLEVEL9K_RUBY_VERSION_CHECKERS=( rvm rbenv chruby ruby )`). In that case we test all of them one by one. That might be slow depending on how many of them you have installed and how many would print a version number.

I'll try to explain some more what is going on in that script:

If the condition to display that segment is true (`+prompt_ruby_version:2> POWERLEVEL9K_RUBY_VERSION_CONDITION='[[ -n $(find . -name "*.rb" -maxdepth 2 | head -n 1) ]]'), we try to render that segment. This default-condition tries to find files with the suffix ".rb" with a max depth of two folders.

First, we check `rvm`. In your case it looks like you have `rvm` installed, but that environment is just "default", so we don't want to print that. But that is an invocation of ruby anyway.

Second, we check `rbenv`. That one is empty in your case, so we skip it.

Third check is `chruby`. That is also empty. Skip it.

Fourth and last check is the ruby version directly from ruby itself. The first three checkers may not need a special condition on when to render this segment as the do it implicitly, but this last checker does. So here we invoke ruby again to get the version string.

I would say, to boost the performance in your case, just enable `rvm` and (if you use it) `rbenv`. A second perfomance boost would be to set the condition to `true`, because you don't need it if you don't use the `ruby` checker.

Tl;dr:
```zsh
POWERLEVEL9K_RUBY_VERSION_CONDITION=true
POWERLEVEL9K_RUBY_VERSION_CHECKERS=('rvm' 'rbenv')
```

alexlafroscia:
I see; that makes a lot of sense.  Thank you for being so thorough and explaining what's going on!

Since I don't actually use `rvm`, I uninstalled that and set

```zsh
POWERLEVEL9K_RUBY_VERSION_CHECKERS=('rbenv')
```

That made *all* the difference!  It's really fast now; you wouldn't even know that it's checking the ruby version at all!

dritter:
@alexlafroscia You are welcome! That just shows that I need to improve the documentation. :wink: 
I think it is important to give users the possibility to decide what they want. And if they don't want to decide, give them a good information at last (in that case: fallback to `ruby` version).

Yes, we just invoke `rvm`. So there is no chance we can speed it up (but I am no ruby expert)..

rjorgenson:
@dritter I used the $GEM_HOME and $MY_RUBY_HOME environment variables to display the version without invoking RVM. You may be able to use this in the new function to avoid invoking RVM for each prompt.

```
  local gemset=$(echo $GEM_HOME | awk -F'@' '{print $2}')
  [ "$gemset" != "" ] && gemset="@$gemset"
  local version=$(echo $MY_RUBY_HOME | awk -F'-' '{print $2}')
  local rvm_prompt="$version$gemset"
  [[ ! -z $rvm_prompt ]] && "$1_prompt_segment" "$0" "red" "$DEFAULT_COLOR" "$(print_icon 'RUBY_ICON')  $rvm_prompt "
```

dritter:
@rjorgenson looks good! I'll try that on friday. Thx.
Is `MY_RUBY_HOME` set by rvm? I am just curious, because I ported that segment from the famous bullettrain-zsh-theme and I wonder why they didn't do it like that in the first place.

rjorgenson:
@dritter Yeah, MY_RUBY_HOME and GEM_HOME are both set by RVM. Though I'm not sure that they aren't also set by other things, but when using RVM, those variables are updated with RVM's information. I don't know that they are exclusive to RVM(never used anything else).

dritter:
@rjorgenson Perfekt! As said, I'll update this PR on Friday. Thank you very much.

Do you you have an opinion on the default condition for this segment? Is searching for *.rb files a good thing? Or better use the existence of a gemfile?

rjorgenson:
As someone who uses RVM strictly to go back and forth between system ruby on OS X and a ruby that isn't over 2 years I don't really have an opinion =] I prefer to always see the ruby version if it's not what I have set as the default, but I'm also not a ruby dev with different projects that need different version of ruby. My only thought would be to test checking for a *.rb file in a directory with tons of files(thousands) to make sure it won't end up causing problems when browsing around a system.

dritter:
I just refactored the internal structure. There is now a utility function `conditional_segment`. All that a segment now has to do to be conditional is to set a configuration array like in the `ruby_version`:
```zsh
prompt_ruby_version() {
   defined POWERLEVEL9K_RUBY_VERSION_CHECKERS || POWERLEVEL9K_RUBY_VERSION_CHECKERS=('rvm' 'rbenv' 'chruby' 'ruby')
   defined POWERLEVEL9K_RUBY_VERSION_CONDITION || POWERLEVEL9K_RUBY_VERSION_CONDITION='[[ -n $(find . -name "*.rb" -maxdepth 2 | head -n 1) ]]'

   typeset -Ah data
   data=(
     'segment'         $0
     'color'           'red'
     'position'        $1
     'icon'            '$(print_icon "RUBY_ICON")'
     'condition'       $POWERLEVEL9K_RUBY_VERSION_CONDITION
     'checker_rvm'     'gemset=$(echo $GEM_HOME | awk -F"@" "{print \$2}")
       [ "$gemset" != "" ] && gemset="@$gemset"
       version=$(echo $MY_RUBY_HOME | awk -F"-" "{print \$2}")
       echo "$version$gemset"
     '
     'checker_chruby'  'chruby 2> /dev/null | sed -e "s/ \* //"'
     'checker_rbenv'   '$RBENV_VERSION'
     'checker_ruby'    'ruby --version 2> /dev/null | grep -oe "ruby [0-9.a-z]*" | grep -oe "[0-9.a-z]*$"'
   )

   conditional_segment data POWERLEVEL9K_RUBY_VERSION_CHECKERS
}
```
What do you think?
Disclaimer: I tested this on OSX only. There may be issues on linux. Just wanted to have something.

I am a bit indifferent about this one. On one hand it is nice that probably every could be conditional, but on the other hand there is a lot going on with eval (which is bad).

As a proof-of-concept I made the `load` segment conditional as well. Some users already requested that. E.g. you could only print that segment if the load average is over a specific threshold:

    # Would render the segment if load is over 3..
    POWERLEVEL9K_LOAD_CONDITION='[[ "$load_avg_5min" -gt 3 ]]'

alexlafroscia:
I think this is a really cool idea; there are a lot of segments that I would love to make conditional and this seems like a really cool way to make that a lot easier to manage.

wadkar:
This looks like quite a promising PR!
Is there any way I can help? Just brewed chruby, ruby-install so I kinda have fresh system.

dritter:
@wadkar Thanks. 
I am kind of stuck with this PR. 
The problem is, if you try this out, you will quickly notice that it feels quite odd to define code in variables that get eval'd later on.. Doesn't it?

I still have a tendency towards having abstract segments. Having a `ruby` segment is better than having a `chruby` etc. segment. The `chruby`, `rvm` whatsoever are just implementation details of _how_ to get the right data. All the user wants is to display the right ruby version..
But I can understand that an abstract `ruby` segment cannot read minds. So there always will be configuration to do for the user to get the right behavior out of the segment.

I still think about a small version of this idea. I still think that in some contexts the user wants to display a conditional segment always (like in a server environment). So having a variable that checks this would be a "small" implementation of this PR. But this is still not the holy grail.. I'll have to wait until the muse kisses me. :wink:

natemccurdy:
@dritter FYI regarding the `rbenv` portion of your work.....

https://github.com/bhilburn/powerlevel9k/pull/216

Using `$RBENV_VERSION` as the way to check which version of Ruby is being used is not complete. There's a better way to do it in the PR linked above.

bhilburn:
Hey all -

What's the status of this PR? Is it still necessary? If so, we need to get it cleaned up before it can be merged.

dritter:
In my opinion it is too heavyweight for the outcome. Most users won't change the conditions of a segment. So this would slow down the theme too much for them..
I'll think over a lighter solution.

wadkar:
Perhaps we could go for a lightweight `chruby` only segment? It would be up to the users who prefer `chruby` over `rbenv` (or other managers for that matter) to pick the right one.

dritter:
@wadkar This is a good idea. Would you mind implementing one?

wadkar:
Sure, see bhilburn/powerlevel9k/pull/227

bhilburn:
Thanks to @wadkar's work in #228, we now have another Ruby segment option, `chruby`. Check out the latest `HEAD` of next.

PR #123, opened 2015-10-18
Fixes #122, SI prefixes should be evaluated as strings.

### LINKED ISSUE #122: f3d2a07 introduces a syntax error?
It has been a while since my last update of this theme.  It looks like f3d2a0755ef421b22b8d5f9a8db6d66e2f774ef9 introduces a regression:
```
powerlevel9k.zsh-theme:65: parse error near `|'
```

This causes the powerline-style markup of the segments to not colorize or show special characters. This issue is present in the current version as well:
```
powerlevel9k.zsh-theme:232: parse error near `|'
```

I'll follow up if I find the issue or a fix.

martinpelikan:


dritter:
Awesome! Thanks!

I am just wondering why I didn't had this issue. Do you have an older Version of ZSH? I am running on: `zsh 5.0.8 (x86_64-apple-darwin14.4.0)`

martinpelikan:
I'm wondering the same thing actually! 
`zsh 5.1.1 (x86_64-unknown-linux-gnu)`

Does my fix actually work and do what you were intending? I didn't read the context/usage and went for the low hanging fruit, so I'm still trying to convince myself that this is what you meant to have.

martinpelikan:
The [release notes for 5.0.8 -> 5.1.0](http://zsh.sourceforge.net/releases.html) do mention something similar with local variables and arrays, but it seems this should still be valid syntax. I'm not much of a bash/zsh scripter so I have yet to come up with a good explanation for what's happening.

dritter:
Your fix works like a charm! And yes, that is exactly what I intended ;)

So I am the one on the older version. :smile:

bhilburn:
Thanks for the fix, @martinpelikan! I really appreciate it!

PR #124, opened 2015-10-18
added battery prompt segment usable on OS X

rjorgenson:
Added a new prompt segment that can show you battery status on OS X machines. Will fail gracefully on any non OS X system or an OS X system without a battery.

screen shot 2015-10-18 at 11 02 52 am


rjorgenson:
~~Stole~~ Borrowed a Linux laptop from work and added linux support to this segment, closing this PR and will submit a new one!

PR #126, opened 2015-10-18
added a prompt segment to display battery status on laptops

rjorgenson:
Added a new prompt segment that can show you battery status on. Supports both OS X and Linux. On OS X it will display the time remaining until charged or discharged(not provided by the subsystem on Linux as far as I can tell). Will fail gracefully on any system without a battery.

screen shot 2015-10-18 at 11 02 52 am

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ configuration is the default:
 The segments that are currently available are:
 
 * [aws](#aws) - The current AWS profile, if active.
+* [battery](#battery) - Current battery status(OS X only).
 * [context](#context) - Your username and host.
 * [dir](#dir) - Your current working directory.
 * **history** - The command number for the current line.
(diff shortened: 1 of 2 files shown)


rjorgenson:
I was able to find a way to give time remaining support on Linux with a dependency(the acpi program - acpi package on ubuntu), so that brings feature parity across platforms for this segment.

DanielGGordon:
This looks pretty awesome - do you need the awesome fonts for this to work?

rjorgenson:
Hi Daniel,

No you don't need it, I added the standard unicode battery icon(🔋) for the awesome-fontconfig(not sure what the awesome unicode value is for this setup) and the catchall. It doesn't look as nice IMO, but it does work =]

bhilburn:
Fantastic contribution, @rjorgenson! Thank you so much!

I'm very happy to have this in the project =)

PR #128, opened 2015-10-19
Rephrasing of error message when TERM not 256 color

DanielGGordon:
- Literal command to be added is highlighted in blue
    - Now beginners won't try to add an incorrect command to ~/.zshrc


DanielGGordon:
The way it was phrased before, it seemed like I could just add `TERM=xterm-256color` to `~/.zshrc`. however I found that you really need to put the `export` before it, and ` "" "" ` around the `TERM` variable.

bhilburn:
Nice reporting improvement, @DanielGGordon! Thanks!

PR #129, opened 2015-10-19
Ram prompt

dritter:
Separated `load` and `ram`. Now ram has its own segment. For me the valuable information about memory is free ram and used swap. The new ram segment is capable of displaying both values.


bhilburn:
Great new segment, @dritter!

Issue #133, opened 2015-11-03
No warning about change in behavior within the `virtualenv` segment


alexlafroscia:
~~I noticed while testing #119 that the `virtualenv` segment seems to be broken.  I tried it against `master` and `next`, with `master` working fine but `next` being broken.~~

Never mind: it seems that instead of being broken, the condition changed.  While it used to check to see if `$VIRTUAL_ENV_DISABLE_PROMPT` was set, it now checks to see if it is not set to `true` before rendering the segment, and only renders it if is it not true.  There doesn't seem to be any documentation about the change, so this might confuse people that upgrade (like me).

alexlafroscia:
Nevermind: it seems that instead of being broken, the condition changed.  While it used to check to see if `$VIRTUAL_ENV_DISABLE_PROMPT` was set, it now checks to see if it is not set to `true` before rendering the segment, and only renders it if is it not true.  There doesn't seem to be any documentation about the change, so this might confuse people that upgrade.

dritter:
I think we need a changelog..

Sorry for the inconvenience. :(

alexlafroscia:
A changelog would really be perfect.  Since the dev flow nowadays seems to be merging a bunch of things into `next` and then from there into `master` it would be great if that commit got tagged as a release, where you can add notes about changes in there.

bhilburn:
A CHANGELOG is now part of the project. Thanks, guys!

alexlafroscia:
:tada:

PR #134, opened 2015-11-04
Added a changelog

dritter:
@alexlafroscia is right in #133. We need a changelog. Here it is (at least a quick proposal).


bhilburn:
Agreed! Thanks, @dritter and @alexlafroscia!

Issue #131, opened 2015-10-23
History substring search


jmiguelv:
When I use this theme the the history substring search plugin does not work. Any ideas why or has anyone seen this problem before?

Thank you

bhilburn:
@jmiguelv - I actually use history-substring-search, and it works fine.

I have had history-substring-search break a number of times, and it is always the result of an update to another plugin I'm using. I've found that changing the order of the plugin loads often fixes the problem. This is what I'm using right now:

```
plugins=(zsh-syntax-highlighting gitfast git-extras svn history-substring-search colored-man colorize copyfile copydir cp dircycle command-not-found)
```

bhilburn:
It could also be your keybindings:

```
### KEY BINDINGS ############################################################
zmodload zsh/terminfo
bindkey "$terminfo[kcuu1]" history-substring-search-up
bindkey "$terminfo[kcud1]" history-substring-search-down

# Have to do this again for tmux
bindkey '^[[B' history-substring-search-down
bindkey '^[[A' history-substring-search-up

## Vim key bindings
bindkey '^?' backward-delete-char
bindkey '^R' history-incremental-search-backward
bindkey '^S' history-incremental-search-forward
bindkey '^P' history-search-backward
bindkey '^N' history-search-forward
```

jmiguelv:
Thank you for the suggestions @bhilburn. The plugin works with other oh-my-zsh themes though. I have also tried using just the `history-substring-search` plugin, but not luck. Here is my configuration:
```zsh
export ZSH=$HOME/.oh-my-zsh

ZSH_THEME="powerlevel9k/powerlevel9k"

export DISABLE_AUTO_TITLE="true"

plugins=(brew colorize command-not-found fabric git git-extras history-substring-search last-working-dir mercurial osx pip python sudo svn-fast-info vagrant zsh-syntax-highlighting)

export PATH=$HOME/bin:/usr/local/bin:/usr/local/sbin:$PATH

source $ZSH/oh-my-zsh.sh

DEFAULT_USER=jvieira

POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(context dir virtualenv vcs)
POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(status history time)
POWERLEVEL9K_STATUS_VERBOSE=false
POWERLEVEL9K_LEFT_SEGMENT_SEPARATOR=''
[...]

PR #136, opened 2015-11-11
Issue #131: This fixes history-substring-search, home and end keys.

dritter:
This PR fixes the `history-substring`-issue as well as the `home` and `end` keys. This works for me with all frameworks _but_ OMZ with `vi-mode` activated. I used the following configuration (in the vagrant, as we need to source the `history-substring-search` widget):
```zsh
source /home/vagrant-omz/.oh-my-zsh/plugins/history-substring-search/history-substring-search.zsh
zmodload zsh/terminfo
bindkey "$terminfo[kcuu1]" history-substring-search-up
bindkey "$terminfo[kcud1]" history-substring-search-down
```

Huge thanks to @apjanke.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -922,12 +922,26 @@ $(print_icon 'MULTILINE_SECOND_PROMPT_PREFIX')"
 
 function zle-line-init {
   powerlevel9k_prepare_prompts
+  if (( ${+terminfo[smkx]} )); then
+    printf '%s' ${terminfo[smkx]}
+  fi
   zle reset-prompt
+  zle -R
+}
+
+function zle-line-finish {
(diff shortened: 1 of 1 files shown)


apjanke:
Cool. You're welcome!

This should also work with OMZ too, as long as you do *not* load its `vi-mode` plugin, which is the part of OMZ that is currently having problems with this stuff. (And hopefully that will eventually get fixed too, and we'll all be consistent.)

dritter:
:+1: Right! I had accidentally `vi-mode` activated. Thanks.

bhilburn:
Just tested this and it works for me - I was even able to remove my hard-coded key presses in my `~/.zshrc`! Great job, @dritter, and thanks for your input, @apjanke!

PR #139, opened 2015-11-13
Get terminal capabilities using a zsh builtin for portability

sdupille:


bhilburn:
Thanks for the great contribution, @sdupille!

DanielGGordon:
Ah I see what I missed. Good fix.

PR #140, opened 2015-11-13
Add shorten strategy for package.json name

alexlafroscia:
I was using this shortening technique before I found `powerlevel9k`, and since I found it so useful I decided to integrate it "properly" as a shortening technique.  Basically, the idea is that if your project has a `package.json` at the root of the `git` project, it will abbreviate the entire path up to the project with just the name of the package, and then show a relative path from there.

For example, if you have a project in `~/projects/my-cool-project` with a `package.json` that looks like this:

```json
{
  "name": "My Cool Project"
}
```

and then navigate to `~/projects/my-cool-project/tests`, the directory will be shown as `My Cool Project/tests`.

It does have a dependency on [`jq`](https://stedolan.github.io/jq/), so if that's a problem I can try to find a different way to extract the name field using more commonly-installed tools.  `jq` just makes working with `JSON` files really easy.

Another thing that I've thought about doing as an improvement is simply traversing upward in the path to the current directory, searching for a `package.json` file.  This would mean that you no longer need the `.git` folder to be present to work, and you can also use multiple `package.json` files inside you project to create shortened names for particular sub-directories.  If you're interested in those changes, I'll take a look at adding them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,8 +127,26 @@ To change the way how the current working directory is truncated, just set:
     # default behaviour is to truncate whole directories
 
 In each case you have to specify the length you want to shorten the directory
-to. So in some cases `POWERLEVEL9K_SHORTEN_DIR_LENGTH` means characters, in 
+to. So in some cases `POWERLEVEL9K_SHORTEN_DIR_LENGTH` means characters, in
 others whole directories.
+
+Another option is to truncate the path such that it is shown relative to
+the current `name` property in a `package.json` file at the root of the current
+`git` project.  You can enable that with the following:
+
+    POWERLEVEL9K_SHORTEN_DIR_LENGTH=true
+    POWERLEVEL9K_SHORTEN_STRATEGY="truncate_with_package_name"
+
+For example, if you have a project in `~/projects/my-project` with a `package.json`
+that looks like this:
+
+    {
+      'name': 'My Cool Project'
+    }
+
+the prompt will display `My Cool Project`.  When you `cd` into a subdirectory,
+it will appear as `My Cool Project/subdirectory`.  Note: this requires
+[jq](https://stedolan.github.io/jq/) to be installed.
 
 ##### ip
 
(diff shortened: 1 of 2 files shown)


bhilburn:
@alexlafroscia - Awesome idea! I like this contribution a lot! 

A couple of things:

1. Can you close this PR and re-submit against the `next` branch? I only merge bug fixes into `master` with the exception of making a release, when `next` gets merged into `master`.
2. The dependency on `jq` is problematic, because while the shortening is specific to JSON, the segment this applies to, `dir` is used by lots of people working with things other than JSON. I think it's useful in it's current form, but would be *really* useful to a lot of people if you found a way to make it work without needing the `jq` interpreter. Do you have ideas for how this could be done? If it's too difficult, we can bring it in as-is, and add a note that this is only available to JSON devs.

alexlafroscia:
Yeah, absolutely on both fronts.

In reference to `jq`, I'll look into other, better ways to parse out the value that we need.  Fortunately, if this shortener isn't enabled, then `jq` isn't necessary at all, but I still agree that finding a way to use it without that being necessary would be preferable.

alexlafroscia:
Sorry I disappeared on this, I'm going through "end of semester" stuff right now,  I'll try to make this happen over Winter break.

bhilburn:
@alexlafroscia - No problem! Best of luck with your exams / projects / school stuff!

bhilburn:
@alexlafroscia - Any interest in this work, still? No pressure at all, but if you need to put it down for a while, I might close this PR for the time being =)

alexlafroscia:
Hey, that's fine. School is really heavy right now (last semester, 3 CS courses) but I'm going to make a personal note to come back to this once I have some more free OSS time. Thanks!

bhilburn:
Sounds good, @alexlafroscia! Good luck with your course work!

alexlafroscia:
I've been looking into ways to parse JSON data without `jq` and most of the answers seem to petty much be "use JQ or a a language library" (at least, that's the advice given [here](http://unix.stackexchange.com/a/121734) and I haven't found any implementation examples online).  Would using _some_ language be OK? I'm not sure if `awk` and `sed` will be enough.

PR #141, opened 2015-11-15
Huge performance improvement if you use `rvm` segment.

dritter:
As #119 is still growing, it is time to harvest some of the fruits there. Here is the performance-improvement part. This is not a pure performance improvement as now the gemset will be printed as well.

Thanks to @rjorgenson for the hint.


bhilburn:
Awesome! Thanks for keeping Goal #4 in mind (a speedy terminal is a happy terminal) while doing your development!

=)

HaoZeke:
Should I revert this and submit a PR?

PR #143, opened 2015-11-17
Added segment deprecation helper

dritter:
Now segments can be deprecated by adding them to the `deprecated_segments` array. This is kind of a backport from #119


dritter:
I flagged `longstatus` as deprecated, but using that segment produces an error, as the function is not defined any more. Maybe we want to check if that segment is still a valid function (`whence -w prompt_longstatus`) and if not, remove it from the array to avoid the error..

Issue #145, opened 2015-11-17
rprompt on same line as lprompt, not one second line


Tritlo:
Greetings,

when I set

    POWERLEVEL9K_PROMPT_ON_NEWLINE=true

the left prompt appears on the same line as the rprompt, and not the line below (as seen in the image associated with the variable).
![rpromptnotonnewline](https://cloud.githubusercontent.com/assets/1667287/11216210/4717295a-8d4a-11e5-9ad8-129905bda6f8.png)

My config is as follows. My terminal is "terminator", though this also happens on "gnome-terminal"

    POWERLEVEL9K_MODE='awesome-fontconfig'
    POWERLEVEL9K_LEFT_SEGMENT_SEPARATOR='▶'
    POWERLEVEL9K_RIGHT_SEGMENT_SEPARATOR='◀'
    POWERLEVEL9K_PROMPT_ON_NEWLINE=true
    POWERLEVEL9K_MULTILINE_FIRST_PROMPT_PREFIX=""
    POWERLEVEL9K_MULTILINE_SECOND_PROMPT_PREFIX="\e[44m\e[1m\e[33m $ \e[0m\e[34m▶\e[0m "

    POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(context time dir)
    POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(status vcs)

    POWERLEVEL9K_SHORTEN_DIR_LENGTH=3
    POWERLEVEL9K_SHORTEN_STRATEGY="truncate_middle"

    POWERLEVEL9K_STATUS_VERBOSE=false
    POWERLEVEL9K_TIME_FORMAT="%D{%T}"

dritter:
@Tritlo just to be clear: You want the RPROMPT on the second line, right? I think there is no flexpoint for that now. We put some effort in showing the RPROMPT on the first line...

Here are the [the relevant lines](https://github.com/bhilburn/powerlevel9k/blob/81a4b3907de32b3dfb54fe6b626c1aa2b5faa13b/powerlevel9k.zsh-theme#L910-L911).

Tritlo:
Yes, on the second line. Thanks, I'll make a fork with the relevant changes, which I can use for me. 
I just want to be able to put more info, yet still be able to read the commands.

bhilburn:
@Tritlo - We would be interested in merging this feature into mainline. Once you have developed it, could you submit it as a PR so that we can pull it in?

[...]

PR #147, opened 2015-11-17
added variable to have rprompt on the newline if desired

Tritlo:
To alleviate issue #145


Tritlo:
Should I set RPROMPT_PREFIX and RPROMPT_SUFFIX explicitly to ''? I believe the current version works because they're unset, and thus render as empty, but this might not always be the case.

PR #148, opened 2015-11-17
added variable to be able to set the end of the prompt explicitly

Tritlo:
This is to allow users to have their prompt end in the way they want, e.g. if you want your prompt to end with "$", you can set POWERLEVEL9K_PROMPT_END="$".


Tritlo:
Wiki documentation to follow, if accepted.

PR #150, opened 2015-11-17
Add icon for end of left prompt.

Tritlo:
This is to allow users to have their prompt end in the way they want, e.g. if you want your prompt to end with "$", you can set POWERLEVEL9K_PROMPT_END="$".

### DIFF
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -22,6 +22,7 @@ case $POWERLEVEL9K_MODE in
     icons=(
       LEFT_SEGMENT_SEPARATOR         $'\UE0B0'              # 
       RIGHT_SEGMENT_SEPARATOR        $'\UE0B2'              # 
+      LEFT_SEGMENT_END_SEPARATOR     ' '                    # Whitespace
       LEFT_SUBSEGMENT_SEPARATOR      $'\UE0B1'              # 
       RIGHT_SUBSEGMENT_SEPARATOR     $'\UE0B3'              # 
       CARRIAGE_RETURN_ICON           $'\U21B5'              # ↵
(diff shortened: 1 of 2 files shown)


dritter:
I think it would be more elegant to introduce a new Icon than a new variable. Users could still overwrite that specific icon. and we just need to change our code to a new icon definition.
Could you change that? For the icon name I suggest `LEFT_SEGMENT_END_SEPARATOR` or `LEFT_SEGMENT_LAST_SEPARATOR`.

dritter:
Ah. Now I got you wrong. I thought you want to substitute the last segment separator with a user defined character. But you want to append a character, so that you have something like 
bildschirmfoto 2015-11-18 um 21 46 26, right?
In that case:
It seems that your code outputs two whitespaces, if `POWERLEVEL9K_PROMPT_END` was not specified.
I would suggest to modify your code like:
```zsh
  echo -n "%f"
  if [[ -n $POWERLEVEL9K_PROMPT_END ]]; then
    echo -n "%{%f%}$POWERLEVEL9K_PROMPT_END"
  else
    echo -n "%f "
  fi
```

To my screenshot: yes, I missed a whitespace, that your code would have, but like this every user could decide where to put whitespaces.

Tritlo:
I'm going to rework this to be more general, i.e. a plugin to put a custom string anywhere the user wants (instead of only at the end of the prompt). Working on it now.

Tritlo:
I realized that this could be useful as well, so I fixed it up to use an icon, as you ( @dritter ) suggested.

dritter:
:+1: That works for me with OMZ, prezto, antigen and plain ZSH.

bhilburn:
Awesome contribution, @Tritlo!

Merging with @dritter's approval.

Issue #151, opened 2015-11-18
Battery displays incorrect time remaining


Tritlo:
In [line 272](https://github.com/bhilburn/powerlevel9k/blob/e79a2ffff002a373f79ee25147df92411d174b26/powerlevel9k.zsh-theme#L272) in the prompt_battery function, there is some attempt at converting the 00:08:47 string that acpi outputs to hours and minutes, but for me, it just outputs an incorrect time, i.e., when acpi says that 00:08:47 is remaining, we get

    $ date -u -d @000847 +%k:%M 
    0:14

which is not the same as the 0:09 that should be reported.

Tritlo:
It seems that it's trying to use 00:08:47 as total seconds, which would amount to 847 seconds, and not 8*60 + 47 = 527 seconds

bhilburn:
Closed by merge of #152.

PR #152, opened 2015-11-18
Battery segment fix and enhancement.

Tritlo:
This closes issue #151, and also makes the coloring make more sense. Now the color of the text follows the color of the icon, i.e. red text for low battery, yellow for charging, etc.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -213,10 +213,11 @@ prompt_battery() {
   # set default values of not specified in shell
   [[ -z $POWERLEVEL9K_BATTERY_CHARGING ]] && POWERLEVEL9K_BATTERY_CHARGING="yellow"
   [[ -z $POWERLEVEL9K_BATTERY_CHARGED ]] && POWERLEVEL9K_BATTERY_CHARGED="green"
-  [[ -z $POWERLEVEL9K_BATTERY_DISCONNECTED ]] && POWERLEVEL9K_BATTERY_DISCONNECTED="$DEFAULT_COLOR"
+  [[ -z $POWERLEVEL9K_BATTERY_DISCONNECTED ]] && POWERLEVEL9K_BATTERY_DISCONNECTED="$DEFAULT_COLOR_INVERTED"
   [[ -z $POWERLEVEL9K_BATTERY_LOW_THRESHOLD ]] && POWERLEVEL9K_BATTERY_LOW_THRESHOLD=10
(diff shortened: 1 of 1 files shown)


Tritlo:
is %f not stacked, i.e. if I do 

    %F{blue}%F{green}%f

will the color not be blue still? I'm worried that we won't have equal numbers of %f and %F

rjorgenson:
I don't believe you need a %f for every %F, in it's current state it has two foreground color declarations(one for the battery icon, one for the time remaining string) and only one %f at the end to reset it. When using your example above, the output after %F{blue} but before %F{green} will be blue, anything after %F{green} and before %f will be green, anything after %f will be the default color of the terminal. Basically %F means "start printing this color until I say otherwise" and %f means "stop printing any color"

Tritlo:
Alright, then your version makes more sense.

rjorgenson:
Also I had $conn only be set in the instance of a color being applied to the icon so that it would inherit the specified foreground color instead of using the default. We can add a new line to explicitly set it to the foreground color at the top like the others and remove line 258

```
[[ -z $POWERLEVEL9K_BATTERY_FOREGROUND ]] && local conn="$DEFAULT_COLOR" || local conn=$POWERLEVEL9K_BATTERY_FOREGROUND
```

The general thinking behind the way I did it was that unless explicitly set in an option, color should only be there when it's indicating something, and should be as minimal as possible since it's supposed to stand out.

rjorgenson:
I guess with the new changes we'd want to set conn to DEFAULT_COLOR_INVERTED instead, otherwise it will just be set to the same as the background color by default.

Tritlo:
alright, that's been fixed, and the commits squashed together.

rjorgenson:
It still shows just explicitly setting conn to the default color, could you add the line from above that allows it to be the foreground color by default if one is specified.

Tritlo:
You want to overwrite the color of the icon as well?

rjorgenson:
The way it's written, $conn takes precedence over the foreground color, but if there's no condition to set $conn(it's only set when fully charged, charging or low), it should display the icon with the foreground color. The way it is right now it will display the icon with the default inverted color regardless of any user settings when it's in it's normal operating range and not being charged. $conn was intentionally left empty before if not explicitly set to inherit the foreground color, whatever it happened to be set to. If we want it explicitly set to something it should still maintain the same functionality.

Tritlo:
Ok, now it defaults to the foreground color.

rjorgenson:
Thanks =]

bhilburn:
Great PR, @Tritlo! And thanks for providing input, @rjorgenson!

PR #153, opened 2015-11-18
Custom command segment.

Tritlo:
This segment allows the user to run an arbitrary command and display the result as a segment. Useful to enable users to monitor whatever they want, without having to muck around in powerlevel9k itself.

    POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(context time battery dir vcs virtualenv custom_command)

    POWERLEVEL9K_CUSTOM_COMMANDS=("echo signal: \$(nmcli device wifi | grep yes | awk '{print \$8}')" )
    POWERLEVEL9K_CUSTOM_COMMAND_BACKGROUNDS=("blue")
    POWERLEVEL9K_CUSTOM_COMMAND_FOREGROUNDS=("yellow")
    POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(aws status load ram)

gives

custom_command

Instead of defining the command inline (if it is kinda long or unreadable), one can also add a function to the .zshrc like:
 
    zsh_signal(){
         local signal=$(nmcli device wifi | grep yes | awk '{print $8}')
         local color='%F{yellow}'
         [[ $signal -gt 75 ]] && color='%F{green}'
         [[ $signal -lt 50 ]] && color='%F{red}'
         echo -n "%{$color%}  $signal%{%f%}"
  
  }

And then just change custom commands array to read

    POWERLEVEL9K_CUSTOM_COMMANDS=("zsh_signal")

With this updated command, this looks like:

signal

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## v0.3.0 (next)
 
+### New segment `custom_command` added
+
+A new segment that allows users to define a custom command was added.
+
 ### `virtualenv` changes
 
 This segment now respects `VIRTUAL_ENV_DISABLE_PROMPT`. If this variable is set 
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ The segments that are currently available are:
 * [aws](#aws) - The current AWS profile, if active.
 * [battery](#battery) - Current battery status.
 * [context](#context) - Your username and host.
+* [custom_command](#custom_command) - A custom command to display the output of.
 * [dir](#dir) - Your current working directory.
 * **go_version** - Show the current GO version.
 * **history** - The command number for the current line.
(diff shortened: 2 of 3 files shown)


Tritlo:
I fixed an issue with the current bg color not being set correctly, and also added documentation on the nicer function way to define custom commands.

Tritlo:
This could be extended to allow the background/foreground color to be dynamic as well, but I think that would be a little overkill.

dritter:
Great Idea!
I have some remarks on that:

1. Why not having a associative array for the configuration? IMHO that would make it somewhat easier than having the need to sync multiple arrays (at least if the user wants to define colors for some of the custom commands, but not all).

2. I think it would be way more usable if users could define multiple `custom command` segments. It may be that they don't want to define a single insertion point for all their custom commands. It could work like (pseudo code):
    - let users define like `POWERLEVEL9K_CUSTOM_COMMAND1=(cmd "echo foo | grep bar" fg "black" bg "green" )` and reference that custom commands in their prompt-segment definitions `POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(time custom_command1 dir custom_command2 vcs custom_command3)`
    - check in `build_left_prompt` and `build_right_prompt` for segments that start with the string "custom_command" (this is kind of ugly) and give the number as parameter to the function
    - check in that function for variables named like `${(kvP)POWERLEVEL9K_${(U)custom_command${number}}}` (read in the associative array by name)

Tritlo:
They can define multiple entry points. I do like the idea of an associative
array

On lau., 21. nóv. 2015 at 00:15, Dominik Ritter <notifications@github.com>
wrote:

> Great Idea!
> I have some remarks on that:
>
>    1.
>
>    Why not having a associative array for the configuration? IMHO that
>    would make it somewhat easier than having the need to sync multiple arrays
>    (at least if the user wants to define colors for some of the custom
>    commands, but not all).
>    2.
>
>    I think it would be way more usable if users could define multiple custom
>    command segments. It may be that they don't want to define a single
>    insertion point for all their custom commands. It could work like (pseudo
>    code):
>    - let users define like POWERLEVEL9K_CUSTOM_COMMAND1=(cmd "echo foo |
>       grep bar" fg "black" bg "green" ) and reference that custom
>       commands in their prompt-segment definitions POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(time
>       custom_command1 dir custom_command2 vcs custom_command3)
>       - check in build_left_prompt and build_right_prompt for segments
>       that start with the string "custom_command" (this is kind of ugly) and give
>       the number as parameter to the function
>       - check in that function for variables named like
>       ${(kvP)POWERLEVEL9K_${(U)custom_command${number}}} (read in the
>       associative array by name)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bhilburn/powerlevel9k/pull/153#issuecomment-158553720>
> .
>
-- 
---
Matthías Páll Gissurarson

dritter:
> They can define multiple entry points.

Yes, but the order of that segments is not defined in the `POWERLEVEL9K_LEFT_PROMPT_ELEMENTS` array any more, right? That now happens in `POWERLEVEL9K_CUSTOM_COMMANDS`.

Tritlo:
Yes. We could fix this by having it called custom_command_1, custom_command_2 etc. in elements, and then use that as a key to look up in the custom commands array. Is that really necessary though? I mean, we're already using the order in left elements to define the order of left elements, why not use the order in custom commands to define the order of custom commands?

Tritlo:
I should perhaps be more clear in the example. If you do

    POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(context time battery custom_command dir vcs virtualenv custom_command )
    POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(aws status load ram custom_command)
    POWERLEVEL9K_CUSTOM_COMMANDS=("zsh_signal" "zsh_unixtime" "zsh_docker" )
    POWERLEVEL9K_CUSTOM_COMMAND_BACKGROUNDS=("black" "yellow" "blue")
    POWERLEVEL9K_CUSTOM_COMMAND_FOREGROUNDS=("white" "black" "white")

With `zsh_signal` as before, and `zsh_docker` and `zsh_unixtime` the following:

    zsh_docker(){
        echo " $(docker ps -a | grep Up | wc -l)"
    }

    zsh_unixtime(){
        echo "$(date +%s)"
    }

Then you get:

Tritlo:
As you can see, I could add a docker command for myself pretty easily, by just defining one in .zshrc, and not having to do anything in the source of powerlevel9k. Pretty useful!

Tritlo:
Implementing the order of custom_commands in the elements array would take a change in how build_$1_prompt works, which would very possibly slow down the entire rendering.

dritter:
Yes, to have the ability to display custom commands in the prompt is very useful. Definitely.

Just saying that I think it is quite confusing to get a sense of which segments are displayed at a quick look. Well, we could accept that if we argument that having custom commands in your prompt is a feature for advanced users anyway.

The performance argument is not a big thing in my eyes. In `build_$1_prompt` there would just be an additional condition that checks if the segment name starts with "custom_command" and if so prepares that segment. But I agree that is bit of a hack as this code doesn't belong there.

Tritlo:
Alright, I've updated it so that it is simpler, and more clear what one is defining where.

Tritlo:
See the updated documentation here:

https://github.com/Tritlo/powerlevel9k/tree/custom_string#custom_command

dritter:
Awesome! Thank you. This is much cleaner in my eyes now.

Last comments:

If you rename the function to `prompt_custom`, you can make use of the default color-overriding mechanism. The function would then look like:
```zsh
prompt_custom() {
   local name=$2:u
   local prefix=POWERLEVEL9K_CUSTOM_$name
   local command_var_name=${prefix}
   local command=${(P)command_var_name}

   "$1_prompt_segment" "${0}_${name}" $DEFAULT_COLOR_INVERTED $DEFAULT_COLOR "$(eval $command)"
 }
```

Some of your screenshots show a lot of black space. Could you crop them?

And could you add an entry to the changelog?

Tritlo:
Done.

Tritlo:
Shortened it even more.

dritter:
:+1:  Great work! Thanks!

Works for me on antigen, prezto, OMZ and plain ZSH.

Just some (really) final things: 
  - could you use the unicode code points for the icons in the documentation, so that other users could copy & paste it? That would be great.
  - in the docs you still state that users have to define a "custom commands" array. That isn't right any more.

Tritlo:
Alright, I've fixed those concerns.

dritter:
You missed one :wink:  
`POWERLEVEL9K_CUSTOM_COMMANDS=("zsh_signal" "zsh_unixtime" "zsh_docker" )`

Tritlo:
Whoops, fixed.

dritter:
Awesome! :+1:

Tritlo:
So it this good to merge @dritter @bhilburn ?

dritter:
From my side: Yes. :)

bhilburn:
@Tritlo - Fantastic contribution!!! Thanks so much for your hard work!

PR #154, opened 2015-11-21
Try to fetch the utility function path automatically

dritter:
For ZSH 5.0.8 `whence -v` tells also the path to the command. In that case we can use that information to get a proper path to the utility functions.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -32,9 +32,14 @@ if [[ $(whence -w prompt_powerlevel9k_setup) =~ "function" ]]; then
       print -P "%F{red}We could not locate the installation path of powerlevel9k.%f"
       print -P "Please specify by setting %F{blue}POWERLEVEL9K_INSTALLATION_PATH%f (full path incl. file name) at the very beginning of your ~/.zshrc"
       return 1
+    elif [[ -L "$POWERLEVEL9K_INSTALLATION_PATH" ]]; then
+      # Symlink
+      0="$POWERLEVEL9K_INSTALLATION_PATH"
     elif [[ -f "$POWERLEVEL9K_INSTALLATION_PATH" ]]; then
+      # File
       0="$POWERLEVEL9K_INSTALLATION_PATH"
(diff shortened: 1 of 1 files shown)


powerlevel9k.zsh-theme
```diff
@@ -19,9 +19,17 @@
...
 # Check if the theme was called as a function.
 if [[ $(whence -w prompt_powerlevel9k_setup) =~ "function" ]]; then
-  # Script is a function! We assume this to happen only in
-  # prezto, as they use the zstyle-builtin to set the theme.
-  0="${ZDOTDIR:-$HOME}/.zprezto/modules/prompt/functions/prompt_powerlevel9k_setup"
+  autoload -U is-at-least
+  if is-at-least 5.0.8; then
+    # Try to find the correct path of the script.
+    0=$(whence -v $0 | sed "s/$0 is a shell function from //")
+  else
```
bhilburn: I'm a little confused on this line.

The `if` conditional is gating on a ZSH version, 5.0.8, but the comments for the `else` code path seems to indicate that it should only be taken for `prezto`. Does this conditional work, then, if someone is using ZSH < 5.0.8 and not using prezto?

dritter: You found exactly the last loophole. ;)
I haven't found a solution for someone who is using ZSH < 5.0.8 and not prezto. And I am afraid there is no solution for this case. Except for the workaround of setting the path by a variable (maybe if the file `${ZDOTDIR:-$HOME}/.zprezto/modules/prompt/functions/prompt_powerlevel9k_setup` is not present). But I think this is not a good solution..
But I should update the comments at least. They are indeed appropriate.

Tritlo: You could propbably just add a check for a file, i.e. see if the 
`.zprezto/modules/prompt/init.zsh` file exists. Then you know that prezto is at least present (and thus probably being used).

dritter:
Okay. Now we try to find a prezto installation. If this failsat , all we can do is print a nice warning that the user should set a variable called `POWERLEVEL9K_INSTALLATION_PATH`.

bhilburn:
@dritter - Great fix to my previous comment! I like the completeness of your location checking, as well.

PR #155, opened 2015-11-21
Fix for the GO version.

dritter:
The `go_version` segment printed just "go", instead of "go1.2.1". Now it works correctly.


bhilburn:
Hah. Great catch!

PR #156, opened 2015-11-21
Node cleanup

dritter:
Just stumbled upon some code-issues in the `node_version` and `nvm` segments.


bhilburn:
Nice catch, @dritter!

PR #157, opened 2015-11-24
Performance improvement of battery segment on OSX

dritter:
I reworked the battery segment for some performance. Mainly this performance is better by avoiding ruby for simple calculations and pulling the data in over and over.

IMHO the linux part could need some more structure as well. I'll do that if I have more time.

/cc @rjorgenson @tritlo


dritter:
I made a new [branch](https://github.com/dritter/powerlevel9k/tree/battery_rework2) on top of this one, where I added the concept of "states" that a battery can have. This makes the code much clearer IMHO. In there I remove the color-only icon version. I hope that is acceptable for you. Just wanted to discuss that beforehand with you @rjorgenson and @tritlo.

What do you think? Is the code really clearer, or is this just my perception? Do we need the color-only version here as well?
Thanks for your feedback.

Tritlo:
I think this is much clearer. Am I correct in reading that this displays the whole segment in the same color?

Additionally:
 Since you're already reworking, could you maybe add a "non-verbose" setting? Where only the icon is displayed?

rjorgenson:
The code looks good to me. I don't like the idea of having the whole string be the color of the battery indicator, but if I'm alone in that then there's no need to have it in the project.

bhilburn:
It looks like @Tritlo's request for a `non-verbose` mode has been incorporated.

What do you think of @rjorgenson's feedback regarding the color, @dritter?

dritter:
@bhilburn It is a perfectly valid request. I am working on it. :)

dritter:
Okay. I tried different things on that and came to the conclusion that we don't need an extra implementation.

But first, here is what I tried:

1. Adding an extra switch to the battery segment. That seems obvious and easy, but I didn't like it very much, because it would make the battery segment quite special over the other ones (all other segments are in one color). Here is the code:
```zsh
  ...
  # prepare string
  local message="$(print_icon 'BATTERY_ICON' $current_state)"
  if [[ "$POWERLEVEL9K_BATTERY_COLORIZE_ICON_ONLY" == true ]]; then
    message="$message%f%F{$DEFAULT_COLOR_INVERTED}"
  fi

  # Default behavior: Be verbose!
  set_default POWERLEVEL9K_BATTERY_VERBOSE true
  ...
```

2. The second try was to introduce "states" of a segment to the `print_icon` function. In that case we would check if the user defined a variable named like the segment and state. That would add the flexibility to have different icons for the different states. Here is a [patch](https://github.com/bhilburn/powerlevel9k/files/44229/battery.patch.txt) of that code.

3. The third iteration was to hack the icon variable. Despite it feels like exploiting some SQL injection, it works like a charm. And if we don't start some wild credit card payment stuff in our theme, I would like to leave this hackish workaround there. @rjorgenson try to set a special icon variable `POWERLEVEL9K_BATTERY_ICON=$'\UE894'"%F{white}"`. What this variable does is: print the icon and start printing all text after it in white. The variable gets called when the battery segment is rendered and voilá we have just the icon in the right color.
That should do it. :wink:

rjorgenson:
Hmm, while that would work, I don't see why we can't do it the way it was done before? Just check for an explicit foreground color, and if it exists, apply it, if not, do nothing and it will inherit the battery color.

```[[ -z $POWERLEVEL9K_BATTERY_FOREGROUND ]] && local fg_color="" || local fg_color="%F{$POWERLEVEL9K_BATTERY_FOREGROUND}"```

that bit there sets the fgcolor variable to either nothing, or to a user defined color. just include $fgcolor in the output after the battery icon and if the user sets it then it will be applied otherwise whatever color the icon is set to will be inherited.

**Comment by dritter** [comment]
@rjorgenson Hmm. That trick only worked well when the segment had only one state and you were rebuilding the general color overriding mechanism. Now users could override colors for all states, and we would have to cover all that. To be honest, I think this would make the segment quite complicated.

But I see that the idea of coloring the icons different from the text is a good one. But I would like to have a more general solution (which is not the workaround I mentioned). I thought the right could be the `print_icon` function. Thinking about it again, why not setting a global `COLORIZE_ICON_ONLY` variable which leads to ending the foreground color after the icon was printed. That would probably work for all segments..

**Comment by rjorgenson** [comment]
@dritter Wouldn't that only reset the color to your terminals default, instead of being what the foreground color is set to? Maybe if you thought about it as overriding the color of the icon output with the state of the battery instead of overriding the text output color? That was my original goal when I wrote it, to fit in with the way the rest of the segments handle color, but enable the battery icon color to be set based on the state. The only other way I could think of to do it was to build the option into the print_icon function, which would have made it more general and usable by all segments, but it would then have to keep track somehow of the color that was set before it to set it back to that after printing the icon and I couldn't find a good way to do that. Any thoughts on that approach?

**Comment by dritter** [comment]
@rjorgenson Yes, it would reset to the terminals default. But that would not be a big problem, no? Currently the icons get printed at the very beginning or the very end of each segment (with the exception of the `vcs` segment). So it would always be a game of "disable the color until or after this icon".. But maybe I am missing a point.

> The only other way I could think of to do it was to build the option into the print_icon function, which would have made it more general and usable by all segments, but it would then have to keep track somehow of the color that was set before it to set it back to that after printing the icon and I couldn't find a good way to do that. Any thoughts on that approach?

Without the "keeping track of the color" idea (which I don't understand), this is a matter of disabling the foreground color after the icon has been printed, right?

So `print_icon`in a "first icon then text"-like segment would behave like:
1. print the icon
2. disable the foreground color if a colorize icon only variable is present

In a "first text then icon"-like segment, we would need to do more (maybe that is what you meant with "keeping track"):
1. the outer segment needs to disable the foreground color if a colorize icon only variable is present
2. we need to provide the color to the `print_icon` function
3. print icon

This is a quick shot, but I am sure we will get to a good solution. :)

**Comment by rjorgenson** [comment]
@dritter So basically my thinking is, say someone want's all the text on their prompt to be blue, but they want their default color to be the standard grey color. If we just reset the color after the icon the text will be the default grey. It would be nice if they could specify the foreground color of the battery icon segment to whatever they wanted and still have the battery icon be colored with the status of the battery.

In terms of "keeping track" I mean we would need to know inside the print_icon function what the color was set to when we called it, so at the end of the icon we could set the color back to what it was before, instead of resetting it to the default. If we could find a way to call print_icon without disrupting the current color, we could theoretically put icons in the segments anywhere we wanted without disrupting any of the previous output. I think that would be the best solution, if it's possible.

**Comment by rjorgenson** [comment]
@dritter Also I don't think we should be focusing on a COLORIZE_ICON_ONLY type setup, because we still want the ability to colorize both the text -and- the icon, but in this specific case, what the user specifies as their color for the segment should not affect the battery icon, it should always be in one of the states defined.

**Comment by dritter** [comment]
Hmm. Yes, it might be not the best solution if users want to set the colors of the icon and text individually.

### print_icon variant
So, why don't we pass the current foreground color and state to `print_icon`?
In this case we could
- check for a special icon for that state
- check for a variable that defines a color for the icon
- set the color again, after we printed the icon

So to get the behavior of your example, you need to set:
```zsh
# Text colors
POWERLEVEL9K_BATTERY_LOW_FOREGROUND='blue'
POWERLEVEL9K_BATTERY_FOREGROUND='blue'
POWERLEVEL9K_BATTERY_CHARGING_FOREGROUND='blue'
POWERLEVEL9K_BATTERY_CHARGED_FOREGROUND='blue'
# Icon colors
POWERLEVEL9K_BATTERY_ICON_LOW_FOREGROUND='red'
POWERLEVEL9K_BATTERY_ICON_FOREGROUND='white'
POWERLEVEL9K_BATTERY_ICON_CHARGING_FOREGROUND='yellow'
POWERLEVEL9K_BATTERY_ICON_CHARGED_FOREGROUND='green'
```

This is a bit verbose, but flexible.

Another solution could be:
### Icon as visual identifier
How about we treat an icon as a visual identifier for the segment. This could be done by passing the icon as a separate parameter to `$1_prompt_segment`.
In there we could:
- Check for a defined variable to change the icon color separately
- Position the icon dependent of the element position (at the beginning for left; at the end for right). 

This has some drawbacks though: 
- We need a mechanism for the whitespace around an icon, as some are double width and need more space.. 
- We would eventually need to pass the state of the segment as well to `$1_prompt_segment`, so that the icon is a flexible as in the above solution. 

A good thing about this is that the code would have a better structure and the concept of segments having "state" would become crucial (at least I like it :smile: ).

Btw. I already implemented some of this in my `improve_rvm` branch.

terabyte:
see #161 for my take on floating point in bash (everyone has "bc" installed, right?) and how I got the power stuff in linux (looks to be similar idea to yours, but I think I'm using a different path under proc - the path I am using has worked for me on every laptop I tried, I think the one you are using might be less universal)

What would be *really* crazy is to include a "time left" mode as well - I prototyped something like this by keeping some state about the rate of change of the current power level, but it never worked too well so I gave up on it.

Cool changes though!

terabyte:
oh, hmm, looks like your code reads a "time left" out of /proc if it is avaiable, that would probably be less "insane" and more "reliable" =)

bhilburn:
@dritter - Looks great! Thanks so much for the improvements!

@terabyte, @rjorgenson, @Tritlo - Thanks for providing excellent feedback and testing for this PR! It is much appreciated.

Issue #158, opened 2015-11-25
Purpose of home icon in dir segment?


keirlawson:
Having updated this theme I now get a home icon in front of my dir segment, however this icon persists even when the path is truncated and thus does not include the home directory.  I guess my expected behaviour would be the home icon replaces ~, but it seems to be in addition to it and regardless of where you are in the tree?

dritter:
@keirlawson Thanks for the report. Yes, you are right. The home icon is always shown, regardless of being in the home directory or not. When I implemented it, I haven't thought about it that way. The reason why this is an home icon is really profane: I just like icons :grinning: 
Almost all other segments in `awesome` mode have icons and I find the home icon quite fitting for the directory segment.

It wouldn't be much work to show the home icon only if we are under the home directory. But I think there should be another icon for the case of being outside of the home folder. Do you have any idea which icon that could be?

rjorgenson:
@dritter I had a similar thought to @keirlawson when I started using powerlevel9k, I ended up adding POWERLEVEL9K_HOME_ICON="\ue818 " to my zshrc which is a folder icon. Looks like 
<img width="661" alt="screen shot 2015-11-25 at 3 52 05 pm" src="https://cloud.githubusercontent.com/assets/44096/11411370/b0f39848-938c-11e5-9b4c-836cdc8c55c3.png">

bhilburn:
I don't use the awesome icons, so I've actually never experienced this. That said, if I did use the awesome icons, I actually would want this behavior:

1. If in the home directory, instead of printing the directory, just show the home icon.
2. In all other directories, print the directory, **without** an icon.

@dritter - How hard would it be to implement the following:

1. If awesome fonts are in use, and we are in the home directory, print the home icon instead of the string directory.
2. Have another config env variable that enables / disables an icon for all other directories - perhaps the folder icon suggested by @rjorgenson?

dritter:
@rjorgenson Great icon! Must have overseen that one. :+1: 

@bhilburn Frankly I don't like the idea of have special code for the different modes. That would clutter up the code quite a bit (too bad that we have to do that for different OS).
Disabling the not-home directory would be rather easy, as users could just set the according icon variable to an empty string.

So, my favorite behavior would be:
1. if in the home directory (or deeper), print the home-icon _and_ the path (even if it is just `~`)
2. In all other directories print the folder icon (thx @rjorgenson ) and the path

bhilburn:
[...]

PR #160, opened 2015-11-27
Home/folder icon fix

dritter:
This is a PR for #158 . Now if you are beneath the home directory, you'll get a home icon. If you are somewhere else, you'll get a folder icon.


Tritlo:
Nice! Looks good to me!

dritter:
@Tritlo Could you do me a favor? I see you use the `awesome-fontconfig` mode. Could you give me the code point for the folder icon: 
bildschirmfoto 2015-11-28 um 00 31 07


Thank you very much!

Tritlo:
The closest that I see are

      : f115  and  : f116 

and 
    
      : f07c   : f07d

dritter:
Now it works correctly. ;) The truncation made it a bit harder, so we had to evaluate the path always by our own.

PR #161, opened 2015-12-06
Adding 'power' segment for laptop battery level

terabyte:
thought you might like this.  I haven't done any font stuff so you might want to add a battery symbol into it or something?  also, not sure how portable my method for getting powerfull/powernow is, but it works on every laptop I own (N=3) across various versions of debian and ubuntu.  Would def. need to be ported to work on OS X.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -644,6 +644,45 @@ prompt_dir() {
   "$1_prompt_segment" "$0" "blue" "$DEFAULT_COLOR" "$(print_icon 'HOME_ICON')$current_path"
 }
 
+# power: power status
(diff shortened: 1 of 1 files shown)


terabyte:
(also, thanks for this awesome theme)

terabyte:
oh man, I just saw https://github.com/bhilburn/powerlevel9k/pull/157, looks way better, gonna go try that now instead. =|

terabyte:
yeah, ok, I'm closing this in favor of 157.  nevermind! =)

dritter:
@terabyte Welcome! The development of this theme is quite hidden in the `next` branch. :wink: Feel  free to try it out.

PR #162, opened 2015-12-09
Adding command prompt for AWS Elastic Beanstalk environment name

bowmessage:
Print out the current AWS Elastic Beanstalk environment name when the ./.elasticbeanstalk folder exists. Great complement to the EB CLI.

I'm not sure about the icons (couldn't find anything equivalent in the other icon sets), so I just used the emoji icon for all sets. Let me know if that's not OK and I can do some searching.

Cheers!

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -208,6 +208,14 @@ prompt_aws() {
   fi
 }
 
+# Current Elastic Beanstalk environment
+prompt_aws_eb_env() {
+  if [ -e .elasticbeanstalk/config.yml ]; then
+    local eb_env=$(cat .elasticbeanstalk/config.yml | grep environment 2> /dev/null | awk '{print $2}')
+    "$1_prompt_segment" "$0" black green "$(print_icon 'AWS_EB_ICON')  $eb_env"
+  fi
+}
+
 # Custom: a way for the user to specify custom commands to run,
 # and display the output of.
 #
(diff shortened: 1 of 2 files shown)


bhilburn:
@bowmessage - Great contribution! Thanks so much for the PR!

Can you update your PR to also include an update to the README that documents the new segment?

@dritter - Thoughts on the icon selection, per @bowmessage's original comment?

dritter:
I like the Icon. Couldn't find a better one. :)
The only problem I have with emojis is that you cannot change the color, so our color changing mechanism won't work.

powerlevel9k.zsh-theme
```diff
@@ -208,6 +208,14 @@ prompt_aws() {
   fi
 }
 
+# Current Elastic Beanstalk environment
+prompt_aws_eb_env() {
+  if [ -e .elasticbeanstalk/config.yml ]; then
+    local eb_env=$(cat .elasticbeanstalk/config.yml | grep environment 2> /dev/null | awk '{print $2}')
+    "$1_prompt_segment" "$0" black green "$(print_icon 'AWS_EB_ICON')  $eb_env"
```
dritter: This should be now:
```zsh
"$1_prompt_segment" "$0" "$2" "black" "green" "$eb_env" 'AWS_EB_ICON'
```
Two parameters are new:
- Second parameter is now the `joined` flag to determine if a segment should be joined with preceding one.
- The last parameter is the visual identifier - a key of the `icons` array.

dritter:
@bowmessage Ping

bhilburn:
We're over a month old on this one.

@dritter - Do you want to claim this and finish it up?

dritter:
@bhilburn Yes, I'll do. Also on 0.4.0.

bhilburn:
Re-done in #203

PR #164, opened 2015-12-13
Split up the `status` segment

### LINKED ISSUE #163: Can't change status icon color
Unfortunately, it is not possible to change the color of the fail icon. I tried to set:

```
POWERLEVEL9K_STATUS_FOREGROUND='white'
POWERLEVEL9K_STATUS_BACKGROUND='red'
```

The background is now red. However, the fail icon is still red, too.

My current fix is to overwrite the `prompt_status()` method, as the color of the icon is hardcoded:

```
[[ "$RETVAL" -ne 0 ]] && symbols+="%{%F{red}%}$(print_icon 'FAIL_ICON')%f"
```
to
```
[[ "$RETVAL" -ne 0 ]] && symbols+="%{%F{white}%}$(print_icon 'FAIL_ICON')%f"
```

dritter:
...into a `background_jobs`, `root_indicator` and `status` segment.

This fixes #163

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## v0.3.0 (next)
 
+### Added ability for "joined" segments
+
+You can now merge segments together by suffixing the segment name with "_joined".
+For Developers: Be aware that the order of parameters in left/right_prompt_segment
+has changed. Now a boolean parameter must be set as second parameter (true if joined).
+
 ### `status` changes
 
 The `status` segment was split up into three segments. `background_jobs` prints
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## v0.3.0 (next)
 
+### `status` changes
+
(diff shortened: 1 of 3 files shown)


dritter:
@tritlo @bhilburn good news! Now the segments can be joined together. You always join a segment with the previous one. If this segment is visually the first one (some segments are conditional), this has no effect. This even works for `custom` segments.

There are two drawbacks IMHO:

1. I changed the order of parameters in `left/right_prompt_segment`, as this new param is mandatory and the following ones are optional.

2. Users could think that joining segments together overrides the color of the second segment. This is not the case. It would look like the `flat` theme here. Although this is possible I retained the segments original color. I think this is more valuable than the overriding variant (e.g. if you join the `vcs`, you loose the coloring information).

Tritlo:
Souns great! I'vebeen wanting to merge some segments for a long time (like the wifi and battery indicator). If they want the same color, they can just customize it themselves, I think. Thanks for making sure it works for the custom segments, I thought that would be a pain.

PR #165, opened 2015-12-15
Battery segment: rework part 2

dritter:
I didn't forget about @rjorgenson's wish to easily change the color of an icon from #157 . Now I think I have a good piece of code. 
Now there is a concept of a "visual identifier" of a segment. This is much like a logo of a segment. Before this change we printed the icons into the message that is passed to `left_prompt_segment` and a big part of segments just have one icon..

What this code does is quite simple, `left_prompt_segment` take a new parameter that can have its own color (by setting a variable like `POWERLEVEL9K_<segment>_<state>_VISUAL_IDENTIFIER_COLOR` (e.g. `POWERLEVEL9K_BATTERY_DISCONNECTED_VISUAL_IDENTIFIER='red'`).

A quite extensive example:
```zsh
POWERLEVEL9K_DIR_HOME_VISUAL_IDENTIFIER_COLOR='green'
POWERLEVEL9K_DIR_HOME_FOREGROUND='cyan'
POWERLEVEL9K_DIR_DEFAULT_VISUAL_IDENTIFIER_COLOR='red'
POWERLEVEL9K_DIR_DEFAULT_FOREGROUND='yellow'
POWERLEVEL9K_FOLDER_ICON='Dir'
```

The benefits in short:
- General code: Every segment can have a visual identifier (and thus change its color)
- Cleaner code: The segments don't have to know anything about this
- Very flexible

I even did a branch for the "print_icon"-variant, but it didn't work out: The `print_icon` function is called first, and we don't know which color the user set for that segment. I think this is what @rjorgenson meant with "keeping track". Sorry for not getting your point.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ A new segment that allows users to define a custom command was added.
 
 ### `virtualenv` changes
 
-This segment now respects `VIRTUAL_ENV_DISABLE_PROMPT`. If this variable is set 
+This segment now respects `VIRTUAL_ENV_DISABLE_PROMPT`. If this variable is set
 to `true`, the segments does not get rendered.
 
 ### `load` changes
@@ -22,7 +22,7 @@ segment is able to show the free ram and used swap.
 ### `vcs` changes
 
 This prompt uses the `VCS_INFO` subsystem by ZSH. From now on this subsystem
-is only invoked if a `vcs` segment was configured. 
+is only invoked if a `vcs` segment was configured.
 
 ### `rvm` changes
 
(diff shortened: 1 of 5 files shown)


dritter:
I know this PR lacks of documentation and changelog. I just wanted to discuss the technical solution first. ;)

rjorgenson:
:thumbsup::thumbsup:

I searched and searched for a way to "set the foreground color to what it was before it was changed" in zsh and couldn't come up with anything realistic. This is a great way of solving the problem! =]

dritter:
Glad you like it. :)

bhilburn:
@dritter - Just to be sure, this PR is deprecated for the branch `integration_165_164`, right?

dritter:
@bhilburn Yep. Correct. :)

bhilburn:
@dritter - Go ahead and update this PR, and I'll merge.

dritter:
@bhilburn done. :)

PR #167, opened 2015-12-22
Joined Segments and visual Identifiers

dritter:
This is an integration PR of #164 and #165.


dritter:
@bhilburn On second thought, I prefer separate PRs 164 and 165, because of prettier history reasons.

Merging this pull request would look like the other two would have been closed without merging them, right?

If you merge 164 into next, I could push my merge to 165, and you could merge it again. Then both pull requests would be merged correctly..

bhilburn:
@dritter - Makes sense!

PR #170, opened 2015-12-29
Add a toggle for the HOME_ICON

natemccurdy:
Prior to this, the HOME_ICON was always visible when using an Awesome or
flat font. Some users might want to hide that icon to save a bit of
horizontal space. Also, as the HOME_ICON is visible even when not in a
home directory, it can be a bit misleading if a user were to think that
it meant they were currently in their home directory (essentially
duplicating the meaning of '~').

This commit adds a toggle called `POWERLEVEL9K_SHOW_HOME_ICON` that, when
`true`, will hide the icon from the `dir` prompt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,8 +129,14 @@ To change the way how the current working directory is truncated, just set:
     # default behaviour is to truncate whole directories
 
 In each case you have to specify the length you want to shorten the directory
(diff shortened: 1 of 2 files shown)


natemccurdy:
Closing this so I can re-create with a fresh check-out of the `next` branch. I should not have based my change off of `master` .

PR #171, opened 2015-12-29
Create a toggle for the icon in the dir prompt

natemccurdy:
Prior to this, the HOME_ICON and FOLDER_ICON was always visible when using an Awesome or flat font. Some users might want to hide that icon to save a bit of horizontal space.

This commit adds a toggle called `POWERLEVEL9K_SHOW_HOME_ICON` that, when `true`, will hide the icon from the `dir` prompt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,12 @@ In each case you have to specify the length you want to shorten the directory
 to. So in some cases `POWERLEVEL9K_SHORTEN_DIR_LENGTH` means characters, in
 others whole directories.
 
+When using the Awesome or flat fonts, there is an icon of a house at the left
+edge of this segment. To hide this icon, set:
+
+    # Hide the home icon
+    POWERLEVEL9K_SHOW_HOME_ICON=false
+
 ##### ip
 
 This segment shows you your current internal IP address. It tries to examine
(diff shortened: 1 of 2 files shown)


natemccurdy:
Hmm.. maybe the toggle should be called `POWERLEVEL9K_SHOW_DIR_ICON` now that the icon could be either HOME or FOLDER.

dritter:
This would have the same effect like setting the icons to an empty string, right?
```zsh
POWERLEVEL9K_HOME_ICON=''
POWERLEVEL9K_FOLDER_ICON=''
```

natemccurdy:
Ah, yes it would! I totally forgot that those could be set as environment variables :-)

natemccurdy:
@dritter Should I close this PR? Manually changing the 2 icons to empty string gets the same net-result. But if you think that a single toggle would be desired, there's the code :)

dritter:
@natemccurdy I am undecided about this one. Lets wait for @bhilburn ;)

bhilburn:
@natemccurdy - I like the idea behind this change, as it provides a single entry-point to achieve the desired output. The issue that I see, though, is that it adds an extra conditional to each render, which we try to minimize where not necessary.

I think the best thing to do at this point is give this configuration as an example in the documentation to make sure it's clear to users how to achieve this.

@dritter - Can you add that configuration to the docs as an example? @natemccurdy's desired goal is good, and I think there are a number of users that will want this.

PR #173, opened 2015-12-29
More consistent visual identifiers

dritter:
Now even segments that just have an icon as content print these icons as visual identifiers. Some of these segments have a verbose mode. If this verbose mode is false, some print just the icon (like the battery segment), which didn't work correctly before this change.
This is because the `$1_prompt_segment` methods checked if the segments had content, and only printed in that case. Now the segments are more in charge of they want to print or not.


bhilburn:
Great job turning a bug into a new feature, @dritter!

PR #174, opened 2015-12-30
added fish-like shortening strategy

niklas-heer:
In the FISH-Shell directories are shorten like this: `/u/s/plasma` for `/usr/share/plasma`

I added a new flag __truncate_fish_like__ to POWERLEVEL9K_SHORTEN_STRATEGY and basically removed the ".." from truncate_from_right.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,17 +207,16 @@ To change the way how the current working directory is truncated, just set:
 
 You can also change the delimiter (the dots in between) which is used to truncate the working directory. This setting is optional. The default are 2 dots.
 
-	# set the delimiter to an empty string to hide it
+    # set the delimiter to an empty string to hide it
     POWERLEVEL9K_SHORTEN_DELIMITER=""
     # or set it to anything else you want (e.g. 3 dots)
     POWERLEVEL9K_SHORTEN_DELIMITER="..."
 
 With this you could achive the truncate behaviour of the fish shell. Which turncates `/usr/share/plasma` to `/u/s/plasma`
 
-	POWERLEVEL9K_SHORTEN_DIR_LENGTH=1
-	POWERLEVEL9K_SHORTEN_DELIMITER=""
-	POWERLEVEL9K_SHORTEN_STRATEGY="truncate_from_right"
-
+    POWERLEVEL9K_SHORTEN_DIR_LENGTH=1
+    POWERLEVEL9K_SHORTEN_DELIMITER=""
+    POWERLEVEL9K_SHORTEN_STRATEGY="truncate_from_right"
 
 In each case you have to specify the length you want to shorten the directory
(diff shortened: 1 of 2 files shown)


dritter:
Cool idea! I recently saw some folks changing the delimiter (the two dots) as well.
I have a suggestion that goes in that direction. Why not introducing a new variable for that delimiter to make it more flexible? So everyone would be able to have one dot, three dots, whatever. You would get a fish like truncation by setting this variable to an empty string (and having `truncate_from_right` as strategy)..

niklas-heer:
@dritter great idea!

I solved it like this:
* I added a new variable which you can set in your config `POWERLEVEL9K_SHORTEN_DELIMITER`
* The new variable can be a string of the characters you want to use. (eg. "..") or false if you don't want to display any characters
* The default value is: ".." (used when nothing is set)
* The setting applies to all shorten strategies, but if nothing is set they behave like before. Except the default shorten strategy, which was altered to `../plasma` from `.../plasma`.

So to achieve FISH-Shell like shortening you would need to set the following:
``` bash
POWERLEVEL9K_SHORTEN_DIR_LENGTH=1
POWERLEVEL9K_SHORTEN_DELIMITER=false
POWERLEVEL9K_SHORTEN_STRATEGY="truncate_from_right"
```

dritter:
Looks great! :)

One short notice: we have a `set_default` method to set the default values for variables.

niklas-heer:
@dritter oh that's a really neat function :)

So the new behavior is this:
* You don't need to set it to false, instead you can set it to an empty string
* The default is still `".."`
* The default strategy is also altered to `../plasma` instead of `.../plasma`

So FISH-Shell like behavior is now achieved by this:
``` bash
POWERLEVEL9K_SHORTEN_DIR_LENGTH=1
POWERLEVEL9K_SHORTEN_DELIMITER=""
POWERLEVEL9K_SHORTEN_STRATEGY="truncate_from_right"
```

dritter:
Could you add some documentation? That would be great.

Happy new year, btw. ;)

niklas-heer:
Sure thing :)

Where should I add the documentation (here as a comment) and in which form (do you have an example)?

And also to you a happy new year. :)

dritter:
We have two locations for documentation. One is for more basic informations and one is for advanced configuration. The first one is the `README.md` right in the root folder. The latter one is in the [wiki](https://github.com/bhilburn/powerlevel9k/wiki/Stylizing-Your-Prompt).
IMHO enough people want to change the delimiter to put it in the readme (in the `dir` configuration block).

Thx.

niklas-heer:
@dritter Alright I've added it to the `README.md` file. I hope this is what you were looking for. :)

dritter:
Exactly. :+1:

bhilburn:
Thanks for the great contribution, @niklas-heer! And thanks for providing guidance, @dritter!

Great job, all.

PR #177, opened 2016-01-06
Added an explanation for the "home" and "folder" icons.

dritter:
As requested in #171 here is the documentation for the "home" and "folder" icons.

I tried to write it in a progressive way (like you want to add these icons), although @natemccurdy 's original intend was to turn these icons off, because I find it hard to formulate the other way around.

The other way would be something like:
```
If you use any of `awesome-patched`, `awesome-fontconfig` or `flat` modes, and you don't
want to show the folder or home icon, you can turn it off by...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,15 +206,16 @@ In each case you have to specify the length you want to shorten the directory
 to. So in some cases `POWERLEVEL9K_SHORTEN_DIR_LENGTH` means characters, in
 others whole directories.
 
-You can even show icons if your are within your homefolder or outside of it
-by setting:
 
-    # Greek "alpha" sign
-    POWERLEVEL9K_HOME_ICON='\u03B1'
-    # Greek capital "omega" sign
-    POWERLEVEL9K_FOLDER_ICON='\u03A9'
+| `Compatible` | `Powerline` | `Awesome Powerline` | Situation
+|------------|-----------|-------------------|----------------------------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -206,6 +206,16 @@ In each case you have to specify the length you want to shorten the directory
...
 
+You can even show icons if your are within your homefolder or outside of it
+by setting:
+
+    # Greek "alpha" sign
+    POWERLEVEL9K_HOME_ICON='\u03B1'
+    # Greek capital "omega" sign
+    POWERLEVEL9K_FOLDER_ICON='\u03A9'
+
+To turn off these icons you could set these variables to an empty string.
```
natemccurdy: As written, it's not entirely clear that there are already icons set for these when using the `awesome` family of fonts. That might make someone think that they need to set these icons to use them; which is not true.

Maybe throw in a sentence along the lines of:

> When using the `awesome` family of fonts, these icons are already set for you.

dritter: Nah. You are right. That is not clear. I'll think the best is to make a table like for the VCS segment..

bhilburn:
Thanks for working on these docs, @dritter and @natemccurdy!

bhilburn:
Also, I managed to totally forget this was outstanding before pushing the v0.3.0 tag. Derp.

May push a v0.3.1 in a few days to get this doc update live.

PR #181, opened 2016-01-08
Split up `ram` segment into `ram` and `swap`.

dritter:
This is more flexible, and we save a variable.

If users wish to have this as one segment, the can join them together..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## v0.4.0 (next)
+
+### `ram` changes
+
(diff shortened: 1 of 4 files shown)


bhilburn:
@dritter - Looks like we've got some merge conflicts on this one. Can you update?

dritter:
I think this is an issue for 0.4.0. I'll update this PR once 0.3.1 is out.

bhilburn:
@dritter - Roger.

bhilburn:
@dritter - Thanks for updating this! Looks ready `=)`

Issue #182, opened 2016-01-09
Unadvertised OSX dependency with coreutils


truthbk:
Just a quick heads up to let you know you have an unadvertised dependency with coreutils (_realpath_, _readlink_) on OSX (tested on 10.11). 

**`brew install coreutils`** will fix it.

dritter:
Ha! You are right. It is already complicated enough to install the fonts etc. I think we want to reduce the dependencies, so I try to get rid of realpath and readlink..

Thx for the report. :+1:

bhilburn:
Closed by merge of #183

PR #183, opened 2016-01-09
Improved symlink dereferencing on OSX by calling a perl script.

dritter:
This is a hotfix for the problem described in #182.
Instead of removing `readlink` and `realpath`, we try a third option (a perl script) to resolve the symlink.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -51,7 +51,7 @@ fi
 # If this theme is sourced as a symlink, we need to locate the true URL
 if [[ -L $0 ]]; then
   # Script is a symlink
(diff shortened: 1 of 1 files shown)


dritter:
@truthbk can you check this pease?

dritter:
Btw. This solution is from [stackoverflow](http://stackoverflow.com/questions/1055671/how-can-i-get-the-behavior-of-gnus-readlink-f-on-a-mac). It is not perfect, I had some situations with symlinked folders that were not dereferenced in linux environments (but there `realpath` and `readlink -f` would work).. But IMHO it should be sufficient in most cases.

bhilburn:
Worked in Linux, as well, actually.

bhilburn:
Merged into `next`.

PR #189, opened 2016-01-13
Add sub folder icon when in home subdirectories

Kayant:
Just wanted to help out but not 100% sure if this is the current way to fix this but I noticed when I was in a directory the home icon was still displayed instead of the folder icon but once this was removed it works correctly.

Thanks again for creating such a wonderful theme with lots of customisation.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -429,8 +429,10 @@ prompt_dir() {
   fi
 
   local current_icon=''
-  if [[ $(print -P "%~") == '~'* ]]; then
+  if [[ $(print -P "%~") == '~' ]]; then
     "$1_prompt_segment" "$0_HOME" "$2" "blue" "$DEFAULT_COLOR" "$current_path" 'HOME_ICON'
+  elif [[ $(print -P "%~") == '~'* ]]; then
+    "$1_prompt_segment" "$0_HOME_SUBFOLDER" "$2" "blue" "$DEFAULT_COLOR" "$current_path" 'HOME_SUB_ICON'
   else
     "$1_prompt_segment" "$0_DEFAULT" "$2" "blue" "$DEFAULT_COLOR" "$current_path" 'FOLDER_ICON'
   fi
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -44,6 +44,7 @@ case $POWERLEVEL9K_MODE in
       LINUX_ICON                     $'\UE271'              # 
       SUNOS_ICON                     $'\U1F31E '            # 🌞
       HOME_ICON                      $'\UE12C'              # 
+      HOME_SUB_ICON                  $'\UE18D'              # 
       FOLDER_ICON                    $'\UE818'              # 
       NETWORK_ICON                   $'\UE1AD'              # 
       LOAD_ICON                      $'\UE190 '             # 
@@ -96,6 +97,7 @@ case $POWERLEVEL9K_MODE in
       LINUX_ICON                     $'\UF17C'              # 
       SUNOS_ICON                     $'\UF185 '             # 
       HOME_ICON                      $'\UF015'              # 
+      HOME_SUB_ICON                  $'\UF07C'              # 
       FOLDER_ICON                    $'\UF115'              # 
       NETWORK_ICON                   $'\UF09E'              # 
       LOAD_ICON                      $'\UF080 '             # 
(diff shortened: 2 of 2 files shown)


rjorgenson:
I think the behavior as it was written is to show the Home icon anywhere underneath your home directory. So '~'* would match ~/, ~/Downloads, ~/.zsh, etc. With this change it would only show the home icon when you are in the root of your home directory. Not sure what the behavior should be, that's a tough decision.

dritter:
@rjorgenson Exactly.

But I see the point to display the home icon only if you are in the root of your home folder. Maybe we need a third icon for being in a subdirectory of your home folder. That would be an quite easy solution.

Kayant:
Ha ok I see. Thanks for the explanation @rjorgenson. That makes sense.

@dritter should I close the PR given it's not the right solution?

dritter:
@Kayant as @rjorgenson said, this is a tough decision. I think there are always folks that favor this or that solution. I'd like to hear some other people about that.
IMHO the best solution is to add an icon for subdirectories of the home folder. I'd be glad if you open a new PR for that.. :wink:

Kayant:
@dritter I [tried] (https://github.com/Kayant/powerlevel9k/commit/61fa7bea79be860ae08adbac3285aa2eac896a9f)
:tongue:

What's your option on something like this. I only half knew what I doing there  :smile:

powerlevel9k.zsh-theme
```diff
@@ -429,10 +429,12 @@ prompt_dir() {
   fi
```
dritter: You are almost there. Imagine this segment has three states:

1. Being at the root of your Home folder
    This is the first condition. So we want to match only that. Looks good in your changeset.

2. Being in a subdirectory of your Home folder
    The condition looks good (matching a path that _begins_ with a tilde.

3. Being anywhere else

powerlevel9k.zsh-theme
```diff
@@ -429,10 +429,12 @@ prompt_dir() {
...
 
   local current_icon=''
-  if [[ $(print -P "%~") == '~'* ]]; then
+  if [[ $(print -P "%~") == '~' ]]; then
     "$1_prompt_segment" "$0_HOME" "$2" "blue" "$DEFAULT_COLOR" "$current_path" 'HOME_ICON'
-  else
+  elif [[ $(print -P "%~") == '~'* ]]; then
     "$1_prompt_segment" "$0_DEFAULT" "$2" "blue" "$DEFAULT_COLOR" "$current_path" 'FOLDER_ICON'
```
dritter: This is the second state. So the icon should be the new `HOME_SUB_ICON` and you need to push that state (something like `$0_HOME_SUBFOLDER` instead of `$0_DEFAULT`) to `$1_prompt_segment`.

powerlevel9k.zsh-theme
```diff
@@ -429,10 +429,12 @@ prompt_dir() {
...
+  else
+    "$1_prompt_segment" "$0_HOME" "$2" "blue" "$DEFAULT_COLOR" "$current_path" 'HOME_SUB_ICON'
```
dritter: This is the third state. So it should be `FOLDER_ICON` and instead of `$_HOME` it should be `$0_DEFAULT`.

dritter:
And I can't see the icon `\U1F5C0`. Is this the correct codepoint for `awesome-patched`?

Kayant:
Wow! Thanks a lot for all the comments that's a big help for a noob like me :smile. Am not 100% sure. I got the code from searching online. I wasn't sure how to find a new folder icon.

Kayant:
I wouldn't have access to a system for awhile so I will submit a reversion a little later.

bhilburn:
@Kayant - It looks like your changeset is coming along well! Thanks, @dritter, for providing guidance!

Don't forget to update the docs to reflect the changes. Also, can you post a screenshot of what the new icon looks like?

dritter:
@Kayant You can select an right icon from your font book on mac. Just hover the icon and you get the code point.
I don't know how to find a code point for an icon..

Kayant:
Thanks a bunch guys for the hints and help. It's feel great being able to contribute something even with my limited knowledge :smile. I will get in the changes tomorrow hopefully.

dritter:
I can only help with the `awesome-patched` mode. That works for me. Well done @Kayant .

Btw. I always do screenshots from the font book.. It was the easiest for me. Here is the one for your Icon:

Kayant:
OMG! Thank you. I don't know why I didn't think about that lool.

Kayant:
Closing this now in favour of #192

PR #191, opened 2016-01-14
added check for nvm command before running

rjorgenson:
Added a check for nvm command, was throwing errors on machines that don't have nvm installed.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -531,7 +531,7 @@ prompt_node_version() {
 # Node version from NVM
 # Only prints the segment if different than the default value
 prompt_nvm() {
-  [[ $(type nvm) =~ 'nvm is a shell function'* ]] && return
+  [[ ! $(type nvm) =~ 'nvm is a shell function'* ]] && return
   local node_version=$(nvm current)
   local nvm_default=$(cat $NVM_DIR/alias/default)
   [[ -z "${node_version}" ]] && return
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -531,14 +531,13 @@ prompt_node_version() {
 # Node version from NVM
 # Only prints the segment if different than the default value
 prompt_nvm() {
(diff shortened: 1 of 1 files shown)


rjorgenson:
Also found a bug in the battery prompt that was throwing the below error. Capturing the command output in quotes fixes the issue.

```prompt_battery:local:16: not valid in this context: |```

**Comment by bhilburn** [comment]
@rjorgenson - Thanks for the fixes!

 Just so I understand, with this change, what is rendered in the `nvm` segment if the command doesn't exist? Does the segment not draw at all, or do you end up with just two segment separators?

**Comment by rjorgenson** [comment]
@bhilburn good catch! Added new commit to not output anything unless $node_version was populated.

File: powerlevel9k.zsh-theme [hunk 1]
```diff
@@ -531,12 +531,14 @@ prompt_node_version() {
...
-  [[ "$node_version" =~ "$nvm_default" ]] && return
+  if [[ $(type nvm) =~ 'nvm is a shell function'* ]]; then
+    local node_version=$(nvm current)
+    local nvm_default=$(cat $NVM_DIR/alias/default)
+    [[ -z "${node_version}" ]] && return
+    [[ "$node_version" =~ "$nvm_default" ]] && return
+  fi
 
-  $1_prompt_segment "$0" "$2" "green" "011" "${node_version:1}" 'NODE_ICON'
+  [[ -n "${node_version}" ]] && $1_prompt_segment "$0" "$2" "green" "011" "${node_version:1}" 'NODE_ICON'
```
dritter: Would that work? `node_version` is defined in the if statement as a local variable..

rjorgenson:
I could be wrong, but I thought it was local to the function. I can rework it to just return if it doesn't see nvm. That might actually be a better solution either way.

rjorgenson:
Looks much cleaner to me, what do you think @dritter

dritter:
:+1: Looks great! :)
Thank you very much.

bhilburn:
Solid contribution, @rjorgenson! Thanks for continuing to help make powerlevel9k even more awesome `=)`

PR #193, opened 2016-01-16
fixing segment call for new state implementation in prompt_load

rjorgenson:
The changes to the new method of tracking state broke custom settings for the prompt_load segment. Without this fix it needs ```POWERLEVEL9K_LOAD_STATEBACKGROUND``` instead of ```POWERLEVEL9K_LOAD_STATE_BACKGROUND``` as with the previous version.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -519,7 +519,7 @@ prompt_load() {
     current_state="normal"
   fi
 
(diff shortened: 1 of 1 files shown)


dritter:
Excellent catch! :+1:

bhilburn:
@rjorgenson - Thanks for catching this and fixing it! Very, very helpful.

PR #195, opened 2016-01-17
Minor improvements

dritter:
Just clarified the README section about the `dir` segment and added the visual identifier concept to the `icons_test` segment.


bhilburn:
Nice.

Issue #186, opened 2016-01-12
joined status segment breaks transition to previous segment


depressiveRobot:
I switched to the new joined segments feature consisting of `background_jobs`, `root_indicator` and `status` segment introduced in #164.

However, using this feature breaks the transition to the previous segment:
```
POWERLEVEL9K_STATUS_VERBOSE=false
POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(time background_jobs root_indicator_joined status_joined dir vcs)
```
![screen shot 2016-01-12 at 14 59 27](https://cloud.githubusercontent.com/assets/6043497/12265554/f6bb67a8-b93d-11e5-8c98-a16697e3586a.png)

Adding the joined segment to the front seems to work:
```
POWERLEVEL9K_STATUS_VERBOSE=false
POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(background_jobs root_indicator_joined status_joined time dir vcs)
```
![screen shot 2016-01-12 at 14 58 50](https://cloud.githubusercontent.com/assets/6043497/12265596/2b8020c8-b93e-11e5-9c92-589dc2daae10.png)

And using the status segment without joining works, too:
```
POWERLEVEL9K_STATUS_VERBOSE=false
POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(time status dir vcs)
```
![screen shot 2016-01-12 at 15 03 18](https://cloud.githubusercontent.com/assets/6043497/12265614/49f0cc60-b93e-11e5-8def-86f0f02ebe93.png)

dritter:
Joining two conditional segments together isn't supported yet.

To be more technical:
If you look into the source, we have four major places which get passed by every segment: `left_prompt_segment` and `build_left_prompt` (the other two are the right ones). So these functions are potentially candidates for checking if a segment should be a joined one or not.

##### `left_prompt_segment`: 
We print the segments by calling the `left_prompt_segment` function in which we don't have knowledge about the last printed segment. So we cannot check if the last one met it's condition to be printed.

##### `build_left_prompt`:
This is a much better place, as we check here if the user wants the segment to be joined. But the condition is within the segment itself, so we have to check if the previous rendered segment had output or not. This could work to transform the joined segment into a real one.
But IMHO this code should be scheduled in the future (would be quite a change; more than just a bugfix).

I'd be happy if you want to submit a pull request. ;)

bhilburn:
[...]

PR #202, opened 2016-01-22
Join conditional segments

dritter:
This PR fixes the joining of conditional segments (#186). It works even with multiple conditional segments.


bhilburn:
Looking good. Pulling this in to `next`.

PR #205, opened 2016-02-03
Change default dir truncation delimitor to ellipsis

dritter:


bhilburn:
@dritter - Is that code point supported in all Powerline fonts, or does it require 'Awesome' patched fonts? Will this work for 'compatible' mode?

dritter:
@bhilburn It should work with all fonts. This is a [standard unicode sign](http://unicode-search.net/unicode-namesearch.pl?term=%E2%80%A6&.submit=Search). I double-checked it with Arial.

bhilburn:
@dritter - Lovely! Merged.

PR #208, opened 2016-02-06
Fix joining segments

dritter:
My Fix for #186 was incomplete. Now it works correctly.


bhilburn:
Thanks for the hotfix!

PR #212, opened 2016-02-09
Add Unit tests

dritter:
This PR adds unit testing to this project. For now, it tests just "joining of segments" and "dynamic coloring of segments", as this is just a proof-of-concept. I we decide this PR is a good idea, then I'll add more tests..

### Why?
This theme is very flexible. As a result every change needs a lot of testing. At least I tend to forget certain tests, so the theme gets error prone.

### Hey! [The Build](https://travis-ci.org/dritter/powerlevel9k) fails!
Yep. Intentionally. :smile:
I branched off of 9e191d43058d66fd52d1c33a244523600848bd25 (right before e9da3d5146d5cf881249fa229e27b60428516aab), where joining of Segments is broken, so `testJoinedSegments` and `testTransitiveJoinedSegments` fail.

### How?
This PR adds [shunit2](https://github.com/kward/shunit2) as a git submodule and execute the unittests in `test/powerlevel9k.spec` with [Travis-CI](https://travis-ci.org). Btw. `shunit2` must not be a git submodule, we could install it directly on travis..
To execute the tests locally, just run `./test/powerlevel9k.spec`. The test have a hardcoded path to `shunit2´ and powerlevel9k, so they need to be executed from the root folder of the project.

### Benefits
Well, once enabled every PR gets tested, so every developer gets nice feedback. And we can add a nice badge to the README, so every user can easily see the build status of the project.

@bhilburn You have to enable this project for [Travis-ci](https://travis-ci.org/profile/bhilburn).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "shunit2"]
+	path = shunit2
+	url = git@github.com:kward/shunit2.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "shunit2"]
 	path = shunit2
-	url = git@github.com:kward/shunit2.git
+	url = https://github.com/kward/shunit2.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,4 +21,5 @@ script:
   - test/powerlevel9k.spec
   - test/functions/utilities.spec
   - test/functions/colors.spec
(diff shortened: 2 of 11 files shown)


bhilburn:
@dritter - I'm *so* excited you added this. I've been thinking about this for a while, actually. Thanks so much for putting it together - this is a really great addition to the project!

PR #214, opened 2016-02-10
Fix vcs foreground color

### LINKED ISSUE #213: VCS_MODIFIED_FOREGROUND not applied to text
I have put the following configuration in my zshrc:

POWERLEVEL9K_VCS_MODIFIED_FOREGROUND='red'

It seems the red color is only applied to the icon of the segment, but not to the part after it.
A picture might make this clearer:
selection_008

I would expect the entire foreground to be modified.

dritter:
The VCS segment defined on every piece the foreground color itself. That was from a time when I wanted to emphasize more relevant information by a stronger color. Other informations needed therefore to be displayed in weaker colors (like "dark grey" instead of "black").
Here I removed this completely, so the dynamic overriding mechanism can take place.

This fixes #213. Thx @bartlibert for the report.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -99,9 +99,6 @@ else
   DEFAULT_COLOR_DARK="236"
 fi
 
-set_default POWERLEVEL9K_VCS_FOREGROUND "$DEFAULT_COLOR"
-set_default POWERLEVEL9K_VCS_DARK_FOREGROUND "$DEFAULT_COLOR_DARK"
-
 ################################################################
 # Prompt Segment Constructors
 #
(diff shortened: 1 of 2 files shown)


powerlevel9k.zsh-theme
```diff
@@ -752,6 +747,15 @@ prompt_vcs() {
   VCS_WORKDIR_DIRTY=false
   VCS_WORKDIR_HALF_DIRTY=false
 
+  # The vcs segment can have four different states - defaults to ''.
```
bhilburn: What's the fourth state?

bhilburn: What's the fourth state?

bhilburn:
I like this change a lot, and I also like your design for addressing it, @dritter.

I have one question about the state map (three vs four states). Once we get that resolved, I can address #217 on top of these changes.

dritter:
@bhilburn Done. Corrected that comment. :)
I had four states in mind thinking that the `action` (like rebasing, merging) is a state of its own.

Issue #215, opened 2016-02-12
rbenv issues


MatthewMDavis:
First, a documentation issue: the [link to rbenv information on the repo main page](https://github.com/bhilburn/powerlevel9k#rbenv) doesn't go anywhere.

Second, I'm trying to understand why the rbenv segment works the way it does. The segment tests for the RBENV_VERSION environment variable, which only returns true if one has set a *shell-specific* rbenv version -- one that trumps the global and local settings. So if I'm in a directory that's under a local rbenv version, the prompt won't tell me that.

Obviously, if I **do** have a shell-specific rbenv version in effect, I'll want my shell to tell me that -- what I want the prompt to tell me is what version of Ruby is in effect, so I'll want to know that my local version is preempted by a shell-specific version. But since I **don't**, I'm not clear on why this segment is keying on this environment variable, rather than testing to see whether there is an rbenv version *in effect* right now.

Should I be rigging something up to set this variable equal to the output of 'rbenv version' if I want to have my prompt tell me what version of Ruby is in effect?

dritter:
> Should I be rigging something up to set this variable equal to the output of 'rbenv version' if I want to have my prompt tell me what version of Ruby is in effect?

Ideally no. But maybe you want to improve the segment and share it with us all? :wink: 

I am not a ruby guy, so I cannot help explain what is happening (or understand you correctly with all that shell-specific rbenv-versions). But maybe @natemccurdy can help?

natemccurdy:
@MatthewMDavis Hi, I'm the one who wrote the original rbenv segment, so maybe I can help.
Firstly, I wrote the segment for my use-case (which I think is pretty standard based on what I've seen online), so if you use rbenv differently from me, you may likely not be seeing the info you want.

**My Usage**
I use rbenv on my Mac when I want to test writing Ruby code in different versions than my system Ruby. To do that, I run `rbenv shell <some_version>` which creates an environmental variable `$RBENV_VERSION` whose value is the version of Ruby that `rbenv` is currently using.

When I stop testing, I run `rbenv shell --unset` which resets the version of Ruby being used to system Ruby. In that process, the environmental variable is unset.

Here's a screenshot that shows that process:
<img width="822" alt="rbenv" src="https://cloud.githubusercontent.com/assets/10200384/13018571/20ba4b92-d17f-11e5-8c9d-9895013ca8fd.png">

[...]

PR #216, opened 2016-02-12
Improve the rbenv prompt by using `version-name`

natemccurdy:
Prior to this, the rbenv prompt only showed something if the
`$RBENV_VERSION` environment variable was set. This was not a complete
solution because rbenv can be configured locally, per directory, with
dotfiles. When using dotfiles, the `$RBENV_VERSION` variable is not set.

This fixes the issue by taking the output of the `rbenv version-name`
command which shows the real rbenv version being used based on any of
the 4 ways to change rbenv as shown here:

https://github.com/rbenv/rbenv#choosing-the-ruby-version

This commit also adds documentation to the README that describes the prompt.

Fixes issue #215 

Thanks to @MatthewMDavis for the original version of the patch: https://github.com/MatthewMDavis/powerlevel9k/commit/96a8408a0c27ed6dc10dfa8829a549419ab5d4ec

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,15 @@ specify the correct network interface by setting:
 |----------|---------------|-------------|
 |`POWERLEVEL9K_IP_INTERFACE`|None|The NIC for which you wish to display the IP address. Example: `eth0`.|
 
+##### rbenv
+
+This segment shows the version of Ruby being used when using `rbenv` to change your current Ruby stack.
+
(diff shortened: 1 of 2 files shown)


bhilburn:
@natemccurdy - Regardless of what comes out of the #119 discussion, this is a great improvement to the existing functionality.

Thanks so much for another solid contribution! Very nice.

PR #227, opened 2016-03-04
Add chruby prompt

wadkar:
This is [WIP] with pending external testing and documentation update.

Also, see postmodern/chruby/issues/245 for why the prompt doesn't get updated on **automagic** switching of ruby versions based on `.ruby_version`.

If this gets :+1: I will update the README.md

Partially closes #119 (?)

PR #228, opened 2016-03-04
Add chruby prompt

wadkar:
New PR based off `origin:next` instead of `origin:master` see #227 

Add new `chruby` prompt to show ruby version using `chruby`.
Update the README accordingly.

See postmodern/chruby#245 for issue with auto switching ruby version using `.ruby_version` file.
Partially closes #119 , replacement for #227.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,8 @@ The segments that are currently available are:
 * **os_icon** - Display a nice little icon, depending on your operating system.
 * **php_version** - Show the current PHP version.
 * **ram** - Show free RAM
-* [rbenv](#rbenv) - Ruby environment information (if one is active).
+* [rbenv](#rbenv) - Ruby environment information using `rbenv` (if one is active).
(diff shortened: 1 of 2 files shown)


wadkar:
This is unrelated, but can I safely use this `origin:next` branch for my production machine? I am not asking for any milestone plans but would be nice to know what branch should I put in my `zgen` config

dritter:
Cool! Thank you very much.

IMHO it is safe to use the next branch on production. We try to keep it as stable as possible, but it may happen that once in a while a feature does not work. Being a theme displaying certain things won't work - no data will be lost (or similar)..

wadkar:
okay, i am using `next` on my machine
what can i do to help close this PR?

bhilburn:
Thanks for the great contribution, @wadkar! Very much appreciate, and happy to have this upstream =)

PR #229, opened 2016-03-06
Add shortening strategy based on package.json name

alexlafroscia:
Finally got around to re-implementing #140 without using the `jq` tool, and against the current `next` branch so it can be pulled in easily.

I'm happy to pull the comments out if you'd prefer; I just left them in now because they were in my original implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,7 @@ Customizations available are:
 | Variable | Default Value | Description |
 |----------|---------------|-------------|
 |`POWERLEVEL9K_SHORTEN_DIR_LENGTH`|`2`|If your shorten strategy, below, is entire directories, this field determines how many directories to leave at the end. If your shorten strategy is by character count, this field determines how many characters to allow per directory string.|
-|`POWERLEVEL9K_SHORTEN_STRATEGY`|None|How the directory strings should be truncated. By default, it will truncate whole directories. Other options are `truncate_middle`, which leaves the start and end of the directory strings, and `truncate_from_right`, which cuts starting from the end of the string.|
+|`POWERLEVEL9K_SHORTEN_STRATEGY`|None|How the directory strings should be truncated. By default, it will truncate whole directories. Other options are `truncate_middle`, which leaves the start and end of the directory strings, and `truncate_from_right`, which cuts starting from the end of the string.  You can also use `truncate_with_package_name` to use the `package.json` `name` field to abbreviate the directory path.|
 |`POWERLEVEL9K_SHORTEN_DELIMITER`|`..`|Delimiter to use in truncated strings. This can be any string you choose, including an empty string if you wish to have no delimiter.|
 
 For example, if you wanted the truncation behavior of the `fish` shell, which
@@ -256,6 +256,15 @@ In each case you have to specify the length you want to shorten the directory
 to. So in some cases `POWERLEVEL9K_SHORTEN_DIR_LENGTH` means characters, in
 others whole directories.
 
+The `truncate_with_package_name` strategy gives your directory path relative to the root of your project.  For example, if you have a project inside `$HOME/projects/my-project` with a `package.json` that looks like:
+
+```json
+{
+  "name": "my-cool-project"
+}
+```
+
+the path shown would be `my-cool-project`.  If you navigate to `$HOME/projects/my-project/src`, then the path shown would be `my-cool-project/src`.  Please note that this currently looks for `.git` directory to determine the root of the project.
 
 ##### ip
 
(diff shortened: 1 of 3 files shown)


alexlafroscia:
Unrelated: it seems like the prompt loads a bit more slowly than the version I was using before (which you can find [here](https://github.com/alexlafroscia/powerlevel9k/tree/add-package-name-shortener)).  Have you noticed this too?

dritter:
Cool thing!

About the speed issues:
The theme got a lot more complex since then. We split it up into several files, added the joining segments mechanism, etc.. Another reason could be the rework of the rbenv segment (it now invokes `rbenv` instead of checking an environment variable).

alexlafroscia:
Ahh, I see. I did have to change to the `rbenv` segment instead of the more-generic `ruby_version` one.  I'll see if disabling that makes a considerable difference.

dritter:
Yep. Unfortunately, `rbenv` does quite a bit on discovering the ruby version (like going up all directories to search for special files, etc.). See #216 for more infos about that.

alexlafroscia:
Thanks :) Any feedback on the PR?

dritter:
Looks good and works fine :+1: . I tested this under plain ZSH, Antigen and on OSX.

Only thing is, that the structure of the whole dir segment isn't the best at the moment, and I think we should do a cleanup-PR in the future for that. What I mean by this is that a user has to set `POWERLEVEL9K_SHORTEN_STRATEGY` (whyever this variable does not contain the segments name) _and_ `POWERLEVEL9K_SHORTEN_DIR_LENGTH` even though this variable is never used by the new `truncate_with_package_name`. But as said, I think this is something for a future PR..

Or on second thought:
Wouldn't it be nice to use the `POWERLEVEL9K_SHORTEN_DIR_LENGTH` to do the same as `truncate_from_right`? So the segment would always show the package name and a path (if variable is set to zero, then show full path)..
In this case, it wouldn't be so confusing that you are required to set an unused variable..

alexlafroscia:
Thanks for testing it out!

> Wouldn't it be nice to use the `POWERLEVEL9K_SHORTEN_DIR_LENGTH` to do the same as `truncate_from_right`?

I'm not sure what you mean, but that's probably because I'm just not 100% up-to-date on how the configuration works right now.

dritter:
Okay. I'll try to explain more thoroughly.

The `dir` segment offers different ways of truncate long directory names. One shortens from the right side, one from left, and one from the middle (they are not very consistent in how the do that, but that's another story). 

You added `truncate_with_package_name`, which is kind of truncation. Kind of, because it replaces the first part of the path, but does nothing if you are in a deep subfolder. So you don't use the [mandatory](https://github.com/bhilburn/powerlevel9k/blob/d0e9baf45390ed8b4d9ae3ca007da526c4eb4f12/powerlevel9k.zsh-theme#L438) variable `POWERLEVEL9K_SHORTEN_DIR_LENGTH`, wich makes it necessary to either refactor the segment, or to make use of that variable..

This is what I suggest: truncate the right part (subfolder part) like the `truncate_from_right` strategy. If  `POWERLEVEL9K_SHORTEN_DIR_LENGTH` is not set, do no truncation. 
This way we get a path like this (assuming we are in `~/test/1/12/123/1234/12345/123456`, and `SHORTEN_DIR_LENGTH` is set to 2, and there is a `package.json` in `~/test` with `name` "XX"): `XX/1/12/12…/12…/12…/123456`.

I hope that made it clearer what I meant. :wink:

alexlafroscia:
Ahh, okay! I can do that. I wrote this change a long time ago -- originally for Agnoster, ported it to Powerlevel before I even knew about shortening strategies (maybe before the were implemented?) and eventually ported that to a real strategy, as seen in this PR.

Your recommendation makes perfect sense, I'll update this PR!

alexlafroscia:
Alright, so I factored the `sed` operation to shorten the path out into a function, and apply that to the path _after_ the `package.json` name.  If `POWERLEVEL9K_SHORTEN_DIR_LENGTH` is `true` then it keeps the same behavior as before; if it is a number, it will shorten each segment in the path to that amount of characters plus the `...`.

bhilburn:
@alexlafroscia - Thanks so much for coming back to this! It looks like a really cool feature.

Do we need to make any adjustments to the README regarding the settings for truncation as part of this merge?

@dritter - Do the CI tests for this test the three different truncation strategies? Might be good to make sure we are doing that, at this point, since this code is fairly complex.

alexlafroscia:
Oh, I totally forgot to update the README.  I'll add that in.

bhilburn:
@alexlafroscia - Sorry it's taken me a while to get back to this - I've been traveling for the last few weeks, and am just now catching up.

Your update to the README looks good, and the change overall is a fantastic contribution. Thanks so much for taking the time to put this together. I'm excited it's going upstream `=)`

alexlafroscia:
I am too!

PR #230, opened 2016-03-06
Add a new segment for current active docker machine

sabricot:
I don't know if you are interested in adding new simple segments? (especially since we can add new custom segments easily).
But in case, this PR add a new segment for display the current active Docker Machine.


dritter:
Yes, we are definitely interested in merging useful segments back to the mainline. Thank you very much for your contribution. :wink:

wadkar:
sweet! :+1:

bhilburn:
@wadkar - Whoa, awesome! Love the new segment.

Thanks for updating the README, too! Great form.

PR #233, opened 2016-03-15
Fix NVM prompt to return if not used

wadkar:
This commit changes nvm prompt setup to return immediately if no node
version is installed/available.
Currently, prompt shows incorrect global node version _one_
Specifically, if you install nvm (from brew or manually) and do not have
global node installed (or set), then the prompt shows `${node_version:1}`
which evaluates to the string `one`. This is incorrect, and nvm should not
show anything. This commit ensures this by immediately returning if
`nvm current` returns "none".

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -560,6 +560,7 @@ prompt_node_version() {
 prompt_nvm() {
   [[ ! $(type nvm) =~ 'nvm is a shell function'* ]] && return
   local node_version=$(nvm current)
+  [[ ${node_version} = "none" ]] && return
   local nvm_default=$(cat $NVM_DIR/alias/default)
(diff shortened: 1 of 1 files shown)


dritter:
Good idea!

Could you change it, so that it is just one line?
Like here (I just merged your check with the check for an empty `node_version`):
```zsh
  local node_version=$(nvm current)
  [[ ${node_version} == "none" ]] || [[ -z "${node_version}" ]] && return
  
  local nvm_default=$(cat $NVM_DIR/alias/default)
  [[ "$node_version" =~ "$nvm_default" ]] && return
```

That would be much clearer.

wadkar:
Done

bhilburn:
Great fix, @wadkar! Thanks so much for submitting!

Excellent contribution.

PR #234, opened 2016-03-15
Git tag not replacing branch

nmaggioni:
Having the git tag replace the branch name can lead to some issues when pushing - _pushing to master instead of development_, anyone? - I think that the former shouldn't shadow the latter.
With this small modification both characteristics will be displayed at the same time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,8 +103,8 @@ The segments that are currently available are:
 * [rspec_stats](#rspec_stats) - Show a ratio of test classes vs code classes for RSpec.
 * **rust_version** - Display the current rust version.
 * [status](#status) - The return code of the previous command.
-* [symphony2_tests](#symphony2_tests) - Show a ratio of test classes vs code classes for Symfony2.
-* **symphony2_version** - Show the current Symfony2 version, if you are in a Symfony2-Project dir.
+* [symfony2_tests](#symfony2_tests) - Show a ratio of test classes vs code classes for Symfony2.
+* **symfony2_version** - Show the current Symfony2 version, if you are in a Symfony2-Project dir.
 * [time](#time) - System time.
 * [todo](http://todotxt.com/) - Shows the number of tasks in your todo.txt tasks file.
 * [vi_mode](#vi_mode)- Vi editing mode (NORMAL|INSERT).
@@ -262,7 +262,7 @@ This segment shows the return code of the last command.
 |----------|---------------|-------------|
 |`POWERLEVEL9K_RAM_ELEMENTS`|Both|Specify `ram_free` or `swap_used` to only show one or the other rather than both.|
 
-##### symphony2_tests
+##### symfony2_tests
 
 See [Unit Test Ratios](#unit-test-ratios), below.
 


bhilburn:
@nmaggioni - Thanks for the PR! Just to make sure I understand, with this change, if you are on a tag, the segment will show both the tag and branch name?

@dritter - This changeset seems pretty simple. Any idea why it's failing CI?

nmaggioni:
@bhilburn Exactly. Here's a visual example;
+ Before: schermata da 2016-03-16 21-34-39
+ After: schermata da 2016-03-16 21-32-53

dritter:
@bhilburn Yep. CI fails, because travis tries to find a configuration file on the branch, but the branch was made from `master`.. CI was added only on `next`.

dritter:
@nmaggioni Cool change, btw. :wink:

nmaggioni:
@dritter thanks :)

bhilburn:
@nmaggioni - I like this change a lot! Really great idea. It's not ready for merge, just yet, though. I think with just a bit of work we can get it in, though. Couple of items:

1) This is new development, so it should go on `next`. You asked about which one it should go into in your original post. For this project, `master` is our stable branch, because it's the branch that most git users will check out by default when they pull this project. We want them to always have a stable experience, so we keep new work out of master until it has been merged into `next` and tested as part of a full release.

Basically, new development goes into `next`, and once we know it's stable, we release it to a tag on master.

2) I went ahead and merged it into `next`. There is a merge conflict due to some changes in `functions/vcs.zsh`. It's an esay resolve, but this is also why it's smart to base your new work off of `next`.

3) Unfortunately, even once the merge conflict is resolved, there are two problems:

Firstly, if the user doesn't have awesome-fonts turned on, then there is no icon for a tag name, which means you just get to strings of text. Secondly, if you are in a DETACHED_HEAD state, it actually just prints the tag name twice (because, there, the tag name is technically the 'branch'). 

You can see both issues in this screenshot:

tag-branch

**Steps Forward**

This is a really great PR, and I love your idea, @nmaggioni. Here's what I think needs to be done to make this mergeable:

1) I think it would be good to pick a non-awesome icon to indicate tags. @dritter might be able to provide some input, here.
2) We need to decide what to do in DETACHED_HEAD states. Printing the tag name twice is a bit silly. I think it makes sense, though, to print the *hash* and the tag name.
3) Close out this PR and submit the new work against `next` in a new PR.

Thanks so much for spending time on making a contribution, @nmaggioni. I can't wait to see this upstream =)

nmaggioni:
Here it is! Forgive me for the mistake, I didn't think of the `next` branch as a development one. Have you ever thought about placing a [**CONTRIBUTING.md** file][1] in the root of the repo?

[1]: https://github.com/blog/1184-contributing-guidelines

bhilburn:
@nmaggioni - Great idea! I'll make sure that's in before the next release.

Thanks for the suggestion, @nmaggioni =)

nmaggioni:
Glad to be contributing! :+1:

PR #238, opened 2016-03-22
Git tag not replacing branch

nmaggioni:
Copy of #234 on right branch.

- [x] Move request on `next` branch.
- [x] Hide tag in DETACHED_HEAD states.
- [x] ~~Show hash in DETACHED_HEAD states.~~ (Already handled by default implementation)
- [x] ~~Handle non-Awesome icons.~~ (Already handled by [an internal function][1])
- [x] Better DETACHED_HEAD state detection.
- [x] Special cases handling

[1]: https://github.com/nmaggioni/powerlevel9k/blob/next/functions/icons.zsh#L201

### DIFF
--- a/functions/vcs.zsh
+++ b/functions/vcs.zsh
@@ -67,8 +67,12 @@ function +vi-git-tagname() {
     local tag
 
     tag=$(git describe --tags --exact-match HEAD 2>/dev/null)
-    # Append the tag segment to the branch one
-    [[ -n "${tag}" ]] && hook_com[branch]+=" $(print_icon 'VCS_TAG_ICON')${tag}"
+    head=$(git describe --all)
+    # Make sure that detached head and tag differ in name
+    if [[ "${head}" != "${tag}" ]]; then
+      # Append the tag segment to the branch one
+      [[ -n "${tag}" ]] && hook_com[branch]+=" $(print_icon 'VCS_TAG_ICON')${tag}"
+    fi
   fi
 }
 
--- a/functions/vcs.zsh
+++ b/functions/vcs.zsh
@@ -62,16 +62,20 @@ function +vi-git-remotebranch() {
 
 function +vi-git-tagname() {
   # Only show the tag name if we are not in DETACHED_HEAD state,
-  # since in that case it would already be displayed in the branch segment
-  if [[ -z "$(git symbolic-ref HEAD 2>/dev/null)" ]] ; then
-    local tag
+  # or if the current branch's HEAD is the same commit as a tag but
+  # doesn't have the same name
+  local tag
+  tag=$(git describe --tags --exact-match HEAD 2>/dev/null)
 
-    tag=$(git describe --tags --exact-match HEAD 2>/dev/null)
+  if [[ -z "$(git symbolic-ref HEAD 2>/dev/null)" || ! -z "${tag}" ]] ; then
     head=$(git describe --all)
-    # Make sure that detached head and tag differ in name
-    if [[ "${head}" != "${tag}" ]]; then
-      # Append the tag segment to the branch one
-      [[ -n "${tag}" ]] && hook_com[branch]+=" $(print_icon 'VCS_TAG_ICON')${tag}"
+    # Make sure that detached head or checked out name differs from tag name
+    if [[ "${head}" != "${tag}" ||
+      "$(git rev-parse --abbrev-ref HEAD)" != "${tag}" &&
+      "$(git rev-parse --abbrev-ref HEAD)" != "HEAD" &&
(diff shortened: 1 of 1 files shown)


dritter:
To answer your last question ("Handle non-Awesome icons): Yes, they are already handled in the `print_icon` function (as you figured out yourself; just want to give some explanation).
The Icons just have to be defined for each `POWERLEVEL9K_MODE` in the [icons](https://github.com/bhilburn/powerlevel9k/blob/2d3131fb8ab712d169539f8e139e7899aa6d014a/functions/icons.zsh#L15-L16) array, and the `print_icon` function prints the right one. So no worries there. :)

nmaggioni:
Awesome :)
Is there anything else that needs to be refined? I have the feeling that my method of detecting a detached head is quite a dirty hack, but scrolling through the theme's source didn't give me any other ideas on how to do that. Suggestions are welcome!

bhilburn:
@nmaggioni - Awesome updates!

For detecting the detached HEAD state, this looks like a cleaner way to go about it:
http://stackoverflow.com/a/1593487

Based on my reading of that, if `"$(git symbolic-ref HEAD 2>/dev/null)"` gives you an empty string, you are in DETACHED_HEAD.

I'm really excited to get this in =)

nmaggioni:
@bhilburn Great find, it looks way cleaner this way!

bhilburn:
Hey @nmaggioni -

Sorry it's been a few weeks on my end. I've been traveling heavily. I'm home, now, though, and eager to get this upstream!! It's an awesome contribution, and I'm really excited about it.

Unfortunately, the latest copy still isn't working for me. Specifically, I'm seeing the following:
* When in DETACHED_HEAD state, it still prints twice.
* When I'm on the same commit as a tag, but on the HEAD of a branch, it actually now only prints the branch name instead of both of them?

You can see both of these issues in the screenshot, below:

state

I'm wondering if there is something specific about my set-up that's preventing this from working correctly. Does this work for you, @dritter?

By the way, I do *not* use Awesome Fonts - just normal Powerline. I wonder if that's where the problem is?

nmaggioni:
@bhilburn I didn't include those cases in my tests, my bad.

Now they should be handled; the code is maybe a bit... _convoluted_, but it gets the job done - it is quite difficult to handle all the possible combinations in a clean way.

Could you confirm that the patch is ok?

bhilburn:
Okay, I started working on this, and realized it would be really cool if we took it one step further: if we are in a detached head state, we show the hash. This way, if you are sitting on a tag, you either see the branchname or hash for the tag. 

Unfortunately, the way ZSH `vcs_info` stores the hash information made this difficult to achieve. From what I could tell, it doesn't get loaded into the `hook_com` array, which means I had to grab it manually and clobber the `hook_com[branch]` string.

Another problem: I could *not* for the life of me get it to properly reference  `VCS_CHANGESET_LENGTH` in the function. As a result, it's hard-coded to `8` right now.

My changes are here in the branch `vcs-tag-display`, and include all of @nmaggioni's change history:
https://github.com/bhilburn/powerlevel9k/tree/vcs-tag-display

@nmaggioni - Can you test that branch and make sure it works?

@dritter - I don't want to merge this without your testing & :+1:, too. Also, if you have any thoughts on fixing the two issues I mentioned above (hook_com and hard-coded hash length), that would be great!

dritter:
@bhilburn I'll have a look on that at the weekend. I'm short on time for this week.

nmaggioni:
@bhilburn Great idea, I like it. By confronting the branch with some casual work and your test case above everything seems ok :+1: .
With all these changes, IMHO, the Git workflow with the theme has improved noticeably.

dritter:
Okay. I started testing this, but for me it is still unclear which behavior we want.

IMHO there are these different cases a git repository can have:

- **Detached Head** The user did an checkout of a specific commit which is not a tag. This is handled by the `vcs-git-remotebranch` function.

  Here the changeset does not get cut off, because VCS_INFO (or git itself, idk) sets the branch name to the changeset (truncated automatically). I am not sure, if we can handle this one (this is the one I am unsure about)..

- **Detached Tag** The user did an checkout of a tag, which is not HEAD. This is handled by the `vcs-git-tagname` function.


Besides that, the user can show the tag, if he wants to. This can lead to an prompt like this:


I did some fixes so that the changeset gets truncated properly in the "detached tag" state.

bhilburn:
@dritter - I think your breakdown is consistent with my desired behavior. And thanks so much for fixing the hash length! I'm going to go ahead and get the branch merged, now `=)`

bhilburn:
Annnd it's upstream! Thanks so much for the great ideas and work, @nmaggioni, and for the excellent testing & fixes, @dritter!

nmaggioni:
:tada:

PR #242, opened 2016-04-03
Fix configurability of actionformat color

dritter:


bhilburn:
Hah, nice, @dritter. Simple, but very very useful commit. Thank you `=)`

PR #247, opened 2016-04-16
rust icon added

orhanbalci:
Here is my screenshot for rust icon

developer_rust_icon

### DIFF
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -122,7 +122,7 @@ case $POWERLEVEL9K_MODE in
       VCS_GIT_ICON                   $'\UF113 '             # 
       VCS_HG_ICON                    $'\UF0C3 '             # 
       RUST_ICON                      $'\UE6A8'              #  
-	  )
+    )
   ;;
   *)
     # Powerline-Patched Font required!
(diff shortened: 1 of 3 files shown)


dritter:
Cool. Thx. Looks awesome 👍 

I think we should set the Image for `awesome-patched` mode to an empty string as well. Being able to use the original code point from devicons is the advantage of fontconfig, and is not possible on the other modes..

orhanbalci:
Set empty icon fore `awesome-patched` mode.

bhilburn:
@orhanbalci - Thanks so much for the great PR! We are really excited to get this upstream.

Two requests:
1. Can you add a commit that adds the Rust icon to the README so that users will know about it?
2. It looks like you may have inadvertently introduced whitespace on one line. I'll mark it, below:

functions/icons.zsh
```diff
@@ -120,7 +121,8 @@ case $POWERLEVEL9K_MODE in
       VCS_REMOTE_BRANCH_ICON         ' '$'\UF204 '          # 
       VCS_GIT_ICON                   $'\UF113 '             # 
       VCS_HG_ICON                    $'\UF0C3 '             # 
-    )
+      RUST_ICON                      $'\UE6A8'              #  
+	  )
```
bhilburn: Can you bring this end-paren back to it's original spot? Thanks!

bhilburn:
@orhanbalci - Looks awesome!! Merged! `=)` Thanks for the great contribution.

orhanbalci:
@bhilburn I added a link to rust icon in README and fixed the whitespace

orhanbalci:
@bhilburn thanks for feedback =)

PR #249, opened 2016-04-19
Add anaconda segment.

jaermanx9:
Shows the currently active anaconda environment different from root.
Requires `ack` or a recent version of `grep` supporting positive lookbehind.


bhilburn:
Thanks so much for the great contribution, @jaermanx9!!

PR #251, opened 2016-04-21
Load warning improvements

rawkode:
Load warnings should be relative to the number of cores and not arbitrary. Also, the variable suggested the 5 minute load was being used, but it was 1 minute. Perhaps we could add this as a configuration option?

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -511,7 +511,7 @@ prompt_ip() {
 prompt_load() {
   # The load segment can have three different states
   local current_state="unknown"
-  local cores=`nproc`
+  local cores=$(nproc)
 
   typeset -AH load_states
   load_states=(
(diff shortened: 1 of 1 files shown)


rawkode:
I don't think this is failing due to my changes ... hopefully :see_no_evil:

dritter:
@rawkode Cool thing!

Don't worry, the tests do not fail because of your implementation. ;) They fail because you branched off of `master`, but the tests just had been introduced on `next` (our development branch, btw.).

Could you rebase your PR against `next`? That would be awesome, and makes it a lot easier to test for us.

Thanks.

dritter:
Btw. `nproc` is linux specific. We need to move that into the OS specific block. On OSX we could use `sysctl -n hw.ncpu` to get the number of cores. See https://coolaj86.com/articles/get-a-count-of-cpu-cores-on-linux-and-os-x/

rawkode:
New PR (#252) to replace this, as I thought it would be better to create a new branch off of next

PR #252, opened 2016-04-22
Improvements to load prompt, now reporting warning and critical at 50…

rawkode:
Load warnings should be relative to the number of cores and not arbitrary. Also, the variable suggested the 5 minute load was being used, but it was 1 minute. Perhaps we could add this as a configuration option?

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -589,7 +589,7 @@ prompt_load() {
 
   if [[ "$OS" == "OSX" ]]; then
     load_avg_1min=$(sysctl vm.loadavg | grep -o -E '[0-9]+(\.|,)[0-9]+' | head -n 1)
-    cores=$(sysctl -n hw.physicalcpu)
+    cores=$(sysctl -n hw.logicalcpu)
   else
     load_avg_1min=$(grep -o "[0-9.]*" /proc/loadavg | head -n 1)
     cores=$(nproc)
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -578,30 +578,39 @@ prompt_ip() {
 prompt_load() {
   # The load segment can have three different states
   local current_state="unknown"
+  local cores
+
+  if [[ "$OS" == "OSX" ]]; then
+    cores=$(sysctl -a | grep machdep.cpu)
+  else
+    cores=$(nproc)
+  fi
+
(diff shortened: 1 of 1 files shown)


powerlevel9k.zsh-theme
```diff
@@ -578,30 +578,39 @@ prompt_ip() {
 prompt_load() {
   # The load segment can have three different states
   local current_state="unknown"
+  local cores
+
+  if [[ "$OS" == "OSX" ]]; then
```
dritter: About 10 lines below, there is already an `if` that sets OS specific values. Could you merge yours with this one?

powerlevel9k.zsh-theme
```diff
@@ -578,30 +578,39 @@ prompt_ip() {
...
+    cores=$(sysctl -a | grep machdep.cpu)
```
dritter: `sysctl` gives you the key and value by default. If you just want the value, use `sysctl -n`. And if you pass a key to it, there is no need to grep.
That being said, the key `machdep.cpu` gives me a whole bunch of information about my cpu, not just my amount of cores..
`sysctl -n hw.ncpu` is deprecated, and equivalent to `sysctl -n hw.logicalcpu`.
To get the amount of physical CPUs we could use either `sysctl -n hw.physicalcpu`.

rawkode:
Thanks, @dritter - I didn't have an OSX machine to test on. Updated with your suggestions :+1:

dritter:
Looks good to me. 👍 
A last question: Don't we want to use the number of cores (logical) instead of real processors (physical) on OSX?

@bhilburn I'll test this on Linux next week.

bhilburn:
@rawkode - Great fixes! Thanks so much for spotting the issue and submitting a resolution - this is excellent.

I tested it in Linux, and it appears to be working for me.

I don't have OSX, so I can't test the difference between logical & physical cores that @dritter is asking about, but I think that's the last thing we need to resolve before we merge this upstream.

If we can close this out soon, it'll make it just in time for our release!

rawkode:
@dritter Using logical CPUs, as this reflect what `nproc` shows on Linux and will provide a fairer representation of the load of the machine

dritter:
@rawkode Perfect! Thank you very much.

bhilburn:
This is really great, @rawkode! Thanks so much for the fantastic contribution =)

PR #255, opened 2016-04-29
Fix incorrect background jobs number

cbourgeois:
The jobs -l command displays a "pwd now :..." line when a job was launched from a different directory than the current one that cause powerlevel9k to display an incorrect number of background jobs. This change fixes it.


dritter:
Awesome! Thanks!

Our development branch is `next`. Could you rebase your PR to `next`?  That would be awesome.

@bhilburn Works for me on OSX (with zplug) and Linux (with plain ZSH).

cbourgeois:
Closing and reopening as requested

PR #256, opened 2016-04-29
Fix incorrect background jobs number

cbourgeois:
The jobs -l command displays a "pwd now :..." line when a job was launched from a different directory than the current one that cause powerlevel9k to display an incorrect number of background jobs. This change fixes it.


bhilburn:
Hey @cbourgeois - Great catch, and thanks for submitting a fix! We really appreciate your contribution!

PR #257, opened 2016-05-04
PYTHON ICON for virtualenv prompt

thuandt:
Rust, Ruby icon already support, I'm so susprise why virtualenv still don't have icon.
I get Python icon Unicode from here: https://vorillaz.github.io/devicons/#/cheat

Preview

PR #259, opened 2016-05-14
Add pyenv segment showing active python version

wadkar:
I created a new segment for use with [pyenv](https://github.com/yyuu/pyenv) to
manage python versions. This prompt can be used in place of `virtualenv`
segment, however, it is targeted at users of pyenv and not meant to
replace the `virtualenv` segment.

The prompt works by parsing output of `pyenv version` and displaying the
first word of the output as segment text. The design (color etc.) is
same as the `virtualenv` segment to respect consistency.

Tested on my personal machines (Mac OSX 10.11.4 and Ubuntu 15.04). The
segment would need to be revisited if `pyenv version` changes its output
format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ The segments that are currently available are:
 * **Python Segments:**
     * `virtualenv` - Your Python [VirtualEnv](https://virtualenv.pypa.io/en/latest/).
     * [`anaconda`](#anaconda) - Your active [Anaconda](https://www.continuum.io/why-anaconda) environment.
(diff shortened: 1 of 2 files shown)


wadkar:
Feel free to reword REAME.md changes if necessary. Also, if you're going to accept PR #257 then I would also like to add it to the `pyenv` segment. However, I haven't tested it, and I usually have not so great luck with icons.

Whatever happened to the master plan for fonts, I wonder... anything I can do to help?

wadkar:
Ouch, I meant to base the PR off `next` branch, closing this and creating another PR for that.

PR #260, opened 2016-05-14
PYTHON ICON for virtualenv prompt

thuandt:
Rust, Ruby icon already support, I'm so susprise why virtualenv still don't have icon.
I get Python icon Unicode from here: https://vorillaz.github.io/devicons/#/cheat

Preview

NOTE: This is re-submmit PR of #257


bhilburn:
Looks good! Thanks so much, @thuandt! Nice eye-candy contribution =)

PR #261, opened 2016-05-14
Add pyenv segment to show active non-system python version

wadkar:
I created a new segment for use with pyenv to manage python versions. This prompt can be used in place of `virtualenv` segment, however, it is targeted at users of [pyenv](https://github.com/yyuu/pyenv) and not meant to replace the `virtualenv` segment.

The prompt works by parsing output of `pyenv version` and displaying the first word of the output as segment text. The design (color etc.) is same as the `virtualenv` segment to respect consistency. The segment would need to be revisited if `pyenv version` changes its output format. ~~Furthermore, a small change would be required if and when PR #260 is accepted to include python icon for the segment.~~ The python icon is now added to the prompt.

Tested on my personal machines (Mac OSX 10.11.4 and Ubuntu 15.04). Feel free to reword the README.md changes to make it more, err, _readable_.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@ The segments that are currently available are:
 * **Python Segments:**
     * `virtualenv` - Your Python [VirtualEnv](https://virtualenv.pypa.io/en/latest/).
     * [`anaconda`](#anaconda) - Your active [Anaconda](https://www.continuum.io/why-anaconda) environment.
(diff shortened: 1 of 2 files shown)


bhilburn:
Hey @wadkar - I just merged #260. Can you make the quick update, here?

wadkar:
Done. I didn't get a chance to test the icon.

bhilburn:
@wadkar - Thanks so much! I really appreciate you taking the time to add it to the README, as well :)

Issue #244, opened 2016-04-08
Very slow in home folder because of VCS prompt segment


dj-hedgehog:
I turned my home-folder `~` into a dotfiles repository with a lot of untracked files.
Ever since I made this change zsh is unusable slow while I'm located in the home-folder. It takes like 5-10 seconds to register keystrokes or execute anything.

Since I like the VCS segment I'd like to know what I can do about it without removing it entirely? Can I disable it in certain folders or is there some git function I can use?

jguenther:
What I do is I make a `~/dotfiles` repo and have a makefile in there that symlinks `~/.file` to `~/dotfiles/file`  if they don't already exist. e.g. `~/.bashrc` is a symlink to `~/dotfiles/bashrc`, etc.

dritter:
@dj-hedgehog Thanks for the report.

The way @jguenther suggests would work. But I don't like it that this theme slows your terminal that much down. So I create a repository with 500.000 untracked files, and voilá it is slow as hell.
For now I can say that it has to do with the checking of untracked files. If I deactivate this (currently only available on the `next` branch) by setting `POWERLEVEL9K_VCS_GIT_HOOKS=(vcs-detect-changes git-aheadbehind git-stash git-remotebranch git-tagname)` (notice that I took out `git-untracked`). This is no real solution, as it deactivates the whole checking of untracked files for all repositories.
I need to investigate this further. Just wanted to inform you about my ongoings.

dritter:
Bad news: This is an issue of git itself. My first thought was, to cut off the response of git (with `tail -n 1`), but git computes the whole list of files before handing over to tail.. Second thought was to add an `.gitignore` with all untracked files in it (or a `*`), but git needs to check each file against this mask (which is slow). There seems no way to get just a boolean value out of git to know if there is an untracked file at all.
Even other commands than `git status` are not fast enough (like `git ls-files`, as suggested [here](http://stackoverflow.com/questions/11021287/git-detect-if-there-are-untracked-files-quickly)).

IMHO there are two ways that would work:

1. as @jguenther suggested, clone your dotfiles repository to another location, and link all relevant files to your home directory.

2. disable the untracked files checking in your home directory. To do so, you could add piece of code in your `.zshrc`:
```zsh
if [[ $(print -P "%~") == '~' ]]; then
  POWERLEVEL9K_VCS_GIT_HOOKS=(vcs-detect-changes git-aheadbehind git-stash git-remotebranch git-tagname)
[...]

PR #262, opened 2016-05-16
Async right prompt

andjscott:
This could be a solution to the lag experienced in #244.
An subprocess writes the prompt to a temporary file (unique to each zsh instance) as each element is loaded. A signal is send to the parent zsh process after each write so that the RPROMPT can be set to the new value. The temporary file is deleted when the zsh exits.

The three `zle reset-prompt` were removed from the other handlers as otherwise RPROMPT was reset to empty after hitting enter. However `powerlevel9k_prepare_prompts` handles `zle reset-prompt` so this shouldn't be a problem

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -973,12 +973,13 @@ build_right_prompt() {
     # Check if it is a custom command, otherwise interpet it as
     # a prompt.
     if [[ $element[0,7] =~ "custom_" ]]; then
-      "prompt_custom" "right" "$index" $element[8,-1]
+      "prompt_custom" "right" "$index" $element[8,-1] >> $socket
     else
-      "prompt_$element" "right" "$index"
+      "prompt_$element" "right" "$index" >> $socket
     fi
 
     index=$((index + 1))
+    kill -s USR1 $$
   done
 }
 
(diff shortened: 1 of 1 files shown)


Tritlo:
Please make a pull request against next, that's where new stuff should come in.

That being said, when I try to run this I get:

    remove /var/folders/wr/fk_4qbzn72x34r7xgs173d_40000gn/T/tmp.g0ohipLH? y
    async:1: no such file or directory:
    build_right_prompt:13: no such file or directory:
    build_right_prompt:13: no such file or directory:

powerlevel9k.zsh-theme
```diff
@@ -1006,18 +1015,36 @@ $(print_icon 'MULTILINE_SECOND_PROMPT_PREFIX')"
...
+    if [[ "${ASYNC_PROC}" != 0 ]]; then
+      kill -TERM $ASYNC_PROC >/dev/null 2>&1
+    fi
+    async &!
+    ASYNC_PROC=$!
   fi
 }
 
+tidy() {
+  rm $socket
```
Tritlo: this should be `rm -f`, since some implementations force a confirm otherwise (e.g. os x)

Tritlo:
Alright, I've forked your fork and fixed it for me, see https://github.com/andjscott/powerlevel9k/pull/1. Looks great!

andjscott:
I don't seem to be able to change the target branch of this pull so should I make a new pull request?

PR #266, opened 2016-05-24
Refactor location installation path code

### LINKED ISSUE #264: `Script location could not be found!` on zsh 5.0.7
I have powerlevel9k (via zim) installed fine on my laptop, which runs Arch Linux and zsh 5.2. I'm trying to install it on my Raspberry Pi server, running Raspbian and zsh 5.0.7. However, when I start a new shell I'm presented with `Script location could not be found!`.

I looked in your code, and I think the problem is that `$0` cannot be reassigned on this version of zsh. As a test, I created the following script (`./foo`).
```
#!/usr/bin/env zsh

echo $0
0=bar
echo $0
```
When run on my Arch zsh 5.2, `$0` is reassigned fine.
```
$ ./foo
./foo
bar
```
However, on my Raspbian zsh 5.0.7, it doesn't reassign.
```
$ ./foo
./foo
./foo
```

As a workaround, I modified `.zim/modules/prompt/external-themes/powerlevel9k/powerlevel9k.zsh-theme` to suppress the error, and hardcode the `filename` as follows.

```
--- powerlevel9k.zsh-theme.orig	2016-05-18 16:30:03.356684456 +1000
+++ powerlevel9k.zsh-theme	2016-05-18 16:29:57.466689508 +1000
@@ -55,10 +55,8 @@
 elif [[ -f $0 ]]; then
   # Script is a file
   filename="$0"
-else
-  print -P "%F{red}Script location could not be found!%f"
-  return 1
 fi
+filename="$POWERLEVEL9K_INSTALLATION_PATH"
 script_location="$(dirname $filename)"
 
```

This now loads fine in zsh 5.0.7.

dritter:
Now it is possible to set the installation path by setting `POWERLEVEL9K_INSTALLATION_PATH`. If set, this path will always be used. If it was not set manually, we try to determine the path automatically.

This fixes #264 

@protist could you double-check? Especially with your ZSH 5.0.7. That would be great. Thanks.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -20,43 +20,44 @@
 #zstyle ':vcs_info:*+*:*' debug true
 #set -o xtrace
 
-# Check if the theme was called as a function (e.g., from prezto)
-if [[ $(whence -w prompt_powerlevel9k_setup) =~ "function" ]]; then
+# Try to set the installation path
+if [[ -n "$POWERLEVEL9K_INSTALLATION_PATH" ]]; then
+  # If an installation path was set manually,
+  # it should trump any other location found.
+  # Do nothing. This is all right, as we use the
+  # POWERLEVEL9K_INSTALLATION_PATH for further processing.
+elif [[ $(whence -w prompt_powerlevel9k_setup) =~ "function" ]]; then
+  # Check if the theme was called as a function (e.g., from prezto)
   autoload -U is-at-least
   if is-at-least 5.0.8; then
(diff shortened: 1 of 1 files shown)


protist:
Patching works perfectly on my zsh 5.0.7. Thank you for the quick fix.

bhilburn:
Lots of changes here, not all of which I can easily test. The code looks good, though! Merging into `next` for further dev testing.

Thanks so much, @dritter! Looks like it fixed @protist's issue, as well, which is awesome. Nice work.

Issue #267, opened 2016-05-24
Conditional on custom commands


mateusmedeiros:
I'm currently having to somewhat monkey patch powerlevel9k to add a custom command that I don't want to always be shown:

```bash
prompt_rvm_version() {
  if hash rvm 2>/dev/null; then
    local current_dir="$(pwd)"
    local has_gemfile=false
    while [[ $current_dir != "/" ]]; do
      if [[ -f $current_dir/Gemfile ]]; then
        has_gemfile=true
        break
      fi

      current_dir=$(dirname $current_dir)
    done

    if $has_gemfile; then
		 "$1_prompt_segment" "$0" "$2" "red" "white" "$(rvm-prompt i v p)" ""
    fi
  fi
}

POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(rvm_version context dir vcs)
```
I reached that after reading the [Developer's Guide](https://github.com/bhilburn/powerlevel9k/wiki/Developer's-Guide) and taking a peek at the source code, but it would be cool if we could add a conditional directly to the custom command functionality. Something like:

```bash
has_gemfile_in_current_or_parent_dir() {
  if hash rvm 2>/dev/null; then
    local current_dir="$(pwd)"
    local has_gemfile=false
    while [[ $current_dir != "/" ]]; do
      if [[ -f $current_dir/Gemfile ]]; then
        has_gemfile=true
        break
      fi

      current_dir=$(dirname $current_dir)
    done

    if $has_gemfile; then
		 return 0
    fi
  fi
  return 1
}

POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(custom_rvm_version context dir vcs)

POWERLEVEL9K_CUSTOM_RVM_VERSION="rvm-prompt i v p"
POWERLEVEL9K_CUSTOM_RVM_VERSION_CONDITION="has_gemfile_in_current_or_parent_dir"
POWERLEVEL9K_CUSTOM_RVM_VERSION_BACKGROUND="red"
POWERLEVEL9K_CUSTOM_RVM_VERSION_FOREGROUND="white"
```

I looked around the documentation and source and couldn't find any indication that this was already possible, so if it is, at least some clarification on the docs would be nice.

Since I could already make every possible customization I wanted (kudos to you guys, btw), I guess this is not super essential, but it sure would be nicer and feel safer to do that without having to use any powerlevel9k internal function.

Tritlo:
[...]

PR #268, opened 2016-05-24
Make prompt_custom not print the segment if empty

mateusmedeiros:
Currently you can't make a conditional custom segment. This changes the
prompt_custom function responsible for printing custom segments so that
it won't print the segment in case the content is empty.

With this it becomes possible to make conditional custom prompts simply
by making it not print anything when it's not supposed to be shown.

This is in response to my own issue #267 about a way to make custom segments conditional without having to monkey patch anything. 

I first imagined adding a new var `POWERLEVEL9K_CUSTOM_NAME_CONDITION` with a condition that could be evaluated and used as a conditional, but @Tritlo insightfully suggested that it would be way simpler to just test if the custom command returned an empty string and only print the segment in case it didn't.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -444,8 +444,11 @@ prompt_context() {
 # the output in a segment.
 prompt_custom() {
   local command=POWERLEVEL9K_CUSTOM_$3:u
+  local segment_content="$(eval ${(P)command})"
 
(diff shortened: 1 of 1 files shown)


bhilburn:
Nice work, @mateusmedeiros! Simple, elegant, and very useful! Thanks so much for the great contribution, and thank you, @Tritlo, for providing guidance!

Tritlo:
Happy to help!

mateusmedeiros:
Thanks! Glad I could help. :smile:

PR #273, opened 2016-06-12
Only match first name in package.json

andjscott:
This ensures that the following package.json extract would just give the
prompt "foo" rather than "foo\nbar".

    {
      "name": "foo",
      "author":
      {
          "name": "bar",
      }
    }


andjscott:
I accidentally made PR to master so am closing this to make another one.

PR #274, opened 2016-06-12
Handle shorting path when in the .git directory

andjscott:
`git rev-parse --git-dir` returns just `.` when in the top level of
a`.git` directory. This would cause a problem when extracting the
package_path. Closes #271

PR #275, opened 2016-06-12
Handle shorting path when in the .git directory

### LINKED ISSUE #271: Error when shortening path with package name when in a .git directory
When shortening the path in `prompt_dir` with `truncate_with_package_name` and within a `.git` directory errors are shown:
``` 
prompt_dir:27: substring expression: -3 < 0
prompt_dir:27: substring expression: -3 < 0
```
`git rev-parse --git-dir` returns just `.` when in the top level of a`.git` directory so `package_path=${repo_path:0:-4}` fails.

andjscott:
`git rev-parse --git-dir` returns just `.` when in the top level of
a`.git` directory. This would cause a problem when extracting the
package_path. Closes #271. Fix of PR #274.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -475,6 +475,10 @@ prompt_dir() {
             package_path="."
             subdirectory_path=""
           else
+            # Handle the edge case where $repo_path is '.' due to the current directory being the .git directory.
+            if [[ "$repo_path" == "." ]]; then
+              repo_path=$(pwd)
+            fi
             # If the current path is something else, get the path to the package.json
             # file by finding the repo path and removing the '.git` from the path
             package_path=${repo_path:0:-4}
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -467,30 +467,18 @@ prompt_dir() {
         local name repo_path package_path current_dir zero
 
         # Get the path of the Git repo, which should have the package.json file
-        if repo_path=$(git rev-parse --git-dir 2>/dev/null); then
-          if [[ "$repo_path" == ".git" ]]; then
-            # If the current path is the root of the project, then the package path is
-            # the current directory and we don't want to append anything to represent
-            # the path to a subdirectory
-            package_path="."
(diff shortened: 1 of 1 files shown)


bhilburn:
Hey @andjscott - 

Thanks so much for submitting the fix!

Out of curiosity, does just *always* setting the `repo_path` to `pwd` not work? I'm wondering if that would be cleaner than adding the conditional?

andjscott:
Setting `repo_path=$(pwd)` would not always work as `repo_path` has to be a path to the `.git` directory in the top level of the git directory. For example `/home/andjscott/projects/project/.git` would be expected at the substring step which removes `.git` to get `package_path="/home/andjscott/projects/project/"`.  Using `$(pwd)` unconditionally would get malformed paths e.g. `"/home/andjscott/projects/project/bin/release" => "/home/andjscott/projects/project/bin/rel"`

There seem to be three cases for `--git-dir`. 

1. `$PWD` is the git root => `.git` (relative path)
2. `$PWD` is within a git directory => absolute path ending in `.git`
3. `$PWD` is the .git directory => `.` (relative path)

Alternative would be to use `--show-toplevel` instead but would need a check for `--is-inside-work-tree` for cases 1 and 2 and `--is-inside-git-dir` for case 3.
```
 --git-dir
	Show $GIT_DIR if defined. Otherwise show the path to the .git directory. The path shown, when
	relative, is relative to the current working directory.

	If $GIT_DIR is not defined and the current directory is not detected to lie in a Git repository
    or work tree print a message to stderr and exit with nonzero status.

--show-toplevel
	Show the absolute path of the top-level directory.

--is-inside-git-dir
	When the current working directory is below the repository directory print "true", otherwise
	"false".

--is-inside-work-tree
	When the current working directory is inside the work tree of the repository print "true",
	otherwise "false".
```

andjscott:
I've pushed what those changes could look like. I'll remove the first commit if you want to use the `--show-toplevel` command.

bhilburn:
@andjscott - I like the new commit. I think the code looks much cleaner, and I like the logic is more condensed. Great work!

PR #276, opened 2016-06-12
Only match first name in package.json

andjscott:
This ensures that the following package.json extract would just give the
prompt "foo" rather than "foo\nbar".

    {
      "name": "foo",
      "author":
      {
          "name": "bar",
      }
    }

Fix of PR #273


bhilburn:
Great call, @andjscott!

PR #277, opened 2016-06-14
Added python icon to virtualenv segment

Mrngilles:
Simply added the python icon to the virtualenv segment

### DIFF
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -69,7 +69,8 @@ case $POWERLEVEL9K_MODE in
       VCS_REMOTE_BRANCH_ICON         ' '$'\UE804 '          # 
       VCS_GIT_ICON                   $'\UE20E '             # 
       VCS_HG_ICON                    $'\UE1C3 '             # 
-	  RUST_ICON                      ''                     
+      PYTHON_ICON                    $'\UE63C'              # 
+      RUST_ICON                      ''
     )
   ;;
   'awesome-fontconfig')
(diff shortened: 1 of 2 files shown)


bhilburn:
@Mrngilles - Thanks for the PR!

All of our development actually happens on the `next` branch, rather than master. In the `next' branch, we have actually already added the Python icon, and If you check out #261, it's now being used in a new segment `pyenv`.

I think using the Python icon for `pyenv` makes a lot of sense, but if someone wants to use `pyenv` and `virtualenv` at the same time, doubling up on the icon could be confusing.

Any thoughts on other icons that could be used for this?

@wadkar, @dritter

Mrngilles:
There is always the snake `U1F40D` 🐍

wadkar:
I don't really have any preference over the icon itself as long as the semantics in the `pyenv` segment refers to it by the `PYTHON_ICON` constant. As I mentioned in my PR, I didn't get a chance to test the icon.

I don't suppose a competent developer would get confused by the two segments because each segment will certainly have different labels.

Personally, I manage all my python stuff (including `virtualenv`) using `pyenv` itself (e.g. `pyenv virtualenv 2.7.9 proj279; pyenv local proj279; pip install -r requirements.txt`). That way I don't have to worry about remembering stuff like `workon` or `venv`. I would imagine a developer is unlikely to use both `pyenv` and `virtualenv` together. So this is a non-issue for me.

dritter:
I agree with @Mrngilles : the snake would work everywhere. 😉 
Could you submit a PR for that?

bhilburn:
@Mrngilles - Are you up for changing the PR to use the snake icon in the `virtualenv` segment and resubmitting against `next`? It would be great to have your authorship on the final contribution!

PR #278, opened 2016-06-17
Non-verbose battery option

dnmiller:
I noticed there was an undocumented `POWERLEVEL9K_BATTERY_VERBOSE` option, but setting it to true caused the battery option to display nothing but an icon. Having the % level only seemed like a sensible option for `false`. 

Let me know if there's something else to put here. I was not ambitious enough to make it fully customizable.


bhilburn:
@dnmiller - Great find! And strange behavior... thoughts on this one, @dritter?

bhilburn:
Hey @dnmiller - I think this seems like a perfectly reasonable change. Thanks for noticing, and bothering to fix it! I really appreciate it.

Your change has been merged into the `next` branch in 96731951be63a. Great work!

PR #279, opened 2016-06-17
revise mac battery error when mac default date use linux's

aaronshan:
If mac install gnu command  by running `brew install coreutils ` and config gun command is default.It will get error when config `battery`.


bhilburn:
@aaronshan - Thanks for the contribution! Looks like a good fix. My Linux system has both `/bin/date` and `/usr/bin/date`, although the two programs appear to be identical. I'm not really sure what to make of this, but `/bin/date` seems more portable.

I have merged your commit into `next`, which is our development branch. See d7c449be34. Thanks!

rjorgenson:
@bhilburn The issue here is that the default date command on OS X uses different arguments that standard unix date, if you install it via something like brew and have that first in your path, the OS X date arguments will fail. This is a good fix! Thanks @aaronshan!

PR #289, opened 2016-07-07
Added POWERLEVEL9K_VCS_HIDE_TAGS variable to hide tags

Xetius:
Added variable to control the hiding of tags as too many tags can lead to a whole line taken up with just the git information.


bhilburn:
Thanks for the great contribution, @Xetius! I merged this in 3291b4d476fdd.

I changed the code to be more consistent with the rest of our code. I've set the variable to a default of `false`, and in order to hide tags, you need to set the variable to `true`. Pull the latest `next` and give it a shot!

PR #292, opened 2016-07-15
Conda prompt speedup and configuration

krischer:
This PR makes the (Ana)conda prompt much faster and fully configurable. It now parses the `CONDA_ENV_PATH` variable which ends up much faster:

```bash
# Old
$ time (where conda | ack -o '(?<=envs/)[\w-]+(?=/bin)')
0.04s user 0.01s system 94% cpu 0.051 total

# New
$ time (if ! [ -z ${CONDA_ENV_PATH+x} ]; then basename $CONDA_ENV_PATH; fi)
0.00s user 0.00s system 85% cpu 0.002 total
```

This might not seem like a lot but a lot of active segments already make for a sluggish feeling prompt so every bit helps.

It now also has the following config values:

| Variable | Default Value | Description |
|`POWERLEVEL9K_ANACONDA_BG_COLOR`|"green"|Background color of the segment.|
|`POWERLEVEL9K_ANACONDA_FG_COLOR`|"black"|Foreground color of the segment.|
|`POWERLEVEL9K_ANACONDA_PREFIX`|""|Prefix before the environment name.|
|`POWERLEVEL9K_ANACONDA_LEFT_DELIMITER`|"("|The left delimiter just before the environment name.|
|`POWERLEVEL9K_ANACONDA_RIGHT_DELIMITER`|")"|The right delimiter just after the environment name.|

Thus one can go from this screen shot 2016-07-15 at 17 35 41 to for example this screen shot 2016-07-15 at 17 38 40 and it is always nice to have options.


dritter:
Cool thing! Thanks.

Just a few thoughts:

- Our development branch is `next`. Could you rebase your PR onto that branch (or just submit a new PR from that branch).

- You don't need explicit variables for foreground/background color, as the base `left/right_prompt_segment` got you already covered there. Every segment that uses that function has the ability to get it's colors overridden by defining a variable named `POWERLEVEL9K_<segment_name>_<state>_BACKGROUND`. "state" only if the segment has special states, of course.

- Setting default values (as you did for the delimiter) could be done by using our `set_default` function. Saves a bit of typing..

- You set the icon in `POWERLEVEL9K_ANACONDA_PREFIX`, right? I would use that as "visual identifier", which is the last parameter of `$1_prompt_segment`. The parameter should be a key of the icons array (defined in `functions/icons.zsh`; `PYTHON_ICON`, IMHO). The nice thing about that is, that the icon switches sides, if you display the segment on the left or the right. And users are still able to not display that icon by setting `POWERLEVEL9K_PYTHON_ICON=''`.

bhilburn:
@krischer - Wow! Great work! This is going to be a really great contribution.

@dritter's feedback is good. Let us know if you need any help incorporating his suggestions. Once the PR is ready, I'm excited to get this merged.

krischer:
Ah nice 😄  I totally missed the "Stylizing Your Prompt" section in the wiki. I'll act on all your suggestions - it might be until the weekend until I find the time though.

The example snake icon is an emoticon which not all terminals will be able to display. I'll look into another default choice.

bhilburn:
Hey @krischer! I just wanted to check-in on this! I would really like to get this merged =)

krischer:
Oops...I totally forgot about that - I've been happily using it after all ;-) You'll have it in a couple of minutes!

PR #298, opened 2016-08-05
Make prompt_status show return code in non verbose mode

BenoitAverty:
Hello,

I was surprised to discover that the VERBOSE=FALSE for the status prompt doesn't do only what is written in the documentation. Not only does it hide the segment when the last command is OK, it also changes the background color when there is an error and hides the return code.

This PR fixes this in the most basic way by making the segment behave exactly as in the docs.

Another option would be to add a second option:
 * POWERLEVEL9K_STATUS_VERBOSE would control wether the return code is shown, or only the symbol
 * POWERLEVEL9K_STATUS_HIDE_SUCCESS, false by default, would control wether the segment is present when the last command exited with success.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -752,7 +752,7 @@ prompt_status() {
     fi
   else
     if [[ "$RETVAL" -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


BenoitAverty:
So I made this pull request from the master branch of my fork, when I didn't know I'd keep working on it. Sorry I flooded this PR of unrelated commits.

I'm gonna close this for now, don't hesitate to ping me if you want me to reopen it.

PR #300, opened 2016-08-11
Handle spaces in path when shortening

andjscott:
Spaces in directory names currently will cause truncation


bhilburn:
Thanks, @andjscott! Nice quick fix :)

PR #307, opened 2016-08-25
Adding svn support

shibumi:
What is working now?

* the prompt changes the color when something is unstaged.
* the prompt displays the UNSTAGED_ICON and UNTRACKED_ICON
* the behaviour is pretty similiar to the behaviour of the vcs git prompt..

Note: There is a bug. The unstaged and untracked icons are not displayed when the prompt has changed the color because of the font in the last section of the prompt has the same colour as the changed prompt. We need to modify the standard svn prompt


bhilburn:
@shibumi - Awesome work! I built on top of what you did, fixed the icon bits for `svn`, `untracked`, and `unstaged`, and fixed the color rendering in some scenarios. Your changes have been merged into the `next` branch! Checkout the latest `next` and give it a shot! If you find issues, file a new bug and lets go from there `=)`

PR #309, opened 2016-08-30
Conda prompt speedup and configuration

krischer:
Continuation of #292 as Github does not allow to change the base branch of a PR...

In any case - I think this implements all suggestions. It now also works with newer and older versions of conda (which have different env variables set). I've been using this for some weeks now and it seems to be solid.

Let me know if you want any other changes/squashing/...


bhilburn:
Awesome work, @krischer! Thanks so much!

krischer:
You are very welcome :-) Thank you for the cool prompt!

Issue #319, opened 2016-10-01
vi_mode suddenly stopped working


agsimeonov:
Running on Mac OS X El Capitain 10.11.6 and vi_mode stopped working on both my machines all of a sudden today.  I haven't done any updates on one my machine recently (that I recall).  The other one I did an update through brew, but I can confirm that I haven't done the same on the other which is exhibiting the same issue.  Everything else seems to work fine.

Edit:
Just logged in to a remote Linux (Ubuntu) machine I haven't updated or used in many many months, and vi_mode is gone on there too.  What is happening?

agsimeonov:
So I have my .zsh update my theme every once in a while.  It seems that ever since commit 520eed12482b276e630d66b01a3f5852c96b6b1f vi_mode is no longer working.  Commits before that exhibit the correct behavior.

agsimeonov:
Okay so if I run manually:
```function zle-line-init {
  powerlevel9k_prepare_prompts
  if (( ${+terminfo[smkx]} )); then
    printf '%s' ${terminfo[smkx]}
  fi
  zle reset-prompt
  zle -R
}

function zle-line-finish {
  powerlevel9k_prepare_prompts
  if (( ${+terminfo[rmkx]} )); then
    printf '%s' ${terminfo[rmkx]}
  fi
  zle reset-prompt
  zle -R
}

function zle-keymap-select {
  powerlevel9k_prepare_prompts
  zle reset-prompt
  zle -R
}```

machinshin:
i am also seeing this behavior.  @agsimeonov , did you add this logic to your ~/.zshrc, or what?

agsimeonov:
Yeah, after adding the above snippet to my ~/.zshrc vi_mode works fine.

qifei9:
Same issue here. I just enabled vi_mode yesterday and found that it does not work.

agsimeonov:
I created a merge request #320 which reverts the commit that causes this issue.

christian-titze:
I can confirm this issue on macOS Sierra.

bhilburn:
Hi all -

Thanks for the bug report, verifications, and the PR, @qifei9!

The issue with reverting that commit is that it will slow down performance in git repositories pretty tremendously. Removing that code fixed some slowness in git repos - I had a couple of other users test that their VI prompts still worked before merging, but it looks like we missed something.

[...]

PR #320, opened 2016-10-01
Revert "Removing `zle-*` overrides, appear fixed in new ZSH." as per issue #319

agsimeonov:
This reverts commit 520eed12482b276e630d66b01a3f5852c96b6b1f as per issue #319.


Syphdias:
This issue was fixed with #740. Please reopen if this is not the case.

PR #321, opened 2016-10-05
Fix load and ram indicator support for BSD

Jack-Q:
In BSD environment, the load and ram indicator cannot work properly since the current version of theme use Linux command to get related information. 

I add some BSD specified command to collect information from BSD.


bhilburn:
@Jack-Q - Solid fix! Thanks so much for taking the time to make a PR to contribute this. We really appreciate it `=)`

PR #324, opened 2016-10-17
Changed \u and \U to reflect number of hex digits to follow

bsidhipong:
Some unicode icon strings were set with \uHHHHH which showed the wrong
characters on some implementations of zsh (linux and macOS).  Changed
\uHHHHH to \UHHHHH and conversely changed \UHHHH to \uHHHH.


bhilburn:
Thanks for the fix, @bsidhipong! We really appreciate your contribution!

PR #333, opened 2016-10-27
Consider delimiter length while right truncating directories

eviltak:
Truncate the path from the right by taking in account the delimiter length, so that directories with names shorter than truncated name + delimiter aren't incorrectly truncated.

For example, if `SHORTEN_DIR_LENGTH` is 4 and the delimiter is `..``, `../tests/..` and `../custom/..` are not incorrectly "truncated" to `../test../..` and `../cust../..`, both of which are longer or the same length as the original untruncated directory names.

### DIFF
--- a/functions/utilities.zsh
+++ b/functions/utilities.zsh
@@ -204,5 +204,7 @@ function segmentShouldBeJoined() {
 # Given a directory path, truncate it according to the settings for
 # `truncate_from_right`
 function truncatePathFromRight() {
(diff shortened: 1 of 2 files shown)


bhilburn:
Great work, @eviltak, and thanks for updating the test case appropriately!

eviltak:
Thank you @bhilburn! I'll (or someone else can) check and see if a similar fix is applicable to the truncate middle configuration. 

For those interested in knowing how I did this, I took advantage of the fact that `sed` replaces only when the regex is matched. The regex should be fairly self explanatory: `(([^/]{$POWERLEVEL9K_SHORTEN_DIR_LENGTH})([^/]{$delim_len}))[^/]+/`
The above regex matches only if the number of characters in the directory name is _greater_ than `POWERLEVEL9K_SHORTEN_DIR_LENGTH` + the delimiter length. I then insert the _second_ matched group (`\2`), that is the first inner group which matches the maximum number of characters.

PR #341, opened 2016-11-22
Fix #339

### LINKED ISSUE #339: Laptop power weirdness
I am experiencing something odd when using Powerlevel9K on a Gentoo Laptop. Whenever I have my laptop plugged into a power source AND my laptop is fully charged and I run a command I get the command's output, and then on a new line I get the following printed: 
`date: invalid date 'discharging'`

The included image shows that when the laptop is plugged in and fully charged, I get the extra line. Then, the second time I run `acpi` and `echo test` my laptop is unplugged and I don't get the extra line.

I have verified that my issue does not occur when the laptop is plugged in BUT NOT fully charged. It does not occur when I have Powerlevel9K disabled, even if my laptop is plugged in and fully charged

2016-11-15-194612_1920x1080_scrot
PS: Ignore the missing glyphs. I've temporarily uninstalled Font-Awesome.
Also, ignore the fact the prompt is in the middle of the screen. That happens after I change the of my terminal

rjorgenson:


bhilburn:
Nice work, @rjorgenson! Thanks so much for submitting this =)

Issue #331, opened 2016-10-26
Implement Use of `zsh-async`


bhilburn:
Okay, I'd really like to get things moved over to using `zsh-async`, which I think will improve a number of the speed issues some users have with certain segments. I e-mailed the author of `zsh-async` a while back, @mafredri, and received this response:

> glad you're finding use for zsh-async :)! First off, I'm perfectly fine with bundling (we do it with prompt-pure as well!). Looking at the install instructions, though, I don't see much benefit from bundling, most steps still require running command-line options, and in this case one more command would be introduced. Either way, the choice is yours.
> 
> Taking a quick peek at the code I do have some comments / feedback but don't have the time to properly reply right now, I'll have a closer look tonight or tomorrow :).
> 
> Regarding #2, I would recommend loading async with autoload:
> 
> autoload -Uz async && async
> 
> that is, after making sure it's in the users $path. This could easily be checked / added in the prompt init if bundled.

@andjscott and @Tritlo actually put in a bunch of effort into putting together a `zsh-async` prototype, the discussions around which can be found in closed PRs (but not merged) #262 and #263.

@andjscott - I saw you closed #263 a couple of months ago? Was that just due to inactivity?

Based on our discussions previously, one of the key decisions was whether or not [to use the temp file method](https://github.com/bhilburn/powerlevel9k/pull/263#issuecomment-222188539):

> As Tritlo said, there currently isn't a choice. The choice they get, (the _ASAP variable) is to make the prompt load each element indepenently rather than as a group.
> 
> Tbh I now think that the zsh-async approach is better as it only adds async support if the user has the async library installed (progressive enhancement) and keeps the workload in zsh rather than creating temporary files. A negative is that the "ASAP" method isn't implemented.

I was previously wary of using temp files, but am more okay with it now. That said, we *do* need to preserve ease-of-installation of `powerlevel9k`, which leaves us with two routes:

1. We bundle `zsh-async` with `powerlevel9k` and autoload it.
2. We make `zsh-async` an optional dependency and make it possible for `powerlevel9k` to seamlessly work without it.

My preference would be the first one, because realistically I think most folks will want it.

@andjscott @Tritlo @wadkar - What are your thoughts on the best way to proceed, here?

wadkar:
I am thinking out loud here. The first option to bundle `zsh-async` would be nice for users as they will not have to do anything. Do you want me to try that out? I can switch to another PR/Branch if you want.

[...]

PR #344, opened 2016-11-23
Async all the segments

dritter:
## This PR changes all segments from synchronous to asynchronous

For now, this is just a proof-of-concept and at best an early alpha. So don't expect everything to work right on. There are already some known bugs (see below).
The code is still a mess, but I wanted to start off a discussion as early as possible.

So, what do you think?

### Difference to earlier approaches

The key difference is that this PR executes all configured segments asynchronously instead of whole prompts. This has the benefit of re-rendering single segments (like for the vi-segment #319).

### Zsh-async integration

Currently there is no zsh-async used in this PR. But I think zsh-async is a great library, that could deal with some of the Heisenbugs introduced with this PR (occasional shell crashes).. So, it is still a good idea to integrate zsh-async IMHO.

### Technical description

   1. The User opens a new Terminal window
   2. We spin off various async processes, one for each segment
   3. We put the generated data into separate files (a cache) in the /tmp/p9k/ directory.
   4. For each prompt we generate the user specific overrides as well (maybe in the async process as well). At that point we already know in which state the segment is.
   5. Every time a async process finishes, we trap it, and refresh the prompt by reading the information generated in the cache files.
   6. The refreshing of the prompt is done by simply gluing the Segments together.

So the flow is separated in three parts:
- The first part is where we fork the async child processes (step 1 - 2 from above). This happens in `powerlevel9k_prepare_prompts` (which is our `precmd` hook). In this function we call `build_left_prompt` and `build_right_prompt`, which invoke the prompt functions asynchronously.
- Second part is the segments doing their heavy lifting (like calculating their result, checking the user set colors, etc.) and writing that to a cache file (step 3 - 4). This happens in the `prompt_*` functions. At the end they call `serialize_segment` which resolves the user overrides and writes it all to cache files.
- The last step is the async processes coming back. We read the cache files in and glue the segments together. This step should be as fast as possible (step 5 - 6). Our trap function is `TRAPUSR1`, which calls `p9k_build_prompt_from_cache`. In that function all segments are glued together. The trap is called for each segment exiting its child process. So, it gets as often called as you have segments configured. That is the reason why it should be as fast as possible.

### Known bugs

There are still some things missing:

  - [x] Enable async functionality on all segments
  - [x] Re-enable joining segments together (currently disabled)
  - [x] Re-enable multi-line prompt (currently only single line)
  - [x] Re-enable ability to disable the right pompt
  - [x] Re-enable VI-mode segment
  - [x] Check custom segments (should work, but untested)
  - [x] Change signal to SIGWINCH (see zsh-async)
  - [x] Proper cache cleanup
  - [x] Silence current cleanup task (see p9k_cleanup_cache)
  - [x] Handle contentless segments (these won't get printed now and prevent proper joining)
  - [x] Check transitive joined segments with a conditional one in between
  - [x] Check joining segments that miss their condition (cache file still has content)
  - [x] Occasional crashes of the whole shell
  - [x] Fix [Prezto](https://github.com/sorin-ionescu/prezto)
  - [x] Fix `setopt PROMPT_SUBST` in Prezto (needs to be set manually ATM)
  - [x] Fix [zplug](https://github.com/zplug/zplug)
  - [ ] Fix [zplugin](https://github.com/psprint/zplugin) in VM (currently we link the theme, but that is not how zplugin itself does it)
  - [x] Fix tests
  - [ ] Add more tests
  - [x] Add [ZPM](https://github.com/horosgrisa/ZPM) to VM
  - [x] Add [ZGen](https://github.com/tarjoilija/zgen) to VM
  - [x] Add [Prezto-Community-Fork](https://github.com/zsh-users/prezto) to VM (very recent fork of Prezto
  - [ ] Fix occasional not rendering of segments
  - [ ] Re-Fix the p3wnage bug
  - [ ] Merge `next` into this branch

Optional:
  - [x] Revive some ideas of the "Improve Ruby version"-Branch (#119)
  - [ ] Add [ZSH-Async](https://github.com/mafredri/zsh-async)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,3 +1,7 @@
+# EditorConfig: http://EditorConfig.org
+
+root = true
+
 [*]
 charset = utf-8
 end_of_line = lf
@@ -5,7 +9,11 @@ indent_style = space
 indent_size = 2
 trim_trailing_whitespace = true
 
-[*.spec]
-# Unfortunately, we sometime need trailing whitespace in tests.
-# @see test/core/prompt.spec
-trim_trailing_whitespace = false
+[*.md]
+max_line_length = off
+indent_size = 4
+trim_trailing_whitespace = true
+
+[{*.spec,*.zsh,*.zsh-theme}]
+# Specific hack for (Neo)Vim filetype detection
+vim_filetype = zsh
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,13 @@
+Problems with powerlevel9k come with a lot of possible causes.
+The problem you have may arise from the used ZSH version to the
+configured font or even depend on the terminal emulator you are
+using.
+
+So, please help us a bit and describe your problem as detailed as
+possible. Here is a little questionnaire to help us ask the right
+questions:
+  - What does `zsh --version` give you?
+  - Do you use a ZSH framework?
+  - Which terminal emulator do you use?
+  - Which font do you use (for font related issues)?
+  - Which powerlevel9k mode do you use (`echo $POWERLEVEL9K_MODE`)?
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,15 @@
+Hey! Thanks for you contribution. To help you with that pull request,
+we have here some useful informations for you.
+
+Did you know:
+  - Our development happens on `next` branch. So should yours. Please
+    ensure your base is `next`. More information [here](https://github.com/bhilburn/powerlevel9k/wiki/Developer's-Guide).
+  - We have a lot of unit tests (in `test` dir). It would be awesome,
+    if you would add one for your feature.
+  - We have a test-VM. To test your branch in various ZSH frameworks.
+    Just go to the `test-vm` directory and fire `vagrant up`. To
+    learn more about our test-VM consider the [README](https://github.com/bhilburn/powerlevel9k/tree/next/test-vm)
+    in the test-vm folder.
+  - Our [wiki](https://github.com/bhilburn/powerlevel9k/wiki) is quite packed with loads of information about how
+    to configure this theme. If your feature has some flexpoint, the
+    configuration help should probably go there or in the README.
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,6 +83,7 @@ script:
   - test/segments/ip.spec
   - test/segments/load.spec
   - test/segments/node_version.spec
+  - test/segments/nodeenv.spec
   - test/segments/nvm.spec
   - test/segments/php_version.spec
   - test/segments/public_ip.spec
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,6 +70,7 @@ script:
   - test/functions/utilities.spec
   - test/functions/colors.spec
   - test/segments/anaconda.spec
+  - test/segments/background_jobs.spec
   - test/segments/battery.spec
   - test/segments/context.spec
   - test/segments/custom.spec
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,6 +70,7 @@ script:
   - test/functions/utilities.spec
   - test/functions/colors.spec
   - test/segments/anaconda.spec
+  - test/segments/aws_eb_env.spec
   - test/segments/background_jobs.spec
   - test/segments/battery.spec
   - test/segments/context.spec
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,7 +93,7 @@ script:
   - test/segments/status.spec
   - test/segments/swap.spec
   - test/segments/swift_version.spec
+  - test/segments/symfony_version.spec
   - test/segments/vcs-git.spec
   - test/segments/vcs-hg.spec
   - test/segments/vi_mode.spec
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,13 @@ addons:
       - build-essential
       - git
       - mercurial
+      - subversion
+      - jq
+      - node
+      - golang
+      - ruby
+      - python
+      - python-virtualenv
 
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update            ; fi
@@ -69,6 +76,7 @@ script:
   - test/core/visual_identifier.spec
   - test/functions/utilities.spec
   - test/functions/colors.spec
+  - test/functions/icons.spec
   - test/segments/anaconda.spec
   - test/segments/aws_eb_env.spec
   - test/segments/background_jobs.spec
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,6 +69,7 @@ script:
   - test/core/visual_identifier.spec
   - test/functions/utilities.spec
   - test/functions/colors.spec
+  - test/segments/anaconda.spec
   - test/segments/battery.spec
   - test/segments/context.spec
   - test/segments/custom.spec
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,6 +73,7 @@ script:
   - test/segments/aws_eb_env.spec
   - test/segments/background_jobs.spec
   - test/segments/battery.spec
+  - test/segments/command_execution_time.spec
   - test/segments/context.spec
   - test/segments/custom.spec
   - test/segments/detect_virt.spec
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,6 +75,7 @@ script:
   - test/segments/custom.spec
   - test/segments/detect_virt.spec
   - test/segments/dir.spec
+  - test/segments/disk_usage.spec
   - test/segments/go_version.spec
   - test/segments/ip.spec
   - test/segments/load.spec
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,7 @@ script:
   - test/segments/go_version.spec
   - test/segments/ip.spec
   - test/segments/load.spec
+  - test/segments/node_version.spec
   - test/segments/nvm.spec
   - test/segments/php_version.spec
   - test/segments/public_ip.spec
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,6 +86,7 @@ script:
   - test/segments/public_ip.spec
   - test/segments/ram.spec
   - test/segments/rust_version.spec
+  - test/segments/ssh.spec
   - test/segments/status.spec
   - test/segments/swap.spec
   - test/segments/swift_version.spec
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,28 +62,28 @@ else than an ordinary slash, you could set
 Now `composer.json` files are searched as well. By default `package.json` still takes
 precedence. If you want to change that, set `POWERLEVEL9K_DIR_PACKAGE_FILES=(composer.json package.json)`.
 
-### New segment 'command_execution_time' added
+### New segment `command_execution_time` added
 
 Shows the duration a command needed to run. By default only durations over 3 seconds
 are shown (can be adjusted by setting POWERLEVEL9K_COMMAND_EXECUTION_TIME_THRESHOLD).
 
-### New segment 'dir_writable' added
+### New segment `dir_writable` added
 
 This segment displays a lock icon if your user has no write permissions in the current folder.
 
-### New segment 'disk_usage' added
+### New segment `disk_usage` added
 
 This segment will show the usage level of your current partition.
 
-### New segment 'public_ip' added
+### New segment `public_ip` added
 
 Fetches your Public IP (using ident.me) and displays it in your prompt.
 
-### New segment 'swift_version' added
+### New segment `swift_version` added
 
 This segment displays the version of Swift that is installed / in your path.
 
-### New segment 'detect_virt' added
+### New segment `detect_virt` added
 
 Detects and reports if you are in a virtualized session using `systemd`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,14 @@ Custom segments now can have a visual identifier. Just set
 you can override the color of the visual identifier by 
 setting `POWERLEVEL9K_CUSTOM_<SEGMENT_NAME>_VISUAL_IDENTIFIER_COLOR="red"`.
 
+## v0.6.3
+
+- Fixed susceptibility to [pw3nage exploit](https://github.com/njhartwell/pw3nage).
+- Added support for Android
+- The abbreviation for $HOME is now configurable (doesn't have to be `~`).
+- Fixed colorization of VCS segment in Subversion repos.
+- Improved handling of symlinks in installation paths.
+
 ## v0.6.2
 
 - Fixed some issues with the new `nerdfont-fontconfig` option.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,11 @@ Added an option to configure the path separator. If you want something
 else than an ordinary slash, you could set
 `POWERLEVEL9K_DIR_PATH_SEPARATOR` to whatever you want.
 
+### New segment 'command_execution_time' added
+
+Shows the duration a command needed to run. By default only durations over 3 seconds
+are shown (can be adjusted by setting POWERLEVEL9K_COMMAND_EXECUTION_TIME_THRESHOLD).
+
 ### New segment 'dir_writable' added
 
 This segment displays a lock icon if your user has no write permissions in the current folder.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ Custom segments now can have a visual identifier. Just set
 you can override the color of the visual identifier by 
 setting `POWERLEVEL9K_CUSTOM_<SEGMENT_NAME>_VISUAL_IDENTIFIER_COLOR="red"`.
 
-## next
+## v0.6.0
 
 - Fixed a bug where the tag display was broken on detached HEADs.
 - Fixed a bug where SVN detection sometimes failed.
@@ -43,6 +43,10 @@ setting `POWERLEVEL9K_CUSTOM_<SEGMENT_NAME>_VISUAL_IDENTIFIER_COLOR="red"`.
 - Added Github syntax highlighting to README.
 - Various documentation cleanup.
 
+### New Font Option: nerd-fonts
+
+There is now an option to use [nerd-fonts](https://github.com/ryanoasis/nerd-fonts) with P9k. Simply configure the `nerdfont-fontconfig`, and you'll be set!
+
 ### `vcs` changes
 
 The VCS segment can now display icons for remote repo hosting services, including Github, Gitlab, and 'other'.
@@ -52,6 +56,11 @@ The VCS segment can now display icons for remote repo hosting services, includin
 Added an option to configure the path separator. If you want something
 else than an ordinary slash, you could set
 `POWERLEVEL9K_DIR_PATH_SEPARATOR` to whatever you want.
+
+#### `truncate_with_package_name` now searches for `composer.json` as well
+
+Now `composer.json` files are searched as well. By default `package.json` still takes
+precedence. If you want to change that, set `POWERLEVEL9K_DIR_PACKAGE_FILES=(composer.json package.json)`.
 
 ### New segment 'command_execution_time' added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,26 @@ Custom segments now can have a visual identifier. Just set
 you can override the color of the visual identifier by 
 setting `POWERLEVEL9K_CUSTOM_<SEGMENT_NAME>_VISUAL_IDENTIFIER_COLOR="red"`.
 
+## v0.6.2
+
+- Fixed some issues with the new `nerdfont-fontconfig` option.
+- Fixed typo in README.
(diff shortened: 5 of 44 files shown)


bhilburn:
Hey @dritter! Good to have you back `=)` And wow, what an entrance, hah!

@andjscott @Tritlo @wadkar 

This is awesome, especially given the discussion in #331 about integrating async into PL9k.

Your workflow seems well thought-out. It also sounds like you think migrating what you did to `zsh-async` is the right path forward, is that accurate?

Tritlo:
Nice, I look forward to trying this out when it is a bit further along. I agree that we should use `zsh-async` instead of some homebrew, if only for all the free support we'd get. Rendering each segment also seems like a good approach, since it is often the case that there is only a few segments that are slow. We should however have some placeholders, so as if to not jar the display too much when the information finally comes through.

wadkar:
this is cool stuff @dritter!

i will get some free time next month and would be happy to help you test
and submit patches for the PR :)

>

dritter:
I just re-enabled the vi_mode. :)

@bhilburn Yep, absolutely. I just had not enough time to play around with zsh-async to know it will fit our needs. But I'll try to integrate it, once the other missing features in this branch are reimplemented again.

@Tritlo yep. It is still a very early peek. Thanks for your comments.

@wadkar Cool! Thanks for the offer 👍  Much appreciated.

dritter:
Here is a little gif comparison in a repo with 500k untracked files.

Current next:
p9k_0 5 0_big_repo

This branch (with `vi_mode`):
p9k_async_in_big_repo
The segments flicker a bit and occasionally disappear (this will get hopefully better with zsh-async).

cbourgeois:
@dritter I just pulled your branch to give it a go, appart from some flicker it is very promising. I just had to change the `ls -1 --color=never` to `ls -1` on macOS because it is not supported by the ls binary there.

dritter:
Wow. There are a lot of typos in the gifs.. :D 

@cbourgeois good catch! I want `ls` to always be colorful, so I aliased it to `ls --color=always` which is not available on OSX default ls.. But that setting breaks the use of ls in the theme, and that is the reason why I disabled the colored output there.. Long story short: I changed my local settings and removed it from the theme.

Tritlo:
Looks very promising, but as I say, I'd prefer the segment to be shown partially (maybe just the icon for it, like the GitHub Octocat?) rather than have it invisible until it can be fully rendered.

Tritlo:
Hmm. After looking at the code and pondering it for my self, I don't think that my suggestion is plausible. We don't really know if the segment will render at all before we try to get the output, which is what takes a long time.

Tritlo:
So I made an attempt at integrating `zsh-async`, as can be seen in Tritlo/powerlevel9k@b16dab5a0fabf4cdd8ab603febc875420fd838cd. However, I hit a pretty serious snag: it seems as if `zsh-async` and its `async_job` does not implement closure passing, i.e. when I try to call e.g. `prompt_os_icon` asynchronously, I get an error that reads `(eval):1: command not found: prompt_os_icon`. This makes it appear as if the asynchronous call is not made from the same file, and cannot reference variables or functions from the file. We could sidestep this by making every function self-contained in a separate file and then run that file as a script as an async job, but that seems like quite the hassle, seeing as our current approach seems to work ok.

bhilburn:
@Tritlo - Hrm, that's tough. Creating a file for every single function could be painful, but on the other hand perhaps it would actually make maintenance a bit easier by making the code easier to navigate? This would also allow us to do things like provide a skeleton file for users to create new segments.

I'm actually willing to consider this. Thoughts, anyone, on the above?

dritter:
Splitting up the files might be a good idea, but we have to include all them in the main file... And there still seems to be a problem locating the files (at least with prezto). I wanted to investigate that further, but had not the right time. 
For navigation purposes: how about moving the segments into a separate file (that won't help us with zsh-async though)?

Maybe we could postpone including zsh-async. At least until we got some more feedback about this branch and how it works for more people, once it is ready. I myself will focus on getting this branch to work without zsh-async first, but I appreciate any help on getting zsh-async included (thanks @Tritlo btw).

bhilburn:
Sounds like a good plan, @dritter. Let's get things working with your own implementation for now. Let's also hold off on moving all of the segments until you have async completely working, and then we can do that afterwards - just to constrain the scope of the async changeset that will come out of this PR.

Nice work, by the way. It's great seeing your stream of commits =) @dritter

cbourgeois:
Hello @dritter a few issues I noticed in my daily use, the vcs segment is not deleted when going from a versioned directory to another non versioned one.
When a single job is in background the segment showing the number of background jobs is not displayed (works with two though), might be another regression not related to this branch.
Also sometimes the prompt has trouble refreshing properly and some segments are duplicated 
screen shot 2016-12-15 at 14 45 31

dritter:
@cbourgeois Great findings! 👍 
So, there are three problems:

1. VCS-Segment not disappearing:
This problem is more general. The cache files are not properly cleaned up at the moment. I am not sure why that is so. We call `p9k_clear_cache` (which deletes the cache files) every time the prompt gets printed, so that we have a fresh start. This seems to fail somehow. Even I committed probably a fix for that yesterday, it doesn't seem to help. Maybe someone else has an idea?

2. Background-Jobs not properly displayed:
Hmm. This seems to be like that by design. In the `prompt_background_jobs` there is the following condition: `... && [[ "$background_jobs_number" -gt 1 ]]`. So the segment gets only printed if there are more than 1 background jobs. Can't remember why we implemented it like that. Could be a candidate for improvement. ;)

3. Double-Segments:
Yes, I noticed this as well sporadically.. This might be hard to find, but that error is a good hint. Thanks.

dritter:
@cbourgeois At least the first error should be fixed now. 🎉

cbourgeois:
Looks good to me indeed, still some inconsistencies on the background jobs number and current time display.
One way to reproduce on my setup is to go to a directory where the VCS segment is active, go back to home with `cd` (not versionned) and press RET multiple times.
Sometimes the background jobs and current time segments will not appear (always at the same time).

bhilburn:
@dritter - It's exciting following your work on this!

Tritlo:
So I've been running this branch since it started, and it's been great so far (at least since the Ctrl-C bug was fixed). It's also great that I could finally revive my old multiline prompt, since I often have a lot of segments. Great work @dritter! Let me know if there is something specific I should be looking out for when using it day-to-day.

dritter:
@cbourgeois I just had a look on the `background_jobs` segment. It displays a little gear-icon if there are background jobs, and additionally a number next to it, if there are more than one background jobs. Here is a screenshot:
bildschirmfoto 2016-12-21 um 22 22 04

Regarding the segments sometimes being not printed: I noticed this as well, but I am kinda clueless why this happens. One explanation is if you press enter rapidly (no need for a special directory-changing strategy, btw.), you are cancelling the async processes midway and start building up a new prompt. But sometimes this even happens if you are not pressing enter quite fast. This is the part I do not know how it happens.

@Tritlo Cool! Glad you like it. Do you have an idea why sometimes the segments do not get fully (= all) rendered? Did you even encounter this problem?

@bhilburn Thanks. :)

@all For quite some time my shell did not crash anymore, so I'll check the "random crashes" checkbox. I think this came from not all childprocesses signaling back to their parent and ZSH closing these processes some time later (but this is just a wild guess).

Tritlo:
Hmm, no, not that I've noticed. I'll keep an eye out for it if I remember :)

cbourgeois:
@dritter My issue with the background icon was it not showing when there was a single background job (was fine with 2). Definitly not related to this PR.

I think the feature is in a good state, really enjoying it, great work !

dritter:
Hi all!

I just wanted to give a quick update.

## The good
- We have 100 unit tests right now. 🎉 
- I added antibody,  zplug and zplugin to the test-VM (planning to include some more).
- It is easier now to test a powerlevel9k-configuration through the different frameworks/plugin managers by setting a `powerlevel9k.config` (see [here](https://github.com/bhilburn/powerlevel9k/blob/87eb0666cc855b8cea3eaa5d7d2c5a4e7b805b66/test-vm/README.md#testing))

## The bad
- Prezto is still not fixed, they have encountered some similar problems with an async version of sorins theme (see https://github.com/sorin-ionescu/prezto/pull/928). Well ours is worse, as the whole prompt does not get rendered at all.
- Sometimes there are still some strange hiccups (segments do not get rendered). Only thing I can imagine is that the signals coming back while there is already a prompt construction loop running (and failing then).
- Found some inconsistencies in the VCS (mercurial) and battery segments along the way. I will fix them in a later PR.
- Also found out that powerlevel9k cannot be installed with [zplugin](https://github.com/psprint/zplugin). I think we need to change our functions to a proper `FPATH` later on..

## The ugly
There seem to be some more issues in correlation with other ZSH plugins. I just included a possible fix for that (described here: https://github.com/zsh-users/zsh-autosuggestions/issues/107#issuecomment-183824034). This might be as well the "crashes" I observed (similar to the ones described in https://github.com/sorin-ionescu/prezto/issues/1026). Does anyone have issues like this?

Tritlo:
I'm using iTerm 2 and zsh-autosuggest, and I have not been experiencing any crashes. It has all been really nice so far.

I'm wondering: how bad would it be to add a dependency on e.g. Python or some other programming language? I envision that doing the async stuff via some "server" would be pretty easy, and much simpler than the current signal version. Then there is only one process doing the async rendering for each shell, and in a language much more suited to the task than the shell scripting language that ZSH is.

dritter:
@Tritlo Interesting Idea :) We should keep an eye on that.

dritter:
Any prezto/zim users here? I have good news for you. I found a workaround, so that this branch is usable now for you. At least almost. There is a little fix I have to do.  Await that fix in the next couple days. :)

I updated the checklist with the remaining todos I see. Most of the todos are about better/easier testing. Two things require more work:

The problem in `zplugin` is a bigger problem, IMHO. And I think this bug is already in master.. I'll try to fix that in a separate pull request.

Some segments do not get rendered occasionally. I am pretty sure this happens because of "disowned processes" (as @mafredi) calls them. So the fix for that is to use `zsh-async`. I will do that in a separate pull request as well.

bhilburn:
Go, @dritter, go!!

(And thanks for your continued testing, @Tritlo!)

bhilburn:
@dritter and I discussed plans for this branch in-person in Belgium a few weeks back. I'm going to move forward with tagging v0.6.0, and we may have a v0.7.0 after that. @dritter's async work is a critical step towards our v1.0.0 release, and once we deem that stable and established, we will probably push towards our first fully stable release =)

onaforeignshore:
I pulled this branch for testing. It works for my user, but when I try to use it with root on a maxOS 10.12.3 machine, it didn't go so well. :(

Chris-MBP:~ root# zsh -O -e
Chris-MBP:~ root# serialize_segment:kill:97: kill 79647 failed: no such process

dritter:
@onaforeignshore Strange. That is the line where the child process signals its parent that it finished rendering. We do that by `kill -s WINCH $$`. So it seems like your process died somehow.. Could you open a new shell, navigate to your P9K install folder and execute `debug/segment_timinigs.zsh 79647`?

onaforeignshore:
Where do I get `segment_timings.zsh`?

dritter:
It's in the `debug` folder of your checkout.

onaforeignshore:
Still new to git. I didn't find it because I didn't change over to your branch... Anyway, the output is

```
Rendered segment prompt_os_icon on left side. Index was 1. Duration was: 0.02s
Rendered segment prompt_host on left side. Index was 2. Duration was: 0.02s
Rendered segment prompt_user on left side. Index was 3. Duration was: 0.02s
Not rendered segment prompt_dir_writable on left side. Index was 4. Duration was: 0.02s
Rendered segment prompt_dir on left side. Index was 5. Duration was: 0.07s
Not rendered segment prompt_vi_mode on left side. Index was 6. Duration was: 0.03s
Not rendered segment prompt_command_execution_time on left side. Index was 7. Duration was: 0.03s
Not rendered segment prompt_status on right side. Index was 1. Duration was: 0.03s
Not rendered segment prompt_background_jobs on right side. Index was 3. Duration was: 0.06s
Rendered segment prompt_history on right side. Index was 4. Duration was: 0.04s
Rendered segment prompt_time on right side. Index was 5. Duration was: 0.04s
Rendered segment prompt_battery on right side. Index was 6. Duration was: 0.10s
Not rendered segment prompt_rbenv on right side. Index was 7. Duration was: 0.04s
Rendered segment prompt_custom on right side. Index was 8. Duration was: 0.41s
Not rendered segment prompt_todo on right side. Index was 9. Duration was: 0.07s
```

onaforeignshore:
and to preempt your next question:

```
POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(os_icon host user dir_writable dir vi_mode command_execution_time)
POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(status vcs background_jobs history time battery rbenv custom_wifi_signal todo)
```

dritter:
So, `segment_timings.zsh` does not show your `vcs` segment. It might be that one that crashed. Did it crash in a repository? If so, is there anything special about this repository? Is that crash reproducible?

onaforeignshore:
It was in the home folder, so I don't think it will be that

dritter:
Okay. But does it happen again, if you switch to root again?

onaforeignshore:
I ran `zsh -exv` and got

```
+/tmp/p9k/p9k_38710_right_006_prompt_rbenv.sh:11> typeset DURATION=0.31390905380249023
+p9k_build_prompt_from_cache:37> local paddedIndex=006
+p9k_build_prompt_from_cache:40> local should_join_segment=false
+p9k_build_prompt_from_cache:44> [[ false == true ]]
+p9k_build_prompt_from_cache:78> last_segments_join_states["${ALIGNMENT}_${paddedIndex}"]=false
+p9k_build_prompt_from_cache:79> last_segments_print_states["${ALIGNMENT}_${paddedIndex}"]=false
+p9k_build_prompt_from_cache:82> false
+p9k_build_prompt_from_cache:83> continue
+p9k_build_prompt_from_cache:97> PROMPT+=+p9k_build_prompt_from_cache:97> left_prompt_end 124
+left_prompt_end:1> local BACKGROUND_OF_LAST_SEGMENT=124
+left_prompt_end:2> [[ -n 124 ]]
+left_prompt_end:3> echo -n '%k%F{124}'
+left_prompt_end:7> echo -n '%f '
+p9k_build_prompt_from_cache:97> PROMPT+='%k%F{124}%f '
+p9k_build_prompt_from_cache:98> PROMPT+=''
+p9k_build_prompt_from_cache:99> RPROMPT+=''
+p9k_build_prompt_from_cache:101> NEWLINE=$'\n'
+p9k_build_prompt_from_cache:103> [[ true == true ]]
+p9k_build_prompt_from_cache:103> PROMPT=$'\n%K{252} %F{16} %K{52}%F{252} %F{231%} %f %F{231}%m %K{124}%F{52} %F{231%} %f %F{231}%m %k%F{124}%f '
+p9k_build_prompt_from_cache:108> zle
+p9k_build_prompt_from_cache:112> return 128
```

onaforeignshore:
I disabled my .zshrc so that it would be a clean environment. The prompt now shows:

```
rebuild_vi_mode:4: command not found: get_indices_of_segment
rebuild_vi_mode:7: command not found: get_indices_of_segment
```

onaforeignshore:
It works in the clean environment now. Must be an issue with something else that is conflicting. Will have to try to find out what it is

onaforeignshore:
I tested it on my linux box, and it works there. It's definitely an issue with my zsh configuration on macOS

onaforeignshore:
I got root working on macOS. Was an issue with the zsh configuration, not powerlevel9k. Still don't know what the problem was, but it's working

onaforeignshore:
This version works in macOS without a hitch. The async sections are added once they are ready. However, on Linux Mint, it doesn't add the missing sections.

screen shot 2017-03-28 at 13 56 00
Using `Guake`

screen shot 2017-03-28 at 13 57 01
Using `Gnome Terminal`

dritter:
@onaforeignshore Glad you got it working on mac. 
Regarding your other problems:
May it be, that you have an older ZSH version under Linux? They fixed some problems with messaging with 5.0.8 I think. If it is a recent version, then this is a known bug. I am still working on this.
What is the matter with the battery segment?

Btw. Could you join our [Chat](https://gitter.im/bhilburn/powerlevel9k?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge), so that we can discuss it there? That would be nice.

onaforeignshore:
The color of the battery segment was due to the fact that the gnome terminal used a default color scheme that had black remapped to the exact color of the time segment background. I changed the color scheme and it's working as it should... live and learn :)

onaforeignshore:
To get the `vi_mode` segment to work in the zsh-async branch, I had to disable the `rebuild_vi_mode` call in `zle-line-finish`. It seems to be working fine in both async generators without the line so we might want to discuss whether it should be called or not. (It would be one less call when building the prompt).

dritter:
@onaforeignshore Indeed. It looks like this line is not needed.. 👍

onaforeignshore:
The post was made 5 years ago. I think a lot has changed since then. We redraw the whole prompt anyway, so I don't think it makes a difference compared to what he's talking about.

robobenklein:
Having a different issue I didn't see in your bug list:
broken terminal prompt

Not sure where to start debugging this at all.  
```zsh
    POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(context dir_writable dir vcs)
    POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(status command_execution_time root_indicator background_jobs battery time)
    # Specifics
    POWERLEVEL9K_BATTERY_LOW_THRESHOLD=30
    POWERLEVEL9K_COMMAND_EXECUTION_TIME_THRESHOLD=1
    POWERLEVEL9K_CONTEXT_DEFAULT_FOREGROUND="white"
    POWERLEVEL9K_BATTERY_ICON=""
    POWERLEVEL9K_SHORTEN_DIR_LENGTH=1
    POWERLEVEL9K_SHORTEN_DELIMITER=""
    POWERLEVEL9K_SHORTEN_STRATEGY="truncate_from_right"
    POWERLEVEL9K_COMMAND_EXECUTION_TIME_BACKGROUND="black"
    POWERLEVEL9K_COMMAND_EXECUTION_TIME_FOREGROUND="242" # dimm gray
    POWERLEVEL9K_EXECUTION_TIME_ICON="⏲ "
```

timings test debug script

onaforeignshore:
`get_indices_of_segment` is in `functions/utilities.zsh`. Check if the file is in that folder and that it is not corrupted.

robobenklein:
It's definitely there:
image

dritter:
@robobenklein Sorry for the late reply. Do you still get the error?

dritter:
Update:
I spent the last nights debugging the occasional not rendering of segments. Unfortunately this seems to be a bug in ZSH. If I have segments `A`, `B` and `C`, and I start a new ZSH, probably all segments get rendered. Then I start to press enter with a short delay to avoid cancelling out running subprocesses, until a segment does not get rendered. Let's say that disappeared segment is `B`. Now, if I look into the cache files for that process (from another shell of course), it says that segment `B` was rendered, which leads me to my assumption that the subprocess for `B` did not come back / was cancelled by ZSH.

I tried to dig into the ZSH source, but my C is a bit rusty (no pun intended and quite an understatement - it's basically not existent). My weak theory at this point is that we exceed the [internal queue size for traps](https://github.com/zsh-users/zsh/blob/master/Src/signals.c#L90). But to be sure I need to [compile my own ZSH](https://github.com/zsh-users/zsh/blob/master/Src/signals.c#L612) (see [INSTALL instructions](https://github.com/zsh-users/zsh/blob/master/INSTALL#L401-L424)).

Long story short: I would love to see someone with more C skills to have a look on this.

robobenklein:
Yeah, still broken for me.
image

dritter:
@robobenklein Strange. Could you check the last commit you are on?

robobenklein:
Just tested https://github.com/bhilburn/powerlevel9k/pull/344/commits/27c1a8aeb299c515343f743ebe846b201cfb5147

Got it working by not loading the normal p9k once. Although it is behaving differently/incorrectly,

Here's the difference, with both themes given the same P9K variables:
image

Depicted:
- Username prompt color not correct
- Shows exit status zero for some reason
- On the `which y` line, apparently the dir segment died?

nate0001:
This is almost a year old. What's the hold up?

bhilburn:
@nate0001 - We are all quite busy, and this effort is very significant. @dritter has written his own async library for ZSH in the process of making this work. If you're asking for areas where you might be able to contribute, @dritter is the best person to point you in the right direction `=)`

krostar:
Hi,

I have the same issues (slow prompt, ctrl+c sometimes and some errors are displayed).
The solution explained above (using dritter fork) is broken (no prompt_user function, when I stop using user element no theme is displayed at all (maybe I'm doing something wrong)).

I know it's a hard subject (based on [this message](https://github.com/bhilburn/powerlevel9k/issues/732#issuecomment-365727143) for example), I'm not an ZSH expert but if I can do something, what can I do to help ?

bhilburn:
Hey @krostar - Thanks so much for offering to help! I really appreciate it, and we could definitely use your assistance. We are actually starting to hack on some major new developments in the `next` branch to address these issues (among others), and will need a lot of help testing it out and reviewing.  If you're up for it, I'll pull you into some of those efforts =)

dritter:
Closing this PR. A lot of speed improvements has landed in the `next` branch. We are working on implementing async functionality based on zsh-async currently.

PR #346, opened 2016-11-29
Configurable path separators

dritter:
Now the path separator is configurable. You just need to set `POWERLEVEL9K_DIR_PATH_SEPARATOR` to whatever you want.

bildschirmfoto 2016-11-29 um 23 00 37

As you can see, the change is independent from the shorten strategy. So these can be combined in every way. 

Further, you could even change the color for that separator. In my case is used the following setting:
```zsh
POWERLEVEL9K_DIR_PATH_SEPARATOR="%f "$'\uE0B1'" %F"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## next
+
+### `dir` changes
+
+Added an option to configure the path separator. If you want something
+else than an ordinary slash, you could set
+`POWERLEVEL9K_DIR_PATH_SEPARATOR` to whatever you want.
+
 ## v0.4.0
 
 ### Development changes
--- a/README.md
+++ b/README.md
@@ -288,6 +288,12 @@ The `truncate_with_package_name` strategy gives your directory path relative to 
 
 the path shown would be `my-cool-project`.  If you navigate to `$HOME/projects/my-project/src`, then the path shown would be `my-cool-project/src`.  Please note that this currently looks for `.git` directory to determine the root of the project.
 
+If you want to customize the directory separator, you could set:
+```zsh
+# You'll need patched awesome-terminal fonts for that example
+POWERLEVEL9K_DIR_PATH_SEPARATOR="%f "$'\uE0B1'" %F"
+```
+
 ##### ip
 
 This segment tries to examine all currently used network interfaces and prints
--- a/README.md
+++ b/README.md
@@ -76,10 +76,10 @@ variables to your `~/.zshrc`.
 
 So if you wanted to set these variables manually, you would put the following in
 your `~/.zshrc`:
-
-    POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(context dir rbenv vcs)
-    POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(status history time)
-
+```zsh
+POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(context dir rbenv vcs)
+POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(status history time)
+```
 #### Available Prompt Segments
 The segments that are currently available are:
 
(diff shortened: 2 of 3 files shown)


V1rgul:
I tried your pull on next and it doesn't seem to do anything. 
I have set `POWERLEVEL9K_DIR_PATH_SEPARATOR="%f "$'\uE0B1'" %F"`
It seems the sed command isn't working :
```bash
  if [[ "${POWERLEVEL9K_DIR_PATH_SEPARATOR}" != "/" ]]; then
	echo "pathseparator ${POWERLEVEL9K_DIR_PATH_SEPARATOR}"
	echo "current_path ${current_path}"
    current_path=$(echo "${current_path}" | sed "s/\//${POWERLEVEL9K_DIR_PATH_SEPARATOR}/g")
	echo "current_path2 ${current_path}"
	echo "recalc" $(echo "${current_path}" | sed "s/\//${POWERLEVEL9K_DIR_PATH_SEPARATOR}/g")
  fi
```
Output:
```
 ~  cd Documents
pathseparator   
current_path ~/Documents
current_path2 ~/Documents
recalc ~/Documents
 ~/Documents   
```
It's very weird because it works in the terminal :
```
 ~/Documents  echo "~/test" | sed "s/\//%f "$'\uE0B1'" %F/g"
~%f  %Ftest
```

dritter:
@V1rgul that is strange. Which OS do you use?
And could you do a `which -a sed` and `sed --version`? That would help finding the problem here.
Strange that it works in terminal, but not in the theme. It may be that the theme uses another version of sed (possibly an old version that was shipped with OSX, if you use mac).

V1rgul:
I'm using ubuntu 15.04
`which -a sed` -> `/bin/sed`
`sed --version` ->`sed (GNU sed) 4.2.2`

V1rgul:
Seem to be a problem with $current_path after all
```bash
echo "separator ${POWERLEVEL9K_DIR_PATH_SEPARATOR}"
echo "current_path ${current_path}"
echo "test $(echo "~/Documents" | sed "s/\//${POWERLEVEL9K_DIR_PATH_SEPARATOR}/g")"
echo "recalc $(echo "${current_path}" | sed "s/\//${POWERLEVEL9K_DIR_PATH_SEPARATOR}/g")"
current_path=$(echo "${current_path}" | sed "s/\//${POWERLEVEL9K_DIR_PATH_SEPARATOR}/g")
echo "current_path ${current_path}"
```
```
separator   
current_path ~/Documents
test ~  Documents
recalc ~/Documents
current_path ~/Documents
```

dritter:
@V1rgul good catch! It seems that the behavior of shadowing variables have changed..
Could you try to rename the second declaration of `current_path` to something else and see if that works?
Out of interest: which version of ZSH do you use?

V1rgul:
The 2nd declaration ? Like that ?
```
current_path2=$(echo "${current_path}" | sed "s/\//${POWERLEVEL9K_DIR_PATH_SEPARATOR}/g")
```
It doesn't work either, still outputs `~/Documents`
I don't have `$POWERLEVEL9K_SHORTEN_DIR_LENGTH` set so `$current_path` only comes from `local current_path='%~'`
I'm currently on branch next, and pulled your pull request, which is commit 11b8545f

dritter:
@V1rgul Thanks for the help. Now it should work. The problem was that `echo` seems to output the ZSH command string and not the final path. `print -P` on the other hand gives us the evaluated path.

V1rgul:
@dritter Thanks, it's working now :)

dritter:
👍

PoLoMoTo:
That fix also fixed the issue I was having.  I just have one question now, how do you change the color?

dritter:
@PoLoMoTo awesome!
Depending on the state, you could set the following variables:
```zsh
## Colors if you are in your home folder
POWERLEVEL9K_DIR_HOME_FOREGROUND='green'
POWERLEVEL9K_DIR_HOME_BACKGROUND='cyan'
# Icon color
POWERLEVEL9K_DIR_HOME_VISUAL_IDENTIFIER_COLOR='red'

## Colors if you are in a subfolder of your home folder
POWERLEVEL9K_DIR_HOME_SUBFOLDER_FOREGROUND='green'
POWERLEVEL9K_DIR_HOME_SUBFOLDER_BACKGROUND='cyan'
# Icon color
POWERLEVEL9K_DIR_HOME_SUBFOLDER_VISUAL_IDENTIFIER_COLOR='red'

## Colors if you are in a folder outside of your home folder
POWERLEVEL9K_DIR_DEFAULT_FOREGROUND='yellow'
POWERLEVEL9K_DIR_DEFAULT_BACKGROUND='black'
# Icon color
POWERLEVEL9K_DIR_DEFAULT_VISUAL_IDENTIFIER_COLOR='red'
```

Btw. you could read more about stateful segments in our [wiki](https://github.com/bhilburn/powerlevel9k/wiki/Stylizing-Your-Prompt#special-segment-colors).

bhilburn:
Nice work, @dritter! This looks awesome!

PoLoMoTo:
@dritter None of those seem to change the color of the separator though?

dritter:
@PoLoMoTo Yes, my example contains a little bit of magic. `%f` ends the foreground color and characters without a foreground color set will be printed in white. `%F` starts a new foreground color and without specifying a new color, ZSH uses the default one, which is black.

A better example is:
```zsh
POWERLEVEL9K_DIR_PATH_SEPARATOR='%F{red} $(print_icon "LEFT_SUBSEGMENT_SEPARATOR") %F{black}'
```
There are some things to know here:
  - Starting a new foreground color automatically ends the previous one
  - `print_icon` is a safer to use in this example as no hardcoded code point is used
  - Because of `print_icon`, we need to wrap the whole definition in single quotes, because `print_icon` is available if the after the theme is sourced

PR #348, opened 2016-12-03
Fixed an issue where full battery registers as disconnected

DeanF:


dritter:
@DeanF Thanks for the contribution.

@rjorgenson What do you think?

bhilburn:
As @dritter said, thanks so much for the contribution, @DeanF! 

It would be great to get @rjorgenson to weigh-in with a review before I merge, as he is most familiar with the power indicators in PL9k '=)'

bhilburn:
Actually, is this PR superceded by #366? And should this change get pulled into that PR?

@rjorgenson

rjorgenson:
@bhilburn this shouldn't conflict a the changes in #366 are all in the OS X battery section and this change is for the linux section. Not sure how I missed the original mention here, sorry about that! This change looks good to me.

bhilburn:
Awesome! Thanks, @rjorgenson!

@DeanF - Nice contribution! Thanks so much for pushing your bug fix upstream `=)`.

PR #355, opened 2016-12-11
added POWERLEVEL9K_ROOT_COLOR

shibumi:
Hey,
I have implemented a little option to specify a custom color for the root user highlight.
On my system both users look nearly equal so I thought it's a good idea.

Cheers 
chris

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,10 @@ elements (it is by default), and define a `DEFAULT_USER` in your `~/.zshrc`:
 |----------|---------------|-------------|
 |`DEFAULT_USER`|None|Username to consider a "default context" (you can also use `$USER`)|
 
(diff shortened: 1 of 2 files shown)


dritter:
Hey Chris!

Thanks for the contribution, but this is already possible. Every segments fore- and background colors can be overwritten. So here. Just set `POWERLEVEL9K_CONTEXT_ROOT_BACKGROUND="red"` and your background will be red in root context.
You can read more about stateful segments [here](https://github.com/bhilburn/powerlevel9k/wiki/Stylizing-Your-Prompt#special-segment-colors)

shibumi:
@dritter 
Ah ok. Didn't know this that this is possible! Thanks!

Issue #335, opened 2016-10-30
Conditional hostname visibility?


spicypixel-:
Is it possible to only show the user@hostname when the session is inside ssh and just user when it's local?

bhilburn:
@spicypixel - I see, so you want it just to `user` when you're local, and `user@hostname` when you are SSH'd. Yeah, that's fairly simple, and a good idea for an enhancement.

Thanks!

shibumi:
I just run powerlevel9k on my server also.. that way it will also change the user and hostname.. Do you really see a need here for an enhancement?

I also don't think thats  so simple. because when you login in to a server via SSH you are on a different system. You will have no powerlevel9k there (except you install it).

spicypixel-:
Well yeah but I tend to install it via my dotfiles repo. Just nice if at
all possible.

shibumi:
@spicypixel- 
Wouldn't it be easier to configure the powerline on your host that your host will just display the user?

spicypixel-:
Yeah though means host specific config to override the local dotfiles. It
would just be merely nice if zsh was aware if was in an SSH session and
report accordingly.

I'm currently doing that though with my local machine having a different
configuration. Just becomes less attractive if I'm locally at the remote
machine and SSHing back into the original local machine.

shibumi:
@spicypixel- 
Ah ok. Now I understand. You want that powerlevel9k is aware of it that it runs in a SSH session. Ok that makes sense. I thought you want to take your powerlevel9k configuration with you on the server. This would be nearly impossible..

spicypixel-:
Correct merely if it detects it's running in SSH to show user@hostname, and
if it's not merely user. Obviously as you said it's going to be limited to
wherever I install the .zshrc but if I'm not going through the trouble of
setting up the host changes are I won't be connecting regularly enough to
worry.

shibumi:
@spicypixel- 
I wrote a first ssh-detection prototype and virtualization prototype for powerlevel9k. You can checkout: https://github.com/bhilburn/powerlevel9k/issues/336

bhilburn:
I think the feature that @spicypixel- is asking for in this issue could easily be integrated into the work that @shibumi started over in #336.

bhilburn:
[...]

PR #356, opened 2016-12-11
added ssh support and POWERLEVEL9K_HIDE_HOST feature

shibumi:
Heyho,
here is the ssh support. It's now possible to hide the hostname.  When you SSH on another host you will see the full prompt with username@hostname. On your localmachine you will just see $USER

This should satisfy https://github.com/bhilburn/powerlevel9k/issues/335

TODO: Adding a SSH identifier in the right prompt for users who doesn't want the hide host feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,11 @@ elements (it is by default), and define a `DEFAULT_USER` in your `~/.zshrc`:
 |----------|---------------|-------------|
 |`DEFAULT_USER`|None|Username to consider a "default context" (you can also use `$USER`)|
 
+You can use POWERLEVEL9K_HIDE_HOST for hiding the hostname in the prompt
+when you are not in a ssh session. For example:
+
+    POWERLEVEL9K_HIDE_HOST="yes"
+
 ##### dir
 
 The `dir` segment shows the current working directory. When using the "Awesome
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -434,29 +434,27 @@ prompt_battery() {
 # Context: user@hostname (who am I and where am I)
 # Note that if $DEFAULT_USER is not set, this prompt segment will always print
 prompt_context() {
+  local current_state="DEFAULT"
+  declare -A context_states
+  context_states=(
+    "ROOT"      "yellow"
+    "DEFAULT"   "011"
+  )
+  local content="$USER"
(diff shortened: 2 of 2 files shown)


dritter:
This is pretty WET with a lot of if/else statements.
A much more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) approach would be to extract the state the segment is in as local variable, and use that to build up the segment. Like this:
```zsh
local current_state="DEFAULT"
typeset -Ah context_states
context_states=(
  "ROOT"        "yellow"
  "DEFAULT"  "011"
)
if [[ $(print -P "%#") == '#' ]]; then
  current_state="ROOT"
fi
local content="${USER}"
if [[ -z "$SSH_CLIENT" && -z "$SSH_TTY" ]]; then
  content="${content}@%m"
elif [[ -n "$POWERLEVEL9K_HIDE_HOST" ]]; then
  content="${content}@%m"
fi
if [[ "$USER" != "$DEFAULT_USER" ]]; then
  "$1_prompt_segment" "${0}_${current_state}" "$2" "$DEFAULT_COLOR" "${context_states[$current_state]}" "${content}"
fi
```
Disclaimer: this is untested code. May contain bugs.

Wouldn't that do pretty much the same, or did I miss something?
This is much cleaner, as you don't print the segment in every if-branch, but build up the informations needed and print the segment only once.
What do you think?

shibumi:
Hey,
@dritter checkout the newest commit. I have refactorized the code.

dritter:
@shibumi much better! 👍 

I just have a few comments:
- Could you change the base of the PR to `next`? -> this is our development branch
- Could you change the value of `POWERLEVEL9K_HIDE_HOST` to `true`, and edit the check as well? Some folks use to try such an switch out and get confused if they set the value to `false`, but that won't change anything..

bhilburn:
@shibumi - This looks awesome! Nice work.

As @dritter said, the only two changes I can see are re-filing the PR against `next`, and changing the variable & conditional to be based on `true/false` rather than `yes/no`. 

Here's an example of that conditional elsewhere in the theme: https://github.com/bhilburn/powerlevel9k/blob/master/powerlevel9k.zsh-theme#L752

Let us know if you need any help or have any questions in finishing this up!

shibumi:
@bhilburn 
I hope I can finish this friday :)

shibumi:
I will close this issue, because I changed to another target branch for the PR

PR #357, opened 2016-12-11
added virtualization detection with systemd 

shibumi:
Hey,
I have added virtualization detection, but its still **broken**. I have no idea why. When I try it with my setup the whole right prompt will be removed... dunno why. Maybe I do something wrong, with my custom_plugin it works but not with this patch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@ The segments that are currently available are:
 * [`custom_command`](#custom_command) - Create a custom segment to display the
   output of an arbitrary command.
 * [`todo`](http://todotxt.com/) - Shows the number of tasks in your todo.txt tasks file.
+* `detect-virt` - Virtualization detection with systemd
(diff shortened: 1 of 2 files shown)


dritter:
I think the problem is that you miss a parameter when you call `$1_prompt_segment`. 
The arguments for `$1_prompt_segment` are:
```zsh
#   * $1: Name of the function that was originally invoked (mandatory).
#         Necessary, to make the dynamic color-overwrite mechanism work.
#   * $2: The array index of the current segment
#   * $3: Background color
#   * $4: Foreground color
#   * $5: The segment content
#   * $6: An identifying icon (must be a key of the icons array)
```
You are passing the content as foreground color, which seems to cause a crash..

shibumi:
Fixed it. Please checkout my other PR.

Issue #359, opened 2016-12-13
Improve CVS icons


davidmpaz:
When using `POWERLEVEL9K_MODE='awesome-fontconfig'`, I think that having Github icon as VCS_GIT_ICON is not as good as it could be.

Would not be nice to show icon from respective service depending of the repo? For example I in my theme setup I have this:
```
get_icon_by_repo() {
  local remote=`git ls-remote --get-url`
  if [[ "$remote" =~ "github" ]] then
    POWERLEVEL9K_VCS_GIT_ICON='\uf113 '
  elif [[ "$remote" =~ "bitbucket" ]] then
    POWERLEVEL9K_VCS_GIT_ICON='\uf171 '
  elif [[ "$remote" =~ "gitlab" ]] then
    POWERLEVEL9K_VCS_GIT_ICON='\uf296'
  else
    POWERLEVEL9K_VCS_GIT_ICON='\uf1d3 '
  fi
}
```
this allow me to have Github where origin point to gibhub, or bitbucket icon when origin point to bitbucket. But has the drawback that it doesn't change dynamically as we change to another repo unless we open a new terminal to refresh icon value.

Could we have this included in a better way in the main stream?

Thanks in advance for such a good project

dritter:
@davidmpaz Cool Idea! 👍 

I think the best place would be in `functions/vcs.zsh` in `+vi-vcs-detect-changes()`. There we determine the right icon:
```zsh
  if [[ "${hook_com[vcs]}" == "git" ]]; then
    vcs_visual_identifier='VCS_GIT_ICON'
  elif [[ "${hook_com[vcs]}" == "hg" ]]; then
    vcs_visual_identifier='VCS_HG_ICON'
  elif [[ "${hook_com[vcs]}" == "svn" ]]; then
    vcs_visual_identifier='VCS_SVN_ICON'
  fi
```
The icon definitions should go into `functions/icons.zsh`.

Would you mind creating a pull request againt `next` branch for that? That would be awesome!

bhilburn:
Agreed with @dritter! This sounds great.

Please submit a PR against `next`! Remember that you'll need to create font points even for non-awesome configurations - make sure to leave them as `VCS_GIT_ICON` so that current configurations aren't broken.

davidmpaz:
Hi @dritter for sure!

Thanks for checking this out. I will update referenced files and add new icons definitions as explained, as soon as possible.

Cheers

davidmpaz:
[...]

PR #360, opened 2016-12-15
Icon by repo

davidmpaz:
Hi,

as talked in #359, here is what I have done. Feel free guys to refactor the code, unfortunately my bash skill were not enough :( to extract code into a function and use it in: `+vi-vcs-detect-changes()` later on.

Cheers,
David

### DIFF
--- a/functions/vcs.zsh
+++ b/functions/vcs.zsh
@@ -110,7 +110,18 @@ function +vi-hg-bookmarks() {
 
 function +vi-vcs-detect-changes() {
   if [[ "${hook_com[vcs]}" == "git" ]]; then
-    vcs_visual_identifier='VCS_GIT_ICON'
+
+    local remote=`git ls-remote --get-url`
+    if [[ "$remote" =~ "github" ]] then
+      vcs_visual_identifier='VCS_GIT_GITHUB_ICON'
+    elif [[ "$remote" =~ "bitbucket" ]] then
+      vcs_visual_identifier='VCS_GIT_BITBUCKET_ICON'
+    elif [[ "$remote" =~ "gitlab" ]] then
+      vcs_visual_identifier='VCS_GIT_GITLAB_ICON'
+    else
+      vcs_visual_identifier='VCS_GIT_ICON'
+    fi
+
   elif [[ "${hook_com[vcs]}" == "hg" ]]; then
     vcs_visual_identifier='VCS_HG_ICON'
   elif [[ "${hook_com[vcs]}" == "svn" ]]; then
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -68,6 +68,9 @@ case $POWERLEVEL9K_MODE in
       VCS_BRANCH_ICON                $'\uE220'              # 
       VCS_REMOTE_BRANCH_ICON         ' '$'\uE804 '          # 
       VCS_GIT_ICON                   $'\uE20E '             # 
+      VCS_GIT_GITHUB_ICON            $'\uE20E '             #
+      VCS_GIT_BITBUCKET_ICON         $'\uE20E '             #
+      VCS_GIT_GITLAB_ICON            $'\uE20E '             #
       VCS_HG_ICON                    $'\uE1C3 '             # 
       VCS_SVN_ICON                   '(svn) '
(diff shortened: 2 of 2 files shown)


dritter:
Wow. That was fast. Great work! 👍 
Looks good to me.

functions/vcs.zsh
```diff
@@ -110,7 +110,18 @@ function +vi-hg-bookmarks() {
 
 function +vi-vcs-detect-changes() {
   if [[ "${hook_com[vcs]}" == "git" ]]; then
-    vcs_visual_identifier='VCS_GIT_ICON'
+
+    local remote=`git ls-remote --get-url`
+    if [[ "$remote" =~ "github" ]] then
```
bhilburn: Isn't `=~` meant for regex comparisons? For direct string-to-string comparisons, I think `==` is preferred?

davidmpaz: Yes, `git ls-remote --get-url` gives me something like `git@github.com:bhilburn/powerlevel9k.git` therefore I wanted to do more like a sub-string search by only looking for: github, bitbucket or gitlab words.

If you think this can be improved, please let me know how, I would gladly change it.

bhilburn:
Thanks again for the great  PR, @davidmpaz! Very excited to have this merged.

functions/vcs.zsh
davidmpaz: Yep. I have one question also. If we would like to refactor that a bit, like extract the code in that condition (I don't like it so much now) how would you do? 

I just want the function `+vi-vcs-detect-changes()` to look simpler. I tried to put is separate function but never make it to work when returning the icon identifier, not so much bash under my belt.

davidmpaz:
Hey, thanks to you guys. It was my pleasure. Looking forward to have it upstream!

functions/vcs.zsh
dritter: @davidmpaz Returning values in ZSH is quite a pain. There are couple of ways to accomplish that (worst first):

1. Use an cache file
    Obviously this is quite some overkill for just one value, but sometimes this is the way to go.
2. Control it via return codes
    You could return an integer as return code in your function and have that interpreted on the caller side of the function. This is stupid, if you return more than `true` (0) or `false` (1) and the return values always confuse me.
3. Fill a global variable
    In the callee function, fill a global variable. This is often done with handling associative arrays.
4. Echo the value
    If you echo the value you want in the callee function, you could catch it by calling it in a subshell (`my_value=$(call_me)`).

For sure I missed some more or less worse method. But to stay on topic, you could do something like (pseudocode):
```zsh
function buildVcsVisualIdentifier() {
    local remote=$(git ls-remote --get-url 2> /dev/null)
    if [[ "$remote" =~ "github" ]] then
        vcs_visual_identifier='VCS_GIT_GITHUB_ICON'
    elif [[ "$remote" =~ "bitbucket" ]] then
        vcs_visual_identifier='VCS_GIT_BITBUCKET_ICON'
    elif [[ "$remote" =~ "gitlab" ]] then
        vcs_visual_identifier='VCS_GIT_GITLAB_ICON'
    else
        vcs_visual_identifier='VCS_GIT_ICON'
    fi

    elif [[ "${hook_com[vcs]}" == "hg" ]]; then
        vcs_visual_identifier='VCS_HG_ICON'
    elif [[ "${hook_com[vcs]}" == "svn" ]]; then
        vcs_visual_identifier='VCS_SVN_ICON'
    fi
}

function +vi-vcs-detect-changes() {
    buildVcsVisualIdentifier()
    ...
}
```

The `vcs` segment is a quite complex one. It interacts heavily with ZSHs own VCS_INFO subsystem. In that subsystem we register a couple of git-hooks (like `vi-vcs-detect-changes`). So these hooks get called by VCS_INFO before the `vcs_prompt` (segment main-) function gets called. Long story short: We cannot control the caller side of the function, that is why we set the `vcs_visual_identifier` as a global variable (which is read in `vcs_prompt`).

davidmpaz: @dritter thanks for the good explanation. I understand better now. I think is fine to leave it as it is then, in any case a refactoring would improve the cyclomatic complexity only which is not so much actually.

PR #362, opened 2016-12-17
Detect virt next

shibumi:
implemented virtualization detection with systemd.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -41,7 +41,7 @@ else
   POWERLEVEL9K_INSTALLATION_PATH="$0"
 fi
 
-# Resolve the instllation path
+# Resolve the installation path
 if [[ -L "$POWERLEVEL9K_INSTALLATION_PATH" ]]; then
   # If this theme is sourced as a symlink, we need to locate the real URL
   filename="$(realpath -P $POWERLEVEL9K_INSTALLATION_PATH 2>/dev/null || readlink -f $POWERLEVEL9K_INSTALLATION_PATH 2>/dev/null || perl -MCwd=abs_path -le 'print abs_path readlink(shift);' $POWERLEVEL9K_INSTALLATION_PATH 2>/dev/null)"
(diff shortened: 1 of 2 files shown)


powerlevel9k.zsh-theme
```diff
@@ -547,6 +547,26 @@ prompt_history() {
...
+prompt_detect_virt() {
+  if ! command -v systemd-detect-virt;then
+    return
+  fi
+  local virt=$(systemd-detect-virt)
+  local color="yellow"
+  if [[ "$virt" == "none" ]]; then
+    if [[ "$(ls -di / | grep -o 2)" != "2" ]]; then
+      virt="chroot"
+      "$1_prompt_segment" "$0" "$2" "$color" "$virt" ""
```
dritter: Here is still a parameter missing, right? Like said in #357

powerlevel9k.zsh-theme
```diff
@@ -547,6 +547,26 @@ prompt_history() {
...
+    else
+      ;
+    fi
+  else
+    "$1_prompt_segment" "$0" "$2" "$color" "$virt" ""
```
dritter: Here as well. ;)

powerlevel9k.zsh-theme
shibumi: why? the last field is for the icon. I have no icon for the virtualization, just text.

dritter: `$1_prompt_segment` takes six argument, you just provide five. The first is the name of the function (`$0`). Second is the index (`$2`). Third is the background color (`$color`). Fourth is foreground color (`$virt` in your case; this is what looks odd to me).. After that the arguments seem shifted.

dritter: Yes, but there you omitted the argument for the icon (a.k.a. visual identifier, as I call it) which is optional. In the `detect-virt` segment, you omit the argument for foreground color which is a mandatory one.
As arguments of a function are positional, omitting one is always the last one (except you pass an empty parameter to the function; in ZSH terms this is an empty string, in other programming languages this is usually `null`). So passing 5 arguments omits the icon, but you pass an empty string. On the other hand you pass `$virt` as foreground color which seems wrong to me, as this should be passed one position later as the content of the segment.

shibumi:
@dritter should be fixed now.

bhilburn:
Nice work, @shibumi!! What a great contribution.

And thanks for the review work, @dritter! As always, much appreciated.

PR #366, opened 2016-12-19
migrated battery segment to use pmset on OSX over ioreg

rjorgenson:
Initial rewrite of battery segment to use pmset. Any help testing and finding edge cases will be helpful =]

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -350,43 +350,35 @@ prompt_battery() {
   # Set default values if the user did not configure them
   set_default POWERLEVEL9K_BATTERY_LOW_THRESHOLD  10
 
-  if [[ $OS =~ OSX && -f /usr/sbin/ioreg && -x /usr/sbin/ioreg ]]; then
-    # Pre-Grep as much information as possible to save some memory and
-    # avoid pollution of the xtrace output.
-    local raw_data="$(ioreg -n AppleSmartBattery | grep -E "MaxCapacity|TimeRemaining|CurrentCapacity|ExternalConnected|IsCharging")"
+  if [[ $OS =~ OSX && -f /usr/bin/pmset && -x /usr/bin/pmset ]]; then
+    # obtain battery information from system
+    local raw_data="$(pmset -g batt)"
     # return if there is no battery on system
-    [[ -z $(echo $raw_data | grep MaxCapacity) ]] && return
+    [[ -z $(echo $raw_data | grep "InternalBattery") ]] && return
 
-    # Convert time remaining from minutes to hours:minutes date string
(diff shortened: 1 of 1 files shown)


bhilburn:
I don't have a Mac, so can't test this on my own machine. 

After using it for a bit, @rjorgenson, what do you think? Better than the existing implementation?

If anyone else is up for trying this out and sharing their opinion, please do!

rjorgenson:
So after a little over a week of using this full time I see no issues with it and haven't run into any edge cases. I did some basic speed testing comparing pmset to ioreg and didn't really see any noticeable difference there. The only difference really between the two implementations is that pmset seems to do it's own thing with what information it presents to the user. It's not the same as the data reported by ioreg but it does appear to be the source of data for the OS X system battery status bar menu. You can see in the screenshot both pmset and the system battery report 3:36 as the time remaining with 100% charge. ioreg reports the same time remining but it shows the battery capacity as 4005/4214 which is about 95%(the battery meter in the lower right corner of the screen is using ioreg). This was about 10 minutes after pulling my laptop off the charger and using it to let it discharge a bit so ioreg seems a bit more accurate and pmset seems to do some voodoo on their own to determine battery percentage. The one thing I didn't check was the ioreg capacity levels when it was fully charged. It's possible the MaxCapacity is actually a bit higher than the circuitry actually charges the battery and pmset takes that into account where ioreg does not. Personally I like that pmset shows the same data as the system battery. Even if it's less 'accurate' it wouldn't give users any reason to wonder why they are different if they end up using both. Anyone else have an opinion on that?

screen shot 2016-12-29 at 5 23 20 pm

bhilburn:
@rjorgenson - Based on your feedback, I think we should proceed with the `pmset` implementation. If we have to choose between accuracy and consistency with the OS, I would choose the latter (it sounds like you agree). It also may be that `pmset` is actually the more accurate final number - Apple does a lot of black magic with OSX power management, so it's possible that while `ioreg` might accurately reflect the actual charge counter status, `pmset` includes the PM voodoo that might make it more accurate.

Is this branch ready for merge, @rjorgenson, or do you want to do some more work on it first?

rjorgenson:
@bhilburn I agree with all points regarding pmset vs ioreg =] As far as merging, as far as I can tell it's good to go. It's possible there are still some edge cases with the output of pmset that I have been unable to run into, but none that I'm aware of. Should be good to merge =]

bhilburn:
@rjorgenson - Fantastic. Thanks for another excellent contribution!

dritter:
Thanks @rjorgenson ! I knew you'd like the neofetch stuff. Great work 👍 
And sorry for not testing it. I am quite busy with the async branch.

Issue #365, opened 2016-12-19
Adapt battery segment from neofetch


dritter:
[Neofetch](https://github.com/dylanaraps/neofetch) is a pretty awesome project. They have found an easier and (maybe) more reliable way for the [battery segment](https://github.com/dylanaraps/neofetch/blob/master/neofetch#L1579-L1657). Maybe we could adapt some of their ideas to our battery segment.

Additionally, they have some other cool stuff, like external IP, played song, etc.

rjorgenson:
I like this a lot, much cleaner than using ioreg. I did a quick rewrite and was doing some testing and noticed that it seems to do be not as exact as ioreg.

<img width="820" alt="screen shot 2016-12-19 at 3 18 21 pm" src="https://cloud.githubusercontent.com/assets/44096/21331233/876e9ff4-c5fe-11e6-89d1-c7c389be198e.png">

I'm not sure how important this is from an actual use case scenario, personally it doesn't bother me a lot assuming this is the extent of it. I'd have to do more testing on it. I'll submit a PR with my changes so others can help test edge cases and speed stuff. In my limited testing it seems to be a bit faster, but I haven't confirmed that with any tests or anything.

rjorgenson:
PR #366

bhilburn:
Merged!

PR #368, opened 2016-12-30
Public ip segment

rjorgenson:
@bhilburn mentioned in #365 that he would be interested in a public ip segment, i hacked one together for review/discussion on behavior/icons/etc. Currently it will only poll for a new IP every 5 minutes which is configurable by the user. It saves the IP to a tmp file also configurable by the user. It will attempt to use dig to get your IP if it's installed, if not it will use either curl or wget to poll an IP checker host in the same manner that neofetch uses. Let me know what you all think.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -428,7 +428,7 @@ prompt_battery() {
 
 prompt_public_ip() {
   # set default values for segment
-  set_default POWERLEVEL9K_PUBLIC_IP_TIMOUT "300"
+  set_default POWERLEVEL9K_PUBLIC_IP_TIMEOUT "300"
   set_default POWERLEVEL9K_PUBLIC_IP_NONE ""
   set_default POWERLEVEL9K_PUBLIC_IP_FILE "/tmp/p9k_public_ip"
   set_default POWERLEVEL9K_PUBLIC_IP_HOST "http://ident.me"
@@ -439,9 +439,10 @@ prompt_public_ip() {
     typeset -i timediff
     # if saved IP is more than
     timediff=$(($(date +%s) - $(date -r $POWERLEVEL9K_PUBLIC_IP_FILE +%s)))
-    [[ $timediff -gt $POWERLEVEL9K_PUBLIC_IP_TIMOUT ]] && refresh_ip=true
+    [[ $timediff -gt $POWERLEVEL9K_PUBLIC_IP_TIMEOUT ]] && refresh_ip=true
     # If tmp file is empty get a fresh IP
-    [[ -z $(cat $POWERLEVEL9K_PUBLIC_IP_FILE) || $(cat $POWERLEVEL9K_PUBLIC_IP_FILE) =~ "$POWERLEVEL9K_PUBLIC_IP_NONE" ]] && refresh_ip=true
+    [[ -z $(cat $POWERLEVEL9K_PUBLIC_IP_FILE) ]] && refresh_ip=true
+    [[ -n $POWERLEVEL9K_PUBLIC_IP_NONE ]] && [[ $(cat $POWERLEVEL9K_PUBLIC_IP_FILE) =~ "$POWERLEVEL9K_PUBLIC_IP_NONE" ]] && refresh_ip=true
   else
     touch $POWERLEVEL9K_PUBLIC_IP_FILE && refresh_ip=true
   fi
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -426,6 +426,51 @@ prompt_battery() {
   fi
 }
 
+prompt_public_ip() {
+  # set default values for segment
+  set_default POWERLEVEL9K_PUBLIC_IP_TIMOUT "300"
+  set_default POWERLEVEL9K_PUBLIC_IP_FILE "/tmp/p9k_public_ip"
+  set_default POWERLEVEL9K_PUBLIC_IP_HOST "http://ident.me"
+
+  # Do we need a fresh IP?
+  local refresh_ip=FALSE
+  if [[ -f $POWERLEVEL9K_PUBLIC_IP_FILE ]]; then
+    typeset -i timediff
+    timediff=$(($(date +%s) - $(date -r $POWERLEVEL9K_PUBLIC_IP_FILE +%s)))
+    [[ $timediff -gt '500' ]] && refresh_ip=TRUE
+    # this will run the IP refresh with each new prompt while disconnected
+    # but will get a new IP immediately once reconnected rather than waiting
+    # for the timeout, not sure if this is ideal behavior or not
+    [[ -z $(cat $POWERLEVEL9K_PUBLIC_IP_FILE) ]] && refresh_ip=TRUE
+  else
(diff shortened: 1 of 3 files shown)


rjorgenson:
currently there's no default icon, i think it looks best without one but then I couldn't really find one that seemed appropriate, maybe with an appropriate one it would look better.

rjorgenson:
I also test killing the internet connection and it will still attempt to get a new IP at each timeout interval but it fails silently. This was the best behavior I could work out for not having a connection to the internet, input is very welcome with regards to how to handle that =]

powerlevel9k.zsh-theme
```diff
@@ -426,6 +426,51 @@ prompt_battery() {
...
+
+  # Do we need a fresh IP?
+  local refresh_ip=false
+  if [[ -f $POWERLEVEL9K_PUBLIC_IP_FILE ]]; then
+    typeset -i timediff
+    timediff=$(($(date +%s) - $(date -r $POWERLEVEL9K_PUBLIC_IP_FILE +%s)))
+    [[ $timediff -gt '500' ]] && refresh_ip=true
+    # this will run the IP refresh with each new prompt while disconnected
+    # but will get a new IP immediately once reconnected rather than waiting
+    # for the timeout, not sure if this is ideal behavior or not
```
bhilburn: I think this is a good default behavior - let's roll with it.

powerlevel9k.zsh-theme
```diff
@@ -426,6 +426,51 @@ prompt_battery() {
...
+    fi
+
+    if [[ -z "$fresh_ip" ]] && type -p wget >/dev/null; then
+        fresh_ip="$(wget -T 10 -qO- "$POWERLEVEL9K_PUBLIC_IP_HOST" 2> /dev/null)"
+    fi
+    [[ -n $fresh_ip ]] && echo $fresh_ip > $POWERLEVEL9K_PUBLIC_IP_FILE
+  fi
+
+  # write IP to tmp file
+  local public_ip=$(cat $POWERLEVEL9K_PUBLIC_IP_FILE)
```
bhilburn: I might be mis-reading this, but the comment on L466 and the code on L467 are confusing me a bit. Doesn't this code just readout the IP, not write it into the file?

bhilburn:
Nice work, @rjorgenson! Very cool to see this get put together so quickly, and I think this will be really useful for our sysadmin users out there.

I just left some questions in a small review =)

powerlevel9k.zsh-theme
rjorgenson: Yes this comment is supposed to be a few lines up. Will fix, sorry about that =]

powerlevel9k.zsh-theme
rjorgenson: So I forgot to reread this section before pushing, I added a conditional to line 463 which makes this not behave exactly like this anymore. As it's written currently it will always create the tmp file if it doesn't exist and it will always check for a new ip if the tmp file is empty but it will no longer overwrite the tmp file with an empty result. If a fresh ip grab is attempted and none of the checks are successful then it will leave the tmp file alone. For example your connection to the internet dies, the segment will continue on doing nothing until the timeout is reached, at which point the prompt will attempt to grab a new IP every time the prompt segment is generated. I like this behavior but I think maybe we could add a touch to the file if the check is unsuccessful, that way it will still only check every X seconds even when your connection is down. What do you think?

bhilburn: I'm good with that plan. I think the important behaviors are that:

* If the connection is lost, on the next update, the segment no longer displays an IP.
* If the connection comes back, the user does not need to manually kick the segment to bring the IP back.

What you described, above, achieves that. I think if you update the comments to reflect the actual behavior, we're good to go.

rjorgenson: Hmm, it should actually still display your last IP once you lose connection. I like that though. let me see what I can do to get that functionality in. I updated the comments already and added the behavior to touch the file to update it's modified time.

rjorgenson:
@bhilburn I updated the offline behavior to the best solution I could get without removing the timeout feature altogether. If you are kicked offline it will continue to show your last known IP until the timeout expires, after that it will display nothing by default, but this can be configured by the user and continue to check for a new IP each time the segment is generated until it gets an IP and the timeout will reset. A user can set the timeout to 0 to have a real time polling with every new prompt of their public IP as well. Then they would notice as soon as they get disconnected.

If this behavior sounds good to you then I'll write up a section in the README on it. I'm still not sure on an icon for this segment, or if it even needs one. Any input on that?

bhilburn:
@rjorgenson - The new changes look great, and the final scheme sounds good. Let's proceed with this.

As for an icon, I just looked through all of the code points in Font Awesome and the standard UTF icons, and didn't see anything obvious. Go ahead and create the icon point in P9k, but we can leave it blank until we come up with something - I'm okay merging it without an icon.

@dritter - Any ideas?

rjorgenson:
Okay I've written up the README section. Writing isn't my strongest so if any of that could be changed to make more sense let me know =]

I've also added the ability to specify a particular method and only ever attempt to use that method. This may save a bit of time if you know a particular method will always be available.

rjorgenson:
Whew, that was a lot of issue I noticed after my last comment =/ I think I got everything though .. i hope lol

dritter:
As this segment is related to the (internal) `ip` segment (which we should rename at some point), I have a couple quick ideas regarding the icon:

1. We could use the [signal](http://fontawesome.io/icon/signal/) icon, as the internal ip segment (mis)uses the [rss](http://fontawesome.io/icon/rss/) icon (at least in `awesome-patched` mode). Patched-Codepoint is `\uE129`.
bildschirmfoto 2017-01-10 um 19 01 59

2. We could use the [wifi](http://fontawesome.io/icon/wifi/) icon, but that would work in `awesome-fontconfig` mode only (not in patched font included).

3. We could use a "radio tower" (don't know the name) icon (`\uE830` in patched), but we should save that for a gnuradio segment ( 😉  @bhilburn ):
bildschirmfoto 2017-01-10 um 19 26 59

4. We could use the [rss](http://fontawesome.io/icon/rss/) icon as well, but set a different default color for the visual identifier.

But I am good to go with no icon as well.

bhilburn:
Hah! Thanks for the `gnuradio` reference, @dritter!

I'm so-so on our icon choices so far - to me, they all imply a wireless connection, which may not be the case.

What do folks think about just using the `@` sign? In lieu of that, I think we should just go with no icon.

dritter:
Yes, I agree. These are not the best symbols, but there are no ethernet icons in awesome terminal fonts. ;)

I am ambivalent with the `@` sign, as this must stand before the address, but the visual identifier stands before the segment only in left aligned ones. We could (of course) put that into the segments content, but then it would be hard to override.

bhilburn:
@dritter - Great point. The `@` sign isn't a great choice.

Let's roll without an icon for now, @rjorgenson.

rjorgenson:
No icon sounds good to me =] Just went over it again and pretty sure it's ready to merge. If anyone else wants to pull this PR and do some user testing in case I missed any bits that would probably be helpful =]

bhilburn:
Thanks for another great contribution, @rjorgenson! Excited to have this merged.

PR #369, opened 2016-12-31
Context host depth

rjorgenson:
I added a feature to the context segment to allow for customizable host depth on the prompt as outlined in the ZSH documentation here - http://zsh.sourceforge.net/Doc/Release/Prompt-Expansion.html#Login-information. I use p9k across a range of servers across a range of sites at work and being able to customize the host depth to include the site is immensely helpful. Thought this might be useful to other folks too!


bhilburn:
This is awesome, @rjorgenson! What a great idea.

PR #376, opened 2017-01-18
Fix MacOS specific battery segment logic.

jshort:
The case statement was causing it to never display that it was charging
and some of the awk statements had an off-by-one bug to grab the correct
values.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -339,7 +339,7 @@ prompt_background_jobs() {
 
 prompt_battery() {
   # The battery can have four different states - default to 'unknown'.
-  local current_state="unknown"
+  local current_state='unknown'
   typeset -AH battery_states
   battery_states=(
     'low'           'red'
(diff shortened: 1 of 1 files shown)


jshort:
Hmm @rjorgenson, looking at https://cloud.githubusercontent.com/assets/44096/21557670/10a0e2dc-cded-11e6-84a8-157c8efd89a3.png

I see that your `pmset -g batt` output is subtly different than mine (wrt to the id=xxxxxxx part) so I think the simple `awk` parsing isn't going to be sufficient.

jshort:
Ok check out the last commit.  I changed it to split on `;` first and then on whitespace if need be.  This will work on `pmset -g batt` output both like mine:

 ```
❯  pmset -g batt                                                                              
Now drawing from 'AC Power'
 -InternalBattery-0 	43%; charging; 2:13 remaining present: true
```

And if (id=xxxxxxxx) is in the `pmset` output like on @rjorgenson's machine.

rjorgenson:
@jshort tested your most recent commit and I didn't run into any problems on my end with the extra id column on pmset. Good catch and great job on the fix! =]

@bhilburn This should probably be merged into next before release because it appears the initial migration to pmset that I did doesn't work on some older versions of OS X, I was only able to test on Sierra but it breaks on El Capitan and presumably versions before that as well.

bhilburn:
@rjorgenson - Thanks for the review & recommendation! Much appreciated.

@jshort - Fantastic contribution! Thanks for getting involved. Happy to have this merged =)

PR #377, opened 2017-01-20
Change powerlevel9k_init to prompt_powerlevel9k_setup

mcgillowen:
Changed init/setup function name to be compatible with zsh promptinit, zsh prompt function only lists themes with prompt_*_setup init/setup functions.

Allows use of this prompt directly with zsh's prompt system by running: 
`prompt powerlevel9k`
at the zsh command line

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+## v0.5.0
+
+### `load` and `ram` changes
+
+These two segments now support BSD.
+
+### `vcs` changes
+
+- We implemented a huge speed improvement for this segment.
+- Now this segment supports Subversion repositories.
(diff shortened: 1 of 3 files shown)


dritter:
Good idea! Thanks for the PR! 👍 

Only thing is: Our development branch is `next`. Could you change the base of your PR to that branch? That would be awesome!

bhilburn:
@mcgillowen - Thanks for the bug fix, and for taking the time to submit a PR! I really appreciate it.

Because your original change was made on top of `master`, when you changed the merge target to `next` it actually carried other changes with it (hence the conflict reported by the CI system). So, I cherry-picked your change on to next and pushed it. You can see your commit (e2b2e8e17) in the tree here:
https://github.com/bhilburn/powerlevel9k/commits/next

Thanks again for getting involved!

PR #352, opened 2016-12-09
Anion's `hdd_usage` Segment

bhilburn:
Hey @anion155! @dritter spotted your custom `hdd_usage` segment in your fork, and we think this would be a great candidate to upstream! Creating this PR to get the discussion going.

This PR is currently based on `master`, but needs to be rebased against `next`.

PR #378, opened 2017-01-21
Anion155's hdd_usage segment

bhilburn:
Okay, this is the revamped version of #352. I cherry-picked @anion155's commits onto a new branch and modified the behavior somewhat. It currently shows the usage level of the partition that your CWD resides on. I think it would be nice to make it possible for users to peg the segment to a specific partition, but that can be a future enhancement. Here is what it looks like:

hdd_usage

It's working fine on my Linux system. It would be great to have a test from an OSX user, as well as a general code review.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -372,7 +372,7 @@ prompt_hdd_usage() {
   fi
 
   current_state='normal'
-  local message="$level%"
+  local message="$level% "
 
   if [ $level -le $POWERLEVEL9K_HDD_USAGE_WARNING_LEVEL ]; then
     current_state='warning'
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -337,10 +337,10 @@ prompt_background_jobs() {
   fi
 }
 
-# Segment to indicate hdd available level.
+# Segment that indicates usage level of current partition.
 set_default POWERLEVEL9K_HDD_USAGE_ONLY_WARNING false
-set_default POWERLEVEL9K_HDD_USAGE_WARNING_LEVEL 10
-set_default POWERLEVEL9K_HDD_USAGE_CRITICAL_LEVEL 5
+set_default POWERLEVEL9K_HDD_USAGE_WARNING_LEVEL 90
+set_default POWERLEVEL9K_HDD_USAGE_CRITICAL_LEVEL 95
 prompt_hdd_usage() {
   local current_state="unknown"
   typeset -AH hdd_usage_forecolors
@@ -356,28 +356,20 @@ prompt_hdd_usage() {
     'critical'      'red'
   )
 
-  # local df="$(df -k . | sed -n '2p')"
-  # local size="$(echo $df | awk '{ print $2 }')"
-  # local avail="$(echo $df | awk '{ print $4 }')"
-  # local level="$(printf %.0f $(( avail * 100.0 / size )))"
-
   local level="${$(df -k . | sed -n '2p' | awk '{ print $5 }')%%\%}"
-  level="$(( 100 - level ))"
 
-  if [ "$level" -gt "$POWERLEVEL9K_HDD_USAGE_WARNING_LEVEL" ]; then
-    # Default behavior: Show message always
-    if [[ "$POWERLEVEL9K_HDD_USAGE_ONLY_WARNING" != false ]]; then
-      return
-    fi
-  fi
-
-  if [ "$level" -le "$POWERLEVEL9K_HDD_USAGE_CRITICAL_LEVEL" ]; then
-    current_state='critical'
-  elif [ "$level" -le "$POWERLEVEL9K_HDD_USAGE_WARNING_LEVEL" ]; then
+  if [ "$level" -ge "$POWERLEVEL9K_HDD_USAGE_WARNING_LEVEL" ]; then
     current_state='warning'
+    if [ "$level" -ge "$POWERLEVEL9K_HDD_USAGE_CRITICAL_LEVEL" ]; then
+        current_state='critical'
+    fi
   else
+    if [[ "$POWERLEVEL9K_HDD_USAGE_ONLY_WARNING" != false ]]; then
+        return
+    fi
     current_state='normal'
   fi
+
   local message="${level}%%"
 
   # Draw the prompt_segment
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -338,6 +338,9 @@ prompt_background_jobs() {
 }
 
 # Segment to indicate hdd available level.
+set_default POWERLEVEL9K_HDD_USAGE_ONLY_WARNING false
+set_default POWERLEVEL9K_HDD_USAGE_WARNING_LEVEL 10
+set_default POWERLEVEL9K_HDD_USAGE_CRITICAL_LEVEL 5
 prompt_hdd_usage() {
   local current_state="unknown"
   typeset -AH hdd_usage_forecolors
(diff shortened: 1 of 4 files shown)


rjorgenson:
df on OS X appears to be different than on linux, the statistics are inverted.

screen shot 2017-01-20 at 9 03 09 pm

it's also throwing critical for some reason, it might be because the output of df -k . | sed -n '2p' | awk '{ print $5 }' is 15% and not 15.

I also tried this out on my Ubuntu 16 box with df (GNU coreutils) 8.25 and the usage percent was column 6, not column 5

screen shot 2017-01-20 at 9 14 21 pm

rjorgenson:
For what it's worth I checked out how neofetch(which dritter pointed out to me earlier for the battery segment updates) handles this and it also assumes column 5 for all operating systems, but it works properly on my ubuntu 16 box with the above df output, i was unable to determine why. Maybe I'm just out of it and missing something obvious lol.

bhilburn:
@rjorgenson - Weird. Okay, so, the `df -k . | sed -n '2p' | awk '{ print $5 }'` code does yield a percent on my system, too, but the full line strips that to yield just the number: 

```zsh
local level="${$(df -k . | sed -n '2p' | awk '{ print $5 }')%%\%}"
```

At least, that works on my local Linux / ZSH system:

df-output

Very obnoxious that `df` is different between Linux and OSX, but, again, perhaps not surprising. It looks like `neofetch` dealt with this problem, too, and conditionalized the flags:

```
    case "$os" in
        "Haiku") err "Disk doesn't work on Haiku due to the non-standard 'df'"; return ;;
        "Mac OS X") df_flags=(-P -h) ;;
        "AIX") df_flags=(-P -g) ;;
        *) df_flags=(-h) ;;
    esac
```

Does using `-P -h` on your OSX system work?

That said, it is really strange that the neofetch version works on your Ubuntu system and this one doesn't. I agree that neofetch appears to be indexing the 5th column for the usage percent:

```
        disk_perc="${disk_info[4]/'%'}"
```

What strange is that your Ubuntu system seems to be outputting the `Type` column even though you didn't ask for it. Is there any chance you have an alias setup for `df` or something?

Thanks for the testing, by the way! Much appreciated.

rjorgenson:
Hmm, so you were dead on in both cases. My ubuntu box does indeed have an alias setup, a good workaround for this would be to call /bin/df directly the way neofetch does rather than use whatever the env returns for df.

I tried the line actually setting level and it does set it without the %, in which case I have no idea why it's triggering a critical with 15 >= 95. We'll still need an OS check to flop the level

`[[ $OS =~ "OSX" ]] && level=$(( 100 - $level ))`

I tested this and it gave the expected state. Though looking over the code again it seems the warning and critical thresholds are set to expect level to be disk usage, not disk free. 

`if [ "$level" -ge "$POWERLEVEL9K_HDD_USAGE_WARNING_LEVEL" ]; then`
`    current_state='warning'`
`    if [ "$level" -ge "$POWERLEVEL9K_HDD_USAGE_CRITICAL_LEVEL" ]; then`
`        current_state='critical'`
`    fi`

I tried inverting this again by adding a ! to the above conditional, and completely commenting it out, and I do not get a critical level anymore. I am at a complete loss on this but am willing to just account this to user error in some way lol. So if we reference the df command specifically to bypass aliases and sort out wether we want it to display disk usage(which makes sense given the segment name) or disk free, and account accordingly by flopping the level where appropriate it should be good.

It may just be my patched font, but uF0A0 doesn't exist for me, I have it set to eE1AE which appears to be the same icon.

screen shot 2017-01-21 at 2 05 16 pm

powerlevel9k.zsh-theme
```diff
@@ -337,6 +337,47 @@ prompt_background_jobs() {
...
+
+  local level="${$(df -k . | sed -n '2p' | awk '{ print $5 }')%%\%}"
+
+  if [ "$level" -ge "$POWERLEVEL9K_HDD_USAGE_WARNING_LEVEL" ]; then
+    current_state='warning'
+    if [ "$level" -ge "$POWERLEVEL9K_HDD_USAGE_CRITICAL_LEVEL" ]; then
+        current_state='critical'
+    fi
+  else
+    if [[ "$POWERLEVEL9K_HDD_USAGE_ONLY_WARNING" != false ]]; then
```
dritter: I would switch the check here to test if `POWERLEVEL9K_HDD_USAGE_ONLY_WARNING` was set to `true`. This is more intuitive, IMHO.

powerlevel9k.zsh-theme
```diff
@@ -337,6 +337,47 @@ prompt_background_jobs() {
...
+    'critical'      'white'
+  )
+  typeset -AH hdd_usage_backcolors
+  hdd_usage_backcolors=(
+    'normal'        $DEFAULT_COLOR
+    'warning'       'yellow'
+    'critical'      'red'
+  )
+
+  local level="${$(df -k . | sed -n '2p' | awk '{ print $5 }')%%\%}"
```
dritter: `level` is a bit ambiguous. Why not `disk_usage` or `capacity`?

powerlevel9k.zsh-theme
```diff
@@ -337,6 +337,47 @@ prompt_background_jobs() {
...
+
+  if [ "$level" -ge "$POWERLEVEL9K_HDD_USAGE_WARNING_LEVEL" ]; then
+    current_state='warning'
+    if [ "$level" -ge "$POWERLEVEL9K_HDD_USAGE_CRITICAL_LEVEL" ]; then
+        current_state='critical'
+    fi
+  else
+    if [[ "$POWERLEVEL9K_HDD_USAGE_ONLY_WARNING" != false ]]; then
+        return
```
dritter: In preparation for merging that in #344 , it would be better to set `level` to an empty value. It won't get printed anyway, as this is checked later on.

powerlevel9k.zsh-theme
dritter: This block could be simplified, like this:
```zsh
local level="${$(df -k . | sed -n '2p' | awk '{ print $5 }')%%\%}"

# Default state is "normal" unless POWERLEVEL9K_HDD_USAGE_ONLY_WARNING
# is set to true, then it is "unknown"
if [[ "$POWERLEVEL9K_HDD_USAGE_ONLY_WARNING" != true ]]; then
  current_state='normal'
fi

if [ "$level" -ge "$POWERLEVEL9K_HDD_USAGE_WARNING_LEVEL" ]; then
  current_state='warning'
elif [ "$level" -ge "$POWERLEVEL9K_HDD_USAGE_CRITICAL_LEVEL" ]; then
  current_state='critical'
fi
```

dritter:
> df on OS X appears to be different than on linux, the statistics are inverted.
@rjorgenson What do you mean by that?
Referencing your screenshot, 33Gi of 233Gi is about 15%, no?

For `awesome-patched` I agree, the icon should be `$'\uE1AE`.

`df -P -h` work on OSX Sierras original `/bin/df` as well as GNU coreutils 8.26s `gdf`:
bildschirmfoto 2017-01-22 um 00 16 20

rjorgenson:
@dritter I must be going crazy, for some reason I was reading linux df as space free and os x as space used but looking at it now, they are both space used. I'm gonna go lay down now =[

powerlevel9k.zsh-theme
bhilburn: Agreed!

powerlevel9k.zsh-theme
bhilburn: Good idea.

powerlevel9k.zsh-theme
bhilburn: :+1:

bhilburn:
So, actually, `-P` also works on Linux. I'm thinking I'll just make them all the same, and let's give it a go.

powerlevel9k.zsh-theme
bhilburn: So the problem with that structure is that the 'CRITICAL_LEVEL` conditional would never return `true`. Because if its `>=` the critical level, it's also `>=` the warning level, and the critical level check is an `else if`, it will never be true. I think that the way that I have it now minimizes the number of conditionals that are evaluated for any one state.

I'm totally open to other structures, though - I agree that the flow is a bit confusing :-/

bhilburn:
@dritter @rjorgenson - Okay, branch updated. Can you give it a go?

1) I changed the code-point for awesome-patched to `\uE1AE`. That shows up for me as an `omega`, so let me know if that came out right
2) Incorporated changes from @dritter's review.
3) So you can bypass aliases if you prefix a command with a backslash. @rjorgenson see if it works on your Ubuntu machine, now.
4) Rename the segment to `disk_usage`, updated docs accordingly.

rjorgenson:
@bhilburn Yup, all good on the icon and the alias bypass.

bhilburn:
@rjorgenson - Happy to hear it!

@rjorgenson @dritter - We might actually want to start using that method to bypass aliases in other places, too, just in case.

powerlevel9k.zsh-theme
dritter: Hmm. I was too tired yesterday as well.. Now I am too late, but this control structure should work:
local level="${$(df -k . | sed -n '2p' | awk '{ print $5 }')%%\%}"
```zsh
# Default state is "normal" unless POWERLEVEL9K_HDD_USAGE_ONLY_WARNING
# is set to true, then it is "unknown"
if [[ "$POWERLEVEL9K_HDD_USAGE_ONLY_WARNING" != true ]]; then
  current_state='normal'
fi

if [ "$level" -ge "$POWERLEVEL9K_HDD_USAGE_CRITICAL_LEVEL" ]; then
  current_state='critical'
elif [ "$level" -ge "$POWERLEVEL9K_HDD_USAGE_WARNING_LEVEL" ]; then
  current_state='warning'
fi
```

PR #379, opened 2017-01-22
Fix swift segment

dritter:
The regex for the swift segment matches too much (matched the "86" from `x86` as second). Here I simplified it to match the first number as this is probably the "main" version number. I assume the original creator of this segment (@thomaspaulmann) wanted the first number only (see https://twitter.com/thomaspaulmann/status/806078915546664960).

Before it matched:
bildschirmfoto 2017-01-22 um 01 50 58

The new one just matches:
bildschirmfoto 2017-01-22 um 01 52 00

@thomaspaulmann This is probably the reason, why you coded `swift_version` as array. But this way it is simpler IMHO.

Disclaimer:
I found this issue in #344 and couldn't reproduce it there in a proper unit test. But it prevented other segments from rendering.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1065,11 +1065,9 @@ prompt_pyenv() {
 
 # Swift version
 prompt_swift_version() {
-  local swift_version=($(swift --version 2>/dev/null))
+  # Get the first number as this is probably the "main" version number..
+  local swift_version=$(swift --version 2>/dev/null | grep -o -E "[0-9.]+" | head -n 1)
   [[ -z "${swift_version}" ]] && return
-
-  # Extract semantic version
-  swift_version=$(echo ${swift_version} | sed -e 's/[^0-9.]*\([0-9.]*\).*/\1/')
 
   "$1_prompt_segment" "$0" "$2" "magenta" "white" "${swift_version}" 'SWIFT_ICON'
 }


bhilburn:
@dritter - Nice catch. It looks like we may once again be dealing with the complications of different `sed` implementations on different OSes. I don't have swift installed, but echoing a bunch of stuff into that `sed` line works for me: 

sed

When first testing, your version was slower than the existing version, but it was because of the complex alias I have set-up for `grep` for code searching. If I escape it (`\grep`), as we discovered in the `disk_usage` development, it's the same speed. It may be worth doing the same thing for `head`, actually.

Thoughts?

dritter:
The thing is that the output of `swift --version` contains a newline. So if I do a `echo "asdf 3.0.2\nx86 4.4.4" | sed -e 's/[^0-9.]*\([0-9.]*\).*/\1/'` it will output:
```
3.0.2
86
```
This happens on OSX and Linux, so it is not a `sed` thing here.

You can find the test [here](https://github.com/bhilburn/powerlevel9k/pull/344/commits/3c7d0c0d7053a2a4a14a1853f8dd75c787988cd2) (forgot to commit the test for that segment earlier). But, as said, the test works mysteriously..

dritter:
Btw. just tested this on `next` branch. Interestingly the segment works there. This is a problem in combination of how I implemented segment rendering in #344.

thomaspaulmann:
@dritter great catch!!! Thanks for contributing the fix for the regex. Looks awesome to me.

bhilburn:
As @thomaspaulmann said, nice work, @dritter =)

PR #385, opened 2017-01-24
update fontconfig icons, using awesome-terminal-fonts maps

### LINKED ISSUE #382: Dealing with `awesome-fontconfig` codepoints
Okay, it looks like we might have a code-point issue with some "awesome" fonts.

While doing some testing recently, I switched over to `awesome-fontconfig` and the `OK_ICON` was actually the [medium.com](medium.com) logo. I'm not the only person it impacted - as an example, check out this bug report that just got filed, #375, and note that where there should be a checkmark there is actually the medium.com logo.

Per @V1rgul's comment, here, though, on his font the other icons were preferrable:
https://github.com/bhilburn/powerlevel9k/commit/95f0c3c6ecd2fc1edfd1b768150237d682426890#commitcomment-20573173

Need to do some digging. I suspect that the solution is going to be that some people need to put manual overrides for certain ICONs in their configs. Not a huge deal, but it is an additional step, so we should try to make it work for as many people as possible out-of-the-box.

It would be great if more people could comment, here, on whether or not the old way or the new way is better for them. **Please also describe how you installed your fonts, which font you are using, and which PL9k you are using.**

@/all
@V1rgul @dritter @rjorgenson @Tritlo @ryanoasis @lcorsini

V1rgul:
fixes #382

### DIFF
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -86,9 +86,13 @@ case $POWERLEVEL9K_MODE in
   'awesome-fontconfig')
     # fontconfig with awesome-font required! See
     # https://github.com/gabrielelana/awesome-terminal-fonts
-    source ~/.fonts/fontawesome-regular.sh
-    # source ~/.fonts/devicons-regular.sh # no named codepoints
-    source ~/.fonts/octicons-regular.sh
+
+    # if not defined, set recommended linux path
+    typeset -p "POWERLEVEL9K_FONTAWESOME_PATH" > /dev/null 2>&1  || POWERLEVEL9K_FONTAWESOME_PATH=~/.fonts  
+
+    source "$POWERLEVEL9K_FONTAWESOME_PATH/fontawesome-regular.sh"
+    # source "$POWERLEVEL9K_FONTAWESOME_PATH/devicons-regular.sh" # no named codepoints
+    source "$POWERLEVEL9K_FONTAWESOME_PATH/octicons-regular.sh"
 
     icons=(
       LEFT_SEGMENT_SEPARATOR         $'\uE0B0'                                      #        
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -141,7 +141,8 @@ case $POWERLEVEL9K_MODE in
       RUST_ICON                      $'\uE6A8'                                      #  
       PYTHON_ICON                    $'\U1F40D'                                     # 🐍
       SWIFT_ICON                     $'\uE655'                                      # 
-      PUBLIC_IP_ICON                 '\u'$CODEPOINT_OF_AWESOME_GLOBE                # 
+      PUBLIC_IP_ICON                 '\u'$CODEPOINT_OF_AWESOME_GLOBE                # 	  
+      LOCK_ICON                      '\u'$CODEPOINT_OF_AWESOME_LOCK                 # 
     )
   ;;
   *)
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -83,61 +83,65 @@ case $POWERLEVEL9K_MODE in
   'awesome-fontconfig')
     # fontconfig with awesome-font required! See
     # https://github.com/gabrielelana/awesome-terminal-fonts
+    source ~/.fonts/fontawesome-regular.sh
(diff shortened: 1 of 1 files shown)


bhilburn:
Thanks so much for making this PR, @V1rgul! I know you made it a couple of weeks ago as a result of our discussion in #382.

While this change fixes the code-point problem for the awesome-fontconfig, it actually further breaks other forms of awesome fontconfigs (baked-in fonts) that have different code-points that the patch.

What I'm planning to do at this point is merge this PR in and create a new font-group option for the baked-in option. This PR will get pushed up to `next` as part of a larger changeset, soon.

Thanks so much for putting in the effort to make this PR, and then keeping up with it over the last couple weeks. I really appreciate it =)

bhilburn:
Haven't dropped this, by the way. Will be going in, in a slightly different form, soon...

bhilburn:
@V1rgul - I just pushed some changes to `next` regarding nerdfonts, and am trying to sort out how to best work in the changes in this PR. An OS question:

This relies on the formal installation of Awesome Fontconfig in ~/.fonts, but I only see that as the recommended installation method on Linux. Does this work on OSX?
OSX Installation Instructions for Awesome Fontconfig: https://github.com/gabrielelana/awesome-terminal-fonts#how-to-install-osx

rhabbachi:
Is it possible to set the path to the mapping files as an overridable variable(s)? I agree that's the safest setting as detailed by the awesome fonts README, but I'm on Linux and I have opted to install an awesome-terminal fonts package which will have a different mapping file path.

V1rgul:
@rhabbachi @bhilburn I fixed the conflicts and implemented a configurable fontawesome path ;)

V1rgul:
@bhilburn does it need anything more before merging ?

bhilburn:
Sorry for being AWOL on this, @V1rgul. I decided to wait until after `v.0.6.0` to get this out, which I neglected to mention in this thread. Also, thanks for updating with a configurable path! That was actually one of my concerns.

The other primary thing I'm concerned about is this breaking other font types using the `awesome-fontconfig` but which aren't actually "Font Awesome". As an example, I use "Source Code Pro", which actually has most of the same code points as "Powerline Font Awesome", and thus works will `awesome-fontconfig` right now - this PR would break that, though, as I don't have the font map.

If this is the "one true way" to do code points with "Font Awesome"-based fonts, then perhaps it's time to break out "Source Code Pro" and other built-in fonts into their own font option. I certainly want to make the use of the 'awesome-*' fonts as easy as possible.

What are your thoughts on this?

dritter:
@bhilburn I am on @V1rgul 's side. 
The problem is that we use the internal API of `awesome-terminal-fonts` (= the codepoints), but we should use the public API (= the variables). Because of the absoluteness of the codepoints it might be broken in some circumstances. If a user still uses an old version of `awesome-terminal-fonts` he would see wrong glyphs now (since we switched the codepoints). And that might happen in the future as well.
And btw. you are using "SouceCode Pro" as a font, but that is just for text and powerline icons, right? The other icons that are displayed in the prompt are taken from "FontAwesome". That is done by fontconfig. [Here](http://adobe-fonts.github.io/source-code-pro/) is a list of glyphs in SourceCodePro.

> If this is the "one true way" to do code points with "Font Awesome"-based fonts, then perhaps it's time to break out "Source Code Pro" and other built-in fonts into their own font option. I certainly want to make the use of the 'awesome-*' fonts as easy as possible.

I agree, maybe we need a new option for "SourceCode Pro". But for `awesome-terminal-fonts` we should stick to the default way, and that is through the font maps.

bhilburn:
Thanks for the feedback, everyone!

Font-maps will be how we proceed, based on input from @V1rgul and @dritter. I will work on integrating this ASAP.

bhilburn:
Just a note that I haven't forgotten about this. Just haven't finished integrating & testing =)

rhabbachi:
Any news for this? Been using it for a while now and works great for me.

bhilburn:
@rhabbachi - We are working on getting it merged in #472 =)

@V1rgul - I'm actually going to close this PR since all of your commits are in #472, anyway.

PR #386, opened 2017-01-24
Avoid error if CLOBBER is not set

dritter:
If the tempfile already exists and CLOBBER is not set, a file exists error occurrs. This commit avoids this problem. See http://unix.stackexchange.com/questions/212127/zsh-disable-file-exists-warning-with-redirection

I backported this fix from #344.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -469,6 +469,11 @@ prompt_battery() {
   fi
 }
 
+# Public IP segment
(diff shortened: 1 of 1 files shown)


rjorgenson:
This looks good, one of these days i will know about things like this =]

bhilburn:
Good call, @dritter! Thanks for the great fix, as always :)

Thanks for the review, @rjorgenson. I really appreciate it!

PR #388, opened 2017-01-26
Added support for customizable context segment

felicitus:
I noticed that the context segment lacks support for customization. It uses `"%n@%m"`, which is username@machine. Machine is usually not including the `FQDN`, so when I have several machines which are named `www1` but have other `FQDNs` (like `www1.foo.example.org` and `www1.bar.example.org`) I don't have a clue if I'm on the correct machine.

This patch allows the user to change the `context` template to their desires. I included the new option in the `README` file and gave an example on how to print the full hostname.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,11 +235,19 @@ it, but only display it if you are not your normal user or on a remote host
 (basically, only print it when it's likely you need it).
 
 To use this feature, make sure the `context` segment is enabled in your prompt
-elements (it is by default), and define a `DEFAULT_USER` in your `~/.zshrc`:
+elements (it is by default), and define a `DEFAULT_USER` in your `~/.zshrc`.
+
+You can customize the `context` segment. For example, you can make it to print the
+full hostname by setting
+```
+POWERLEVEL9K_CONTEXT_TEMPLATE="%n@`hostname -f`"
+```
+
 
 | Variable | Default Value | Description |
 |----------|---------------|-------------|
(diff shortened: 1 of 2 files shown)


rjorgenson:
I wrote an enhancement for this segment a couple weeks ago and it's in next, but mine only allowed you to adjust the host depth, this is a good idea =]

bhilburn:
@felicitus - This is a great enhancement! Thanks so much for contributing it back to the project!

I need to pull this in on the `next` branch rather than `master`. I tried to do it real quickly just now, but there is unfortunately a conflict. If you have the time today, would you mind rebasing your change onto master? If you don't have time, I'll be able to do it before Monday =)

bhilburn:
@felicitus @rjorgenson - Okay, I got this merged into a branch based on `next`. There were some conflicts with not only @rjorgenson's previous change that he mentioned, but with the SSH detection that I recently merged from @shibumi.

Can both of you give the branh `merging-fel` a shot and let me know if it works as expected for you? It appears good for me.

https://github.com/bhilburn/powerlevel9k/tree/merging-fel

rjorgenson:
@bhilburn Just tested that branch and the custom context template works for me =]

bhilburn:
Fantastic! Thanks for the quick review, @rjorgenson! 

@felicitus - Thanks for the great contribution! I especially appreciate you taking the time to update the documentation along with your code change. Great work.

bhilburn:
Changes merged into `next`!

PR #391, opened 2017-01-28
Simplify `public_ip` segment

dritter:
I changed the `POWERLEVEL9K_PUBLIC_IP_METHOD` to an array. That way the code could be simplified a lot.

/cc @rjorgenson

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,7 +348,7 @@ segment will not be displayed.
 |`POWERLEVEL9K_PUBLIC_IP_FILE`|'/tmp/p8k_public_ip'|This is the file your public IP is cached in.|
 |`POWERLEVEL9K_PUBLIC_IP_HOST`|'http://ident.me'|This is the default host to get your public IP.|
 |`POWERLEVEL9K_PUBLIC_IP_TIMEOUT`|300|The amount of time in seconds between refreshing your cached IP.|
-|`POWERLEVEL9K_PUBLIC_IP_METHOD`|None|You can set this to any of 'dig', 'curl', or 'wget' to only use that method to refresh your IP.|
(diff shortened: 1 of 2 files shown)


rjorgenson:
This looks good, we'll need to update the documentation to reflect the new variable name.

dritter:
Yes, I'll update the documentation. 
And you had the methods checked in different order. I will update this as well (dig, curl, wget).

Making `POWERLEVEL9K_PUBLIC_IP_METHODS` an array looks to me like people could add their own method (which is not the case right now). Maybe we should/could add a special key `custom`, which, if specified, evaluates a special variable. But that is just an idea.

Btw. when using `dig` I get an ipv4 address, but with the other two methods I get an ipv6 address. What is odd to me is that I have ipv6 only, AFAIK..

rjorgenson:
@dritter Hmm, that is strange. I don't think the myip.opendns.com method supports ipv6, at least I couldn't find anything on the internet about it supporting it. Have you looked into the IP it returns? Maybe your ISP has it's own ipv4 gateway when going to a host that doesn't support ipv6? You can try running the dig command manually with the -6 flag, that should force ipv6 if it's available. If that works perhaps we could add a variable to force ipv6 that users could set.

If you want to force ipv6 you can use v6.ident.me as the host(as well as a slew of others out there) for curl and wget according to their documentation, I was unable to test it as my ISP doesn't provide ipv6 addresses.

dritter:
@rjorgenson Your assumption was right. myip.opendns.com does not support ipv6, but the returned ipv4 address is mine. So my ISP does indeed run a ipv4 gateway.
The curl and wget methods return the ipv6 address anyway. Personally I like ipv4 addresses more. With them I have at least a chance of memorizing them.. :D

bhilburn:
Excellent work, @dritter!

Thanks for the technical discussion, @rjorgenson, and review. It is much appreciated :)

PR #393, opened 2017-01-29
Fix theme for use with plain "prompt" command

dritter:
ZSH has the ability to preview and set themes directly from the `prompt`
builtin. To do so, the theme must be in FPATH and named
"prompt_theme_setup".

This PR adds a symlink pointing `powerlevel9k.zsh-theme` to `prompt_powerlevel9k_setup`. For full compatibility with `prompt` builtin, we still need to add our functions to `FPATH` though.

### DIFF
--- a/prompt_powerlevel9k_setup
+++ b/prompt_powerlevel9k_setup
@@ -1,0 +1,1 @@
+powerlevel9k.zsh-theme


bhilburn:
:+1: 

I'm glad you keep an eye on these sorts of things, @dritter. I don't think we would be able to support anything close to the usage methods we do without your effort.

PR #394, opened 2017-01-30
Add ability to omit the first character in dir path

dritter:
As @Hebgbs suggested in the chat, here is the option to omit the first character in dir segment.

Before (with `POWERLEVEL9K_DIR_PATH_SEPARATOR` set):
bildschirmfoto 2017-01-30 um 21 13 02

After:
bildschirmfoto 2017-01-30 um 20 52 48

@Hebgbs Could you confirm that this is what you wanted?

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,9 +304,12 @@ the path shown would be `my-cool-project`.  If you navigate to `$HOME/projects/m
 
 If you want to customize the directory separator, you could set:
 ```zsh
-# You'll need patched awesome-terminal fonts for that example
-POWERLEVEL9K_DIR_PATH_SEPARATOR="%f "$'\uE0B1'" %F"
+# Double quotes are important here!
+POWERLEVEL9K_DIR_PATH_SEPARATOR="%F{red} $(print_icon 'LEFT_SUBSEGMENT_SEPARATOR') %F{black}"
 ```
+To omit the first character (usually a slash that gets replaced if you set `POWERLEVEL9K_DIR_PATH_SEPARATOR`),
+you could set `POWERLEVEL9K_DIR_OMIT_FIRST_CHARACTER=true`.
+
 
 ##### disk_usage
 
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -611,7 +611,7 @@ prompt_dir() {
         fi
       ;;
       *)
-        current_path="%$((POWERLEVEL9K_SHORTEN_DIR_LENGTH+1))(c:$POWERLEVEL9K_SHORTEN_DELIMITER/:)%${POWERLEVEL9K_SHORTEN_DIR_LENGTH}c"
+        current_path="$(print -P "%$((POWERLEVEL9K_SHORTEN_DIR_LENGTH+1))(c:$POWERLEVEL9K_SHORTEN_DELIMITER/:)%${POWERLEVEL9K_SHORTEN_DIR_LENGTH}c")"
       ;;
     esac
   fi
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -570,7 +570,7 @@ prompt_custom() {
 # Dir: current working directory
 set_default POWERLEVEL9K_DIR_PATH_SEPARATOR "/"
 prompt_dir() {
-  local current_path='%~'
+  local current_path="$(print -P "%~")"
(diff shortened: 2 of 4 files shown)


Hebgbs:
This is the behaviour I wanted, though I wouldn't want to have this strictly imposed as _the only way_ if somebody doesn't want it; Is it toggleable for users who don't want this behaviour? If so, what's the proposed Boolean argument for it?

Cool if it's not, just looking out for people who would dislike this.

dritter:
Well, if somebody does not want this feature, they could just not set this variable. Or - as we check for `true` - just set it to any other value..

V1rgul:
This will produce ambiguous output with zsh aliased dir :
with`dev=~/dev`, 
- `~/dev`-> `< open_folder > dev` 
- ` /dev `  -> `<closed_folder> dev`

I think we should default the HOME_SUB_ICON to the same icon as HOME_ICON because it was already misleading without this

V1rgul:
There's the same problem as stated in https://github.com/bhilburn/powerlevel9k/pull/346#issuecomment-264660826 fixed by 202ab1500727e4f37545eaff849d4f9a52b823c4
I suggest we define `current_path` with `local current_path=$(print -P '%~')` (immediately evaluate) and revert 202ab1500727e4f37545eaff849d4f9a52b823c4

dritter:
@V1rgul Yes, I see the ambiguity as well. The icon is with this option set a crucial part of the path. So two folders with the same name, one in the home folder, and one in the root folder would only be distinguishable by the icon. To make it even worse, the icons for a subfolder of the home dir (open-folder) and for a subfolder of root (closed folder) are very similar.
Changing the default behaviour could be an option. I do not have a strong opinion on that. What do others think?

Regarding your second comment: I don't think I understand you right. Changing that line to what you suggest would it make impossible to use tuncation together with changing the dir path separator, no? Is anything broken for you?

Hebgbs:
@V1rgul about the ambiguity, I can see where you are coming from. It's also _super-ambiguous_ when you're in root regardless which is a _very bad thing_ to say the least.

I propose a means to figure out if a user is in any root path that doesn't begin with M produce a briefcase icon, similar to old Apple's "System suitcase" feature. While not exactly a 1:1 representation, it would at least prevent ambiguity by making `/dev`, `~/dev`, `/mnt/.../dev` and `/media/$USER/.../dev` represented as such;
💼>dev
🏠>dev
↻ >mnt/.../dev (assuming /mnt is for loopback devices / files as devices)
➝>media/$USER/dev

Noticed how I dropped the folder completely? 📂 should be for an `else` condition when the shell does not know how to represent the path. Does that mean that there may end up being ambiguity when a root path can be accessed in a different fashion? _Maybe._ Is this a lot of unnecessary work to make conditional when you're in a root path? Oh yeah. Tat's why I was fairly hesitant about suggesting iconifying the first dir segment based on location on system, because I was afraid this was going to get real stupid, real quick.

Is it worth the hassle then?

V1rgul:
@dritter Yes, the pull request doesn't work on my machine, I have a problem similar to https://github.com/bhilburn/powerlevel9k/pull/346#issuecomment-264660826 but for this segment
`%~` with the 1st character omitted just leads to `~` which isn't evaluated latter

in this case evaluating `%~` could be done immediately after the truncation, because it is needed for `POWERLEVEL9K_DIR_PATH_SEPARATOR` and `POWERLEVEL9K_DIR_OMIT_FIRST_CHARACTER`

@Hebgbs that could be great

My ideal scenario for the icon/first character (using `POWERLEVEL9K_DIR_PATH_SEPARATOR='>'`) would be
```
~     -> 🏠
~/... -> 🏠>...
~...  -> 🏠 ~... under home, 💼 ~... else
/     -> 💼
/...  -> 💼>...
```

dritter:
@V1rgul Hah! You are right! I'll fix this in the next couple of days. Thank you very much!

bhilburn:
@V1rgul - Would you mind reviewing this change and making sure it has your use-case covered?

V1rgul:
@bhilburn It is working now.
Concerning my use case, named directories are still shown as home subfolder, because they start with `~`.

dritter:
@V1rgul Yes, I didn't change the `HOME_SUBFOLDER_ICON`.. For me that is a matter of taste and not very important. You could file a PR for that anyway. 😉

bhilburn:
Just resolved a minor conflict that resulted from merging the new path shorten strategy, and all looks good. Nice coding, @dritter, and thanks for the testing & feedback, @V1rgul!

PR #395, opened 2017-01-31
Benoits dir writable segment

dritter:
This is the first (and easier) segment of @BenoitAverty .

I just renamed it from `dir_permissions` to `dir_writable` and added some documentation.


BenoitAverty:
Very thoughtful of you to keep my commit and not just the patch 👍 
Thanks for taking over, I would have kept you all waiting.

bhilburn:
This is awesome! I like this segment a lot... it may be worth considering adding this to the default segment configuration, actually.

Great work spotting this enhancement and porting over the functionality, @dritter, and great idea, @BenoitAverty! :+1:

PR #398, opened 2017-02-05
Add "truncate_with_folder_marker" shorten strategy

dritter:
This adds @BenoitAverty `truncate_with_folder_marker` strategy for the `dir` segment.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -615,7 +615,7 @@ prompt_dir() {
         dir_truncate_root=$(upsearch $POWERLEVEL9K_ROOT_MARKER_FILE)
         zero='%([BSUbfksu]|([FB]|){*})'
 
-        if [[ "$dir_truncate_root" == "/" || "$dir_truncate_root" == "$HOME" ]]; then
+        if [[ "$dir_truncate_root" == "/" || "$dir_truncate_root" == "$HOME" || "${dir_truncate_root%/*}" == "$HOME" || ${dir_truncate_root%/*} == "/" ]]; then
           current_path='%~'
         elif [[ "$dir_truncate_root" == "$HOME"* ]]; then
           current_path="~/$POWERLEVEL9K_SHORTEN_DELIMITER/${dir_truncate_root##*/}${PWD:${#${(S%%)dir_truncate_root//$~zero/}}}"
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -571,8 +571,7 @@ prompt_custom() {
 set_default POWERLEVEL9K_DIR_PATH_SEPARATOR "/"
 prompt_dir() {
   local current_path='%~'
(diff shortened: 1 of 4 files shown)


powerlevel9k.zsh-theme
```diff
@@ -610,6 +609,28 @@ prompt_dir() {
...
+            current_path="${current_path%/}/${marked_folder##*/}"
+          else
+            current_path="${current_path%/}/$POWERLEVEL9K_SHORTEN_DELIMITER/${marked_folder##*/}"
+          fi
+          last_marked_folder=$marked_folder
+        done
+
+        zero='%([BSUbfksu]|([FB]|){*})'
+        current_path=$current_path${PWD:${#${(S%%)last_marked_folder//$~zero/}}}
+      ;;
```
BenoitAverty: I didn't really know what I was doing with this function. The "loop" technique works but I'm not sure what effet it has on performance. Maybe someone with a stronger knowledge of bash or zsh can optimize it later...

bhilburn:
This is really well done, @dritter. I especially like your approach to creating user documentation, which I think makes things easier to use.

@BenoitAverty - This is a great idea, and I appreciate you providing input & review on this PR. Thanks for coming up with so many great improvements for the project!

powerlevel9k.zsh-theme
bhilburn: @dritter @BenoitAverty - What are your thoughts on Benoit's comment, here? Do you want to iterate on this function or should we merge as-is?

dritter: @bhilburn Thanks for pointing me to this again. This is a quite complicated pattern to count the visible characters in the prompt. In our case we count the characters in `$last_marked_folder` (which should not contain any non-visible characters) and replace the amount of characters in `$PWD` from the left with nothing. IMHO this could be done way easier (just replace the characters from the left), but I think @BenoitAverty just took this pattern from the `truncate_with_package_name` strategy. As far as I can see this pattern originates from [barts prompt](https://opensource.apple.com/source/zsh/zsh-55/zsh/Functions/Prompts/prompt_bart_setup). For more information about this pattern, see [this answer](http://stackoverflow.com/a/40855342/5586433) on StackOverflow.

Regarding the performance of `upsearch` function. For me it looks quite solid. Nothing to optimize here.

dritter:
I just got rid of the ugly pattern. Now the current path is truncated by removing the matched path from `upsearch` from `$PWD`. This is much more readable in my opinion.

@BenoitAverty could you confirm that everything still works as expected? I mean, I have tested it as well, the tests are green, but the more eyes, the better. 😉  Thanks in advance.

dritter:
Btw. There is still this ugly pattern in `truncate_with_package_name`, I will fix this in a different PR.

BenoitAverty:
I did a quick test just now, it seems to work OK. I may have forgotten some edge cases, but it works with one and two markers in the hierarchy so it should work in all cases.

dritter:
Good to hear! Thanks for testing @BenoitAverty

bhilburn:
Awesome :)

PR #411, opened 2017-02-23
Fix parsing package name from complex package.json files

### LINKED ISSUE #350: truncate_with_package_name sometimes grabs the wrong name
If the repo "name" isn't the first name in the package.json file, it'll print out the wrong name. 

Example:

https://github.com/gajus/babel-plugin-react-css-modules/blob/master/package.json

spits out "Gajus Kuizinas" as the project name instead of "babel-plugin-react-css-modules"

dritter:
when using `truncate_with_package_name` strategy.

This fixes #350

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -655,12 +655,15 @@ prompt_dir() {
         subdirectory_path=$(truncatePathFromRight "${current_dir:${#${(S%%)package_path//$~zero/}}}")
         # Parse the 'name' from the package.json; if there are any problems, just
         # print the file path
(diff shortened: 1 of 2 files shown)


bhilburn:
@V1rgul @fracmak - Does this branch fix the issues discussed in #350 for you?

fracmak:
it does, don't we need to update the README to point out installing jq will dramatically improve the parsing speed when truncate_with_package_name is specified?

bhilburn:
Excellent point, @fracmak! I have merged this PR and made the change to the README. 

Thanks for testing, @fracmak, and nice work, @dritter.

PR #412, opened 2017-02-24
Truncate via name in composer json

### LINKED ISSUE #349: Add support for composer.json when using truncate_with_package_name
It would be nice to be able to truncate directory names with the name of a composer/PHP project. The strategy should use the name present in either package.json or composer.json depending on which one is present. When both are present, allow the user to choose which one has precedence.

Any suggestions?

dritter:
This PR is based on #411 and should be merged after it.

Now when `truncate_with_package_name` is set, we search for a `package.json` and - if not found `composer.json` and parse the "name" field out of it. The order and the search files itself could be specified by setting `POWERLEVEL9K_DIR_PACKAGE_FILES` (defaults to `(package.json composer.json)`).

Fixes #349

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ Added an option to configure the path separator. If you want something
 else than an ordinary slash, you could set
 `POWERLEVEL9K_DIR_PATH_SEPARATOR` to whatever you want.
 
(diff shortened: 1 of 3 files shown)


bhilburn:
Excellent work, as usual, @dritter!

PR #413, opened 2017-02-24
Hide error when no origin in git repo

davidmpaz:
Hi,

Sorry for coming with this so late in time. But  I missed some details when calling `git ls-remote`, this fix git ls-remote call; match [suggested code](https://github.com/bhilburn/powerlevel9k/pull/360#discussion_r93804449) for redirect errors to null when no origin is in git repo.

Related to [PR 360](https://github.com/bhilburn/powerlevel9k/pull/360)

thanks!


dritter:
Good catch @davidmpaz !
Thanks for the contribution.

bhilburn:
Agreed! Nice work, @davidmpaz, and thanks so much for contributing this! It is much appreciated.

PR #419, opened 2017-03-02
update

angkyria:


rjorgenson:
@angkyria just FYI, you can overwrite the icons and colors used by p9k in your .zshrc so you don't need to keep your own copy of the codebase just for that =] There's more information on this in the README, and [here](https://github.com/bhilburn/powerlevel9k/wiki/Stylizing-Your-Prompt).

PR #422, opened 2017-03-04
Improves ram prompt #415

Darkheir:
For now here are the changes:

* For OS X display the sum of Pages free and Pages inactive. See https://support.apple.com/en-us/HT201538
* For Non BSD use MemAvailable instead of MemFree. See http://www.linuxatemyram.com/

For now no satisfying solution has been found for BSD systems

Issue #415, opened 2017-02-24
Use memory available instead of free


Darkheir:
Currently the prompt_ram function is getting the free ram. Wouldn't it be better to compute the value using the available value ?

In many unix system the free memory will be close to zero but in fact a lot of ram is still usable by other applications.

See: http://www.linuxatemyram.com/ (The part "How do I see how much free ram I really have?")

dritter:
Hey @Darkheir !

Thanks for that suggestion. Indeed, that would improve the segment greatly! Could you do a PR for that? This would be awesome.

Darkheir:
Okay I will create a pull request. I just have to check how to get this value in :

* Mac OSX (I have one)
* BSD (Don't have one)
* Others Linux distribution (I have one)

dritter:
👍 
Btw. our development branch is `next` (just wanted to let you know, before you need to rebase ;) ).

alilotfi:
@Darkheir 
This is a piece of code I wrote for my (Arch) linux.
```bash
prompt_ram() {
  local ramfree=0
  local ramtotal=0
  ramfree=$(grep -o -E "MemAvailable:\s+[0-9]+" /proc/meminfo | grep -o "[0-9]*")
  ramtotal=$(grep -o -E "MemTotal:\s+[0-9]+" /proc/meminfo | grep -o "[0-9]*")
  local rampercent=$(( ramfree * 100 / ramtotal ))

  "$1_prompt_segment" "$0" "$2" "yellow" "$DEFAULT_COLOR" "$rampercent%%" 'RAM_ICON'
}
```

Darkheir:
Thanks, it will help me.

Right now the thing that I can't do is the BSD part:

```bash
 if [[ "$OS" == "OSX" ]]; then
    ramfree=$(vm_stat | grep "Pages free" | grep -o -E '[0-9]+')
    # Convert pages into Bytes
    ramfree=$(( ramfree * 4096 ))
  else
    if [[ "$OS" == "BSD" ]]; then
      ramfree=$(vmstat | grep -E '([0-9]+\w+)+' | awk '{print $5}')
      base='M'
    else
      ramfree=$(grep -o -E "MemFree:\s+[0-9]+" /proc/meminfo | grep -o "[0-9]*")
      base='K'
    fi
fi
```
I will check the documentation for `vmstat` this week-end

rjorgenson:
FWIW here is the output of vmstat on a FreeBSD box I have access to.
```
[...]

PR #424, opened 2017-03-07
Add FreeBSD VM

dritter:
This PR adds a BSD box, so we can easily test P9K there (like #415 and #422). Works the same as the ubuntu VM.

## TODO
I want to move the other ZSH-Framework-Providers into the share `test-vm-providers`, so that we can install that frameworks in the BSD machine as well.

## Prerequisites
You need to install [vagrant](https://www.vagrantup.com/) and [VirtualBox](https://www.virtualbox.org/).

## Working with the VM
  1. Go into the according directory (either `test-vm` or `test-bsd-vm`)
  2. Type `vagrant up` in your terminal (may take a while at first startup)
  3. SSH into the machine via `vagrant ssh`
  4. When you are done, log out of the machine and type `vagrant halt`

### DIFF
--- a/TESTS.md
+++ b/TESTS.md
@@ -12,3 +12,11 @@ Tests in separate files under `test/functions`.
 
 These Tests tend to be more complex in setup than the basic tests. To avoid ending
 up in a huge single file, there is one file per segment in `test/segments`.
+
+# Test-VMs
+
+If unit tests are not sufficient (e.g. you have an issue with your prompt that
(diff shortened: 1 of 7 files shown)


bhilburn:
Wunderbar!

dritter:
It was surprisingly hard to get BSD running.

And we should do a BSD quality month to get more segments running there (e.g. battery & ip segments are not). :)

PR #426, opened 2017-03-09
Fix changelog formatting

dritter:
Just a little formatting fix for the changelog, so that it reads beautiful on [github](https://github.com/dritter/powerlevel9k/blob/fix_changelog/CHANGELOG.md) ([master for comparison](https://github.com/bhilburn/powerlevel9k/blob/master/CHANGELOG.md)).

Btw. this belongs to master, but is not very important. That's why I based it against `next`.


bhilburn:
Hah, yes, thanks for catching that, @dritter :)

PR #433, opened 2017-03-12
Add font debugger

dritter:
This adds a quite basic font "debugger". Well. It is a script that checks the installed font and outputs a warning message if the font does not contain "nerd", "awesome" or "powerline" in its name..

Btw. I am planning to add more debugging scripts. ;)


bhilburn:
What an awesome idea, @dritter!

This is really going to be helpful =)

PR #446, opened 2017-03-18
Protect special characters in various modes

### LINKED ISSUE #442: nerd-fonts not working on macOS
https://github.com/ryanoasis/nerd-fonts/issues/143

dritter:
I added a default `LC_CTYPE` setting to protect special characters. Interestingly we had such settings already in `functions/icons.zsh`, but for some modes this setting got lost.

This should fix #442 (already fixed by workaround).

In addition I added some tests to check if Icons were added in all modes (see #448 ). The tests fail because of that. Once the missing icons got added, the tests should work. ;)
bildschirmfoto 2017-03-20 um 00 31 26

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,9 +21,9 @@ script:
   - test/powerlevel9k.spec
   - test/functions/utilities.spec
   - test/functions/colors.spec
+  - test/functions/icons.spec
   - test/segments/command_execution_time.spec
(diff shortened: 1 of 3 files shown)


bhilburn:
Really great idea - the fact that I merged #448 without catching those is a pretty clear indicator of how important this is. Thanks, @dritter!

PR #449, opened 2017-03-19
Sort the output of `get_icon_names` alphabetically

dritter:
Now the icons are sorted alphabetically, and the users icon overrides are taken into account (by default). If `get_icon_names` is called with "original" as parameter, it prints the original icons.
bildschirmfoto 2017-03-19 um 17 54 41


bhilburn:
Great idea, @dritter. Should be very useful as a debug tool.

PR #461, opened 2017-03-31
Make home shorthand configurable for dir prompt

julienfalque:
Replaces `~` at the beginning of the directory for `dir` prompt, just like `/`. Not sure about the variable name, please tell me if you have a better idea :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,12 @@ POWERLEVEL9K_DIR_PATH_SEPARATOR="%F{red} $(print_icon 'LEFT_SUBSEGMENT_SEPARATOR
 To omit the first character (usually a slash that gets replaced if you set `POWERLEVEL9K_DIR_PATH_SEPARATOR`),
 you could set `POWERLEVEL9K_DIR_OMIT_FIRST_CHARACTER=true`.
 
+You can also customize the leading tilde character when you are in `$HOME` using:
+```zsh
+# Double quotes are important here!
+POWERLEVEL9K_HOME_FOLDER_ABBREVIATION="%F{red} $(print_icon 'HOME_ICON') %F{black}"
+```
+
 
 ##### disk_usage
 
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -613,6 +613,7 @@ prompt_command_execution_time() {
 
 # Dir: current working directory
 set_default POWERLEVEL9K_DIR_PATH_SEPARATOR "/"
+set_default POWERLEVEL9K_HOME_FOLDER_ABBREVIATION "~"
 prompt_dir() {
   local current_path="$(print -P "%~")"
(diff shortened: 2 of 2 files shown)


dritter:
Ha! Awesome! Thanks for your contribution.
I would have named it `POWERLEVEL9K_SHORTEN_HOME_DELIMITER`.. What do you think?

julienfalque:
I don't think it can be considered as a delimiter. What about `POWERLEVEL9K_DIR_LEADING_TILDE`?

dritter:
Yes, naming it "delimiter" was my initial reflex, because I wanted it to make similar to the other replacement variable (which is `POWERLEVEL9K_SHORTEN_DELIMITER`). But you are probably right. Neither my suggestion nor the delimiter name is very good.

So. How about `POWERLEVEL9K_HOME_FOLDER_ABBREVIATION`?

We should rename the old variable at some point to `POWERLEVEL9K_FOLDER_ABBREVIATION`.. 🤔

julienfalque:
`POWERLEVEL9K_HOME_FOLDER_ABBREVIATION` seems good to me, I updated it.

onaforeignshore:
Devil's advocate here - what if someone does `POWERLEVEL9K_HOME_FOLDER_ABBREVIATION="Home Folder"`. Then it's hardly an abbreviation. Wouldn't `POWERLEVEL9K_HOME_FOLDER_NAME` be better then?

dritter:
@onaforeignshore For that name you have to think outside of the box. I see you are coming from replacing `~` with something longer is not an abbreviation any more, but that train of thought is because you have seen the code / know that *nix OSes abbreviate the users home folder with a tilde.
For me that the OS abbreviates `/home/user` with `~` *is* the default value of `POWERLEVEL9K_HOME_FOLDER_ABBREVIATION`. We don't set it with our `set_default` function, because it is already done. If this was our code that name would make total sense. And from the outside perspective it looks like we are doing this..

`POWERLEVEL9K_HOME_FOLDER` is not specific enough in my opinion..

bhilburn:
I think this is a neat idea! Thanks for contributing, @julienfalque!

One item before I merge it: can you update the README to include this new variable?

Regarding the name of the variable:
@dritter's suggestion does a good job of capturing how people will think about it, but @onaforeignshore raises an interesting point about what it actually represents. Interestingly, using `~` to represent `$HOME` is actually a BASH thing, and can be different from shell-to-shell. ZSH gets it from BASH. The technical name for it is the "BASH Tilde Expansion", in that it expands to `$HOME`. Here, we are going the opposite direction. I can't think of a name better than `ABBREVIATION`, though.

onaforeignshore:
@bhilburn if it is "BASH Tilde Expansion", the opposite would be `CONTRACTION` not `ABBREVIATION`, from a purely linguistic POV

julienfalque:
Updated the README file. About the variable name, I'm out of ideas too :)

bhilburn:
Thanks so much for the great contribution, @julienfalque! Nice work, and welcome to the project :)

PR #463, opened 2017-04-01
Gracefully handle `echotc Co` failure

liblit:
In Emacs, `M-x shell` creates a shell buffer with *very* rudimentary terminal capabilities. `$TERM` is set to `dumb`, and `echotc Co` fails after printing `echotc: no such capability: Co` to standard error. If our goal was to determine whether the terminal supports 256 colors, we really have no information to go on. I recommend that we trust the user and hope for the best. That means we should (1) discard any `echotc Co` error output, and (2) forego the warning message about having fewer than 256 colors if `echotc Co` failed.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1317,8 +1317,8 @@ prompt_powerlevel9k_setup() {
 
   # Display a warning if the terminal does not support 256 colors
   local term_colors
-  term_colors=$(echotc Co)
-  if (( $term_colors < 256 )); then
+  term_colors=$(echotc Co 2>/dev/null)
+  if (( ! $? && ${term_colors:-0} < 256 )); then
     print -P "%F{red}WARNING!%f Your terminal appears to support less than 256 colors!"
     print -P "If your terminal supports 256 colors, please export the appropriate environment variable"
     print -P "_before_ loading this theme in your \~\/.zshrc. In most terminal emulators, putting"


dritter:
Yep, the only intention was to print a nice error if the terminal supports less than 256 colors.. Our hope was that `echotc` always returns the correct number of colors.. Obviously it doesn't. Great catch. 👍

powerlevel9k.zsh-theme
```diff
@@ -1317,8 +1317,8 @@ prompt_powerlevel9k_setup() {
 
   # Display a warning if the terminal does not support 256 colors
   local term_colors
-  term_colors=$(echotc Co)
-  if (( $term_colors < 256 )); then
+  term_colors=$(echotc Co 2>/dev/null)
+  if (( ! $? && ${term_colors:-0} < 256 )); then
```
dritter: I think the test if the previous command (`echotc Co`) is unnecessary. A `print ${XXX_UNKNOWN:-0}` always returns zero, which is - surprisingly - always smaller than 256..

liblit: Are you suggesting that the `if` condition be simplified to

```zsh
if (( ${term_colors:-0} < 256 )); then
```

That will still print the warning if `echotc Co` failed. However, we could use this instead:

```zsh
if (( ${term_colors:-256} < 256 )); then
```

That's more concise, though perhaps also more obscure. I can certainly revise my pull request to use that approach if you prefer it. Just let me know.

dritter: Ha. You are right. 👍  I was on the wrong track.. ;)

liblit: It certainly took me a few rounds to get the `$?` test right. If that confused me *and* you, it will probably confuse other people too. Perhaps the most readable (maintainable) approach really is to use `${term_colors:-256}` and get rid of the `$?` check entirely.

dritter: I am not sure if that will bring the desired clarity. I'd like to hear some other people on this (ping @bhilburn ).
And probably a comment will do. ;)

bhilburn: I actually like what's there, now. Great work, @liblit, and thanks for the review, @dritter!

bhilburn:
Merged into `master` and then `master` --> `next`

powerlevel9k.zsh-theme
liblit: You're very welcome. Thank *you* for making my shell more awesome!

bhilburn:
Thanks for the great contribution, @liblit! Even small contributions like these have a really important impact for the project, and it's how we can keep something like P9k well-maintained as our codebase grows. Thanks!

PR #471, opened 2017-04-06
Engines

onaforeignshore:
This separates out the prompt generation logic from the main file so that you can switch between the two easily.


onaforeignshore:
Replaced by #474

PR #472, opened 2017-04-06
Merging Awesome Font Fix

bhilburn:
This is the PR effort that merges @V1rgul's #385, which we need as soon as possible.

It is separate from #385 because I added a new icon set for `AdobeSourceCodePro`, for which we were previously telling users to use `awesome-fontconfig` since the codepoints lined up reasonably well. That was a mistake on my part, and now we are going to break this for those users.

More on this in the review that I'm about to leave. I would like to resolve my review comments before completing this merge.

@V1rgul - Your testing of this branch will be critical. It would also be great to get @protist testing this.

### DIFF
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -13,6 +13,7 @@
 
 # Initialize the icon list according to the user's `POWERLEVEL9K_MODE`.
 typeset -gAH icons
+
 case $POWERLEVEL9K_MODE in
   'flat'|'awesome-patched')
     # Awesome-Patched Font required! See:
@@ -86,9 +87,8 @@ case $POWERLEVEL9K_MODE in
   'awesome-fontconfig')
     # fontconfig with awesome-font required! See
     # https://github.com/gabrielelana/awesome-terminal-fonts
-
     # if not defined, set recommended linux path
-    typeset -p "POWERLEVEL9K_FONTAWESOME_PATH" > /dev/null 2>&1  || POWERLEVEL9K_FONTAWESOME_PATH=~/.fonts  
+    typeset -p "POWERLEVEL9K_FONTAWESOME_PATH" > /dev/null 2>&1  || POWERLEVEL9K_FONTAWESOME_PATH=~/.fonts
 
     source "$POWERLEVEL9K_FONTAWESOME_PATH/fontawesome-regular.sh"
     # source "$POWERLEVEL9K_FONTAWESOME_PATH/devicons-regular.sh" # no named codepoints
@@ -97,7 +97,7 @@ case $POWERLEVEL9K_MODE in
     # Set the right locale to protect special characters
     local LC_ALL="" LC_CTYPE="en_US.UTF-8"
     icons=(
-      LEFT_SEGMENT_SEPARATOR         $'\uE0B0'                                      #        
+      LEFT_SEGMENT_SEPARATOR         $'\uE0B0'                                      # 
       RIGHT_SEGMENT_SEPARATOR        $'\uE0B2'                                      # 
       LEFT_SEGMENT_END_SEPARATOR     ' '                                            # Whitespace
       LEFT_SUBSEGMENT_SEPARATOR      $'\uE0B1'                                      # 
@@ -106,7 +106,7 @@ case $POWERLEVEL9K_MODE in
       ROOT_ICON                      '\u'$CODEPOINT_OF_OCTICONS_ZAP                 # 
       RUBY_ICON                      '\u'$CODEPOINT_OF_OCTICONS_RUBY' '             # 
       AWS_ICON                       '\u'$CODEPOINT_OF_AWESOME_SERVER               # 
-      AWS_EB_ICON                    $'\U1F331 '                                    # 🌱     
+      AWS_EB_ICON                    $'\U1F331 '                                    # 🌱
       BACKGROUND_JOBS_ICON           '\u'$CODEPOINT_OF_AWESOME_COG' '               # 
       TEST_ICON                      '\u'$CODEPOINT_OF_AWESOME_BUG                  # 
       TODO_ICON                      '\u'$CODEPOINT_OF_AWESOME_CHECK_SQUARE_O       # 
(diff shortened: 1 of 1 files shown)


functions/icons.zsh
```diff
@@ -86,6 +87,76 @@ case $POWERLEVEL9K_MODE in
   'awesome-fontconfig')
     # fontconfig with awesome-font required! See
     # https://github.com/gabrielelana/awesome-terminal-fonts
+    # if not defined, set recommended linux path
+    typeset -p "POWERLEVEL9K_FONTAWESOME_PATH" > /dev/null 2>&1  || POWERLEVEL9K_FONTAWESOME_PATH=~/.fonts
```
bhilburn: We have a lot of OSX users, and I'm concerned this will break things for them. However, looking at the "Awesome Font" installation instructions for OSX, it doesn't talk about installing these scripts, anywhere.

We either need some OSX users to weigh-in, here, or even better if we could get @gabrielelana's thoughts on how to handle this for OSX, that would be even better.

functions/icons.zsh
```diff
@@ -86,6 +87,76 @@ case $POWERLEVEL9K_MODE in
...
+      RUST_ICON                      $'\uE6A8'                                      # 
+      PYTHON_ICON                    $'\U1F40D'                                     # 🐍
+      SWIFT_ICON                     $'\uE655'                                      # 
+      PUBLIC_IP_ICON                 '\u'$CODEPOINT_OF_AWESOME_GLOBE                # 
+      LOCK_ICON                      '\u'$CODEPOINT_OF_AWESOME_LOCK                 # 
+      EXECUTION_TIME_ICON            '\u'$CODEPOINT_OF_AWESOME_HOURGLASS_END        # 
+      SSH_ICON                       '(ssh)'
+    )
+  ;;
+  'AdobeSourceCodePro')
```
bhilburn: So, creating icon sets is cheap for us. Since only the one that the user actually specified actually gets loaded, we can specify as many as we want without burdening the user. I should have just created this font set ages ago instead of telling SourceCodePro users to use `awesome-fontconfig`, because now this is going to break.

Which takes me to my question: in `awesome-fontconfig`, is there a way we can have ZSH fall-through to the next option, `AdobeSourceCodePro`, if the Awesome Font scripts aren't found? This would be fairly easy if there was something like a `continue` statement, but I actually can't find any documentation of that for ZSH. Thoughts, @dritter?

functions/icons.zsh
gabrielelana: Yep, unfortunately AFAIK on OSX doesn't work like that, and it's even more difficult with Sierra. Anyway, the mapping part is good even for OSX, the environment variable above probably doesn't work but it's harmless, I don't think it will "break" anything

@bhilburn what do you think it will break?

V1rgul:
Ok, every icon is working for me :)

functions/icons.zsh
V1rgul: We could print an error if the scripts aren't found, but I don't think we should silently fallback on `AdobeSourceCodePro`

protist:
Thanks for the fix, but unfortunately I'm away from my computer on holidays right now. I'm definitely happy to help test, but it might be ~2 weeks from now.

functions/icons.zsh
bhilburn: @gabrielelana - Thanks for the response! The piece I don't understand for OSX is where is the `*.sh` script that we need to execute to get the mapping variables? I'm assuming that is a requirement on OSX, as well?

gabrielelana: Yes, you need them on OSX as well, I guess we can keep them in the same place as on Linux (aka `~/.fonts` directory), I will integrate my documentation to tell OSX users to always copy the mapping files

bhilburn: @gabrielelana - Okay, awesome. Can you let us know once the Awesome docs are updated? If I merge this before that's done, we are both likely to get hammered with support issues :)

protist:
I'm back from holidays… 🙂

This seems to work well for me! Thank you! FWIW on Arch Linux, the default package-manager–installed location for the awesome-terminal-fonts `*.sh` scripts is as follows.

    POWERLEVEL9K_FONTAWESOME_PATH='/usr/share/fonts/awesome-terminal-fonts/'

I'm not sure if that would be universal enough (even in Linux) to set as a default, though. Thanks for the PR!

functions/icons.zsh
bhilburn: @gabrielelana - Just checking in on this :) We can't proceed on our end until OSX has the script in the user directory, otherwise we'll break everything for a large number of users.

bhilburn:
Thanks for testing, @protist!

Okay, it seems like the code-points themselves are ready to go (makes sense since they are just @V1rgul's commits), but we have an issue with regard to installation path of the the font-mapping script.

If we can't find the script, and thus execute it, we will break things for a lot of users. There is the obvious fix of users simply finding the script themselves and symlinking it into a location where we are looking, but that is:

a) A painful user experience.
b) Will generate a high support load for us.
c) Opens the door to lots of accidental breakages (e.g., a user copies the script instead of symlinking it, @gabrielelana updates the code-points, P9k breaks).

It's clear that using the script that @gabrielelana packages in the correct way to do this, but we need to be sure that we aren't making things more difficult for our users. Unfortunately, I don't know how to do this without just making a giant list of directories to check (it's not like this is going to be in pkgconfig or in the env).

Any other ideas, here?

protist:
No worries @bhilburn! It's probably not useful, but it looks like the `awesome-terminal-font` [package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=awesome-terminal-fonts#n21) from Arch Linux's AUR just puts the `*.sh` scripts in the same directory as the `*.ttf` fonts themselves. However, I'm not sure if this is (a) consistent in other systems, or (b) a path that can be extracted programatically anyway.

gabrielelana:
@bhilburn sorry to be so late, I have updated the `README.md` of [awesome-terminal-fonts](https://github.com/gabrielelana/awesome-terminal-fonts), take a look and let me know if this will suffice or if you need something else

functions/icons.zsh
dritter: Btw. the docs are already updated: https://github.com/gabrielelana/awesome-terminal-fonts/commit/1397337e349c8791b64088995bfba820784aed18
Thx. @gabrielelana 👍

functions/icons.zsh
dritter: Hmm. I think the easiest way is to rename the icon sets.. I would leave the old one `awesome-fontconfig` and rename the new one to something like `awesome-fontmap` or `awesome-mapped-fontconfig`.
That way we can leave the old way like it is and recommend every user that has trouble with updated fonts to use the new one with the maps.
Btw. IMHO we should favour the new mapped way over the old one and reflect that in the documentation.

bhilburn:
This work was picked up by @pfrybar  and continued in #599

PR #473, opened 2017-04-07
Only show go_version segment if inside GOPATH

zikes:
As of Go v1.8 GOPATH need not be set, as it will default to $HOME/go. By using `go env GOPATH` to retrieve the value first, this technique will work regardless of whether GOPATH is currently set by the user.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -746,7 +746,7 @@ prompt_go_version() {
   local go_version
   local go_path
   go_version=$(go version 2>/dev/null | sed -E "s/.*(go[0-9.]*).*/\1/")
-  go_path=$(go env GOPATH)
+  go_path=$(go env GOPATH 2>/dev/null)
 
   if [[ -n "$go_version" && "${PWD##$go_path}" != "$PWD" ]]; then
     "$1_prompt_segment" "$0" "$2" "green" "255" "$go_version"
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -744,9 +744,11 @@ prompt_docker_machine() {
 # GO prompt
 prompt_go_version() {
   local go_version
+  local go_path
   go_version=$(go version 2>/dev/null | sed -E "s/.*(go[0-9.]*).*/\1/")
+  go_path=$(go env GOPATH)
 
-  if [[ -n "$go_version" ]]; then
+  if [[ -n "$go_version" && "${PWD##$go_path}" != "$PWD" ]]; then
     "$1_prompt_segment" "$0" "$2" "green" "255" "$go_version"
   fi
 }
--- a/test/segments/go_version.spec
+++ b/test/segments/go_version.spec
@@ -19,8 +19,11 @@ function testGo() {
   alias go=mockGo
   POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(go_version)
 
+  PWD="$HOME/go/src/github.com/bhilburn/powerlevel9k"
+
   assertEquals "%K{green} %F{255}go1.5.3 %k%F{green}%f " "$(build_left_prompt)"
 
+  unset PWD
   unset POWERLEVEL9K_LEFT_PROMPT_ELEMENTS
   unalias go
 }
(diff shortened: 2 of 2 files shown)


bhilburn:
Thanks for the PR, @zikes! It looks like this PR is failing CI tests with:

```
$ test/segments/go_version.spec

testGo
ASSERT:expected:<%K{green} %F{255}go1.5.3 %k%F{green}%f > but was:<%k%F{NONE}%f >
testGoSegmentPrintsNothingIfGoIsNotAvailable
prompt_go_version:4: command not found: noGo

Ran 2 tests.

FAILED (failures=1)
The command "test/segments/go_version.spec" exited with 1.
```

So, the test that is failing is supposed to not be able to find Go and then confirm that the Go segment is not, in fact, printed. It's failing because something *is in fact* printed. I'm not familiar with Go myself, but based on @zikes' description of the change, perhaps this is expected.

@dritter - Is this a case of the test needing to be updated or this change doing something unexpected?

dritter:
I would say both. 😉 

`testGoSegmentPrintsNothingIfGoIsNotAvailable` sets an alias to an unknown function (`noGo`) and assures that nothing is printed. The new query of the `GOPATH` does not redirect STDERR, which is why this fails (an error is printed).

And the assertion for `testGo` fails, because there is nothing printed, where it should..

powerlevel9k.zsh-theme
```diff
@@ -744,9 +744,11 @@ prompt_docker_machine() {
 # GO prompt
 prompt_go_version() {
   local go_version
+  local go_path
   go_version=$(go version 2>/dev/null | sed -E "s/.*(go[0-9.]*).*/\1/")
+  go_path=$(go env GOPATH)
```
dritter: This prints an error message if no `go` is installed. I know one should not activate the go segment, if he does not intend to install go, but I like segments to degrade gracefully.. So a `$(go env GOPATH 2>/dev/null)` would be nice.

powerlevel9k.zsh-theme
```diff
@@ -744,9 +744,11 @@ prompt_docker_machine() {
...
 
-  if [[ -n "$go_version" ]]; then
+  if [[ -n "$go_version" && "${PWD##$go_path}" != "$PWD" ]]; then
```
dritter: This removes the longest submatch of `GOPATH` from `PWD`, and returns true if the remainder of the truncation is not equal to `PWD`. If `GOPATH` is empty, then nothing is truncated from `PWD`, and the go version is not printed..

zikes: Anything I should change about this? I had a hard time finding a proper method for determining whether PWD was a subset of GOPATH, and this seemed to be the most acceptable method I could find on StackOverflow.

dritter: Do you want the segment to be printed if the GOPATH is empty? If the answer is yes, then this conditional needs to be changed. If no, then you should change the test to cover that.

bhilburn: Hm, so this is interesting. `GOPATH` being empty doesn't necessarily mean that golang isn't installed, but as far as P9k is concerned, it might as well not be. I think the best answer is that if `GOPATH` is empty the segment is not printed.

@zikes - Do you need help modifying the test per @dritter's comment? I do want to get this PR merged as soon as we can =)

bhilburn:
Looks like there is a minor question of handling `GOPATH`, per the comment thread in your recent changes. As soon as we know that out, I'm happy to merge :)

powerlevel9k.zsh-theme
zikes: Apologies, the tests were more complicated than I expected. There should now be tests for whether or not GOPATH is set, as well as whether or not PWD is a child directory of GOPATH.

bhilburn:
I don't want to lose track of this one, as I think @zikes has some really good improvements here. I'd love to get this merged ASAP now that v0.6.3 is out.

I'm actually having a bit of trouble following the review / changelog, here, on Github. @dritter, it looks like you requested changes, @zikes made changes, and then you have not yet re-reviewed. Is that correct?

dritter:
That is correct @bhilburn . Apologies for the late review @zikes , I lost this one out of sight.
Reviewing now.

bhilburn:
Fantastic! Thanks so much for the PR, @zikes! Welcome to P9k, and thanks for a great contribution =)

Issue #479, opened 2017-04-12
A couple of issues when using powerlevel9k on Android


guixxx:
I have two issues using powerlevel9k on Termux. The first one is when when I first open a terminal window:

![screenshot_20170412-100930](https://cloud.githubusercontent.com/assets/27077338/24959185/505bc820-1f68-11e7-963d-6a893e4e29a3.png)

`powerlevel9k_prepare_prompts:3 number truncated after 10 digits: 99999999999`

I don't know if it affects anything, but that warning message is kinda annoying, tbh. The second issue is related to the `command_execution_time` plug-in. As you can see in the screenshot above, the plug-in is reporting the startup time wrongly, but after that, any command I run the plug-in reports the execution time correctly:

![screenshot_20170412-100949](https://cloud.githubusercontent.com/assets/27077338/24959188/541f56fc-1f68-11e7-9788-1be1665a22e2.png)

My `.zshrc`:

```zsh
DEFAULT_USER="${USER}"
DISABLE_AUTO_UPDATE=true
POWERLEVEL9K_INSTALLATION_PATH="${HOME}/.antigen/bundles/bhilburn/powerlevel9k"
POWERLEVEL9K_COMMAND_EXECUTION_TIME_THRESHOLD=0
POWERLEVEL9K_COMMAND_EXECUTION_TIME_BACKGROUND=245
POWERLEVEL9K_COMMAND_EXECUTION_TIME_FOREGROUND='black'
POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(status context dir vcs)
POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(root_indicator background_jobs command_execution_time time)
POWERLEVEL9K_STATUS_VERBOSE=false
POWERLEVEL9K_MODE='nerdfont-fontconfig'
source $HOME/.antigen-git/antigen.zsh
antigen use oh-my-zsh
antigen bundle zsh-users/zsh-syntax-highlighting
antigen theme bhilburn/powerlevel9k powerlevel9k
antigen apply
```

I'm running Android 6.0.1 and using Termux as my emulator. Additional info:

```
$ zsh --version
zsh 5.3 (arm-unknown-linux-androideabi)
$ uname -a
Linux localhost 3.4.42-ge642ba1 #1 SMP PREEMPT Tue Dec 20 13:28:16 PST 2016 armv7l Android
```

Thank you.

Edit: fixed typos in .zshrc

[...]

PR #480, opened 2017-04-12
Adds Android icon/battery plugin support

guixxx:
Plus fixes bhilburn/powerlevel9k#479, which was causing integer overflow on 32-bit CPUs.

screenshot_20170412-162318

### DIFF
--- a/functions/utilities.zsh
+++ b/functions/utilities.zsh
@@ -100,6 +100,14 @@ case $(uname) in
     Linux)
       OS='Linux'
       OS_ICON=$(print_icon 'LINUX_ICON')
+
+      # Check if we're running on Android
+      case $(uname -o 2>/dev/null) in
+        Android)
+          OS='Android'
+          OS_ICON=$(print_icon 'ANDROID_ICON')
+          ;;
+      esac
       ;;
     SunOS)
       OS='Solaris'
@@ -108,15 +116,6 @@ case $(uname) in
     *)
       OS=''
       OS_ICON=''
-      ;;
-esac
-
-# Not all OSes support the '-o' parameter
-# That's why this second condition is needed 
-case $(uname -o 2>/dev/null) in
-    Android)
-      OS='Android'
-      OS_ICON=$(print_icon 'ANDROID_ICON')
       ;;
 esac
 
--- a/functions/utilities.zsh
+++ b/functions/utilities.zsh
@@ -111,6 +111,15 @@ case $(uname) in
       ;;
 esac
 
+# Not all OSes support the '-o' parameter
+# That's why this second condition is needed 
+case $(uname -o 2>/dev/null) in
+    Android)
+      OS='Android'
+      OS_ICON=$(print_icon 'ANDROID_ICON')
+      ;;
+esac
+
 # Determine the correct sed parameter.
 #
 # `sed` is unfortunately not consistent across OSes when it comes to flags.
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -425,11 +425,16 @@ prompt_battery() {
     esac
   fi
 
-  if [[ $OS =~ Linux ]]; then
+  if [[ "$OS" == 'Linux' ]] || [[ "$OS" == 'Android' ]]; then
     local sysp="/sys/class/power_supply"
+
     # Reported BAT0 or BAT1 depending on kernel version
     [[ -a $sysp/BAT0 ]] && local bat=$sysp/BAT0
     [[ -a $sysp/BAT1 ]] && local bat=$sysp/BAT1
+
+    # Android-related
+    # Tested on: Moto G falcon (CM 13.0)
+    [[ -a $sysp/battery ]] && local bat=$sysp/battery
 
     # Return if no battery found
     [[ -z $bat ]] && return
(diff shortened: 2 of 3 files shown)


dritter:
😆  Awesome! That was fast. Thank you very much for your contribution!

powerlevel9k.zsh-theme
```diff
@@ -1278,8 +1283,9 @@ powerlevel9k_prepare_prompts() {
   RETVAL=$?
 
   _P9K_COMMAND_DURATION=$((EPOCHREALTIME - _P9K_TIMER_START))
+
   # Reset start time
-  _P9K_TIMER_START=99999999999
+  _P9K_TIMER_START=2147483647
```
dritter: Could you add a comment, that this is the maximum integer on 32bit?

powerlevel9k.zsh-theme
```diff
@@ -1312,8 +1318,11 @@ NEWLINE='
 }
 
 prompt_powerlevel9k_setup() {
+  # I decided to use the value below for better supporting 32-bit CPUs, since the previous value "99999999999" was causing issues on my Android phone, which is powered by an armv7l
+  # We don't have to change that until 19 January of 2038! :)
+
   # Disable false display of command execution time
-  _P9K_TIMER_START=99999999999
+  _P9K_TIMER_START=2147483647
```
dritter: Could you add a comment, that this is the maximum integer on 32bit? <- Here as well

dritter:
Awesome! One thing: Our development branch is `next`. Could you rebase your PR against that branch? That would be nice. 👍

guixxx:
How can I do that? 😐 I'm completely new to the git universe, I created my account a few days ago. I could search on how to do that, but I don't want to risk messing things up. Thanks.

dritter:
@guixxx Don't worry, I guess @bhilburn can change the base of this PR. So you won't have a hassle with it. 😉

functions/utilities.zsh
```diff
@@ -111,6 +111,15 @@ case $(uname) in
       ;;
 esac
 
+# Not all OSes support the '-o' parameter
+# That's why this second condition is needed 
+case $(uname -o 2>/dev/null) in
```
bhilburn: Hm, I see why you did this, but at the same time, breaking this out of the main `case` statement will result in every OS evaluating this conditional, even if the OS has already been discovered.

What does `uname` return on Android? `Linux`? If so, perhaps we can add this as a second conditional within the `Linux` case in the main statement.

dritter:
I just installed `Termux` on my phone. I don't face this issues, as I have there 64bit as well 🙄 

And @bhilburn : Yes, `uname` returns `Linux` on Android. Good catch.

functions/utilities.zsh
guixxx: Thanks for your kind words! Yes, you got that right and I've fixed it. Let me know what you think now!

onaforeignshore:
I was wondering if it would be better to do something like ...

functions/utilities.zsh
```
# Architecture detection
case $(uname -m 2>/dev/null) in
    x86_64)
      ARCH='x64'
      ;;
    *)
      ARCH='x86'
      ;;
esac
```

Main file:
```
[[ "$ARCH" == "x64" ]] && _P9K_TIMER_START=99999999999 || _P9K_TIMER_START=2147483647
```

That way we can add support for 32-bit without changing 64-bit, and also use `$ARCH` globally if future segments need them.

onaforeignshore:
Don't forget that the `OS_ICON` is set in `prompt_os_icon()`, not in `utilities`...

```
prompt_os_icon() {
  local OS_ICON
  case "${OS}" in
    OSX)
      OS_ICON=$(print_icon 'APPLE_ICON')
      ;;
    BSD)
      OS_ICON=$(print_icon 'FREEBSD_ICON')
      ;;
    Linux)
      OS_ICON=$(print_icon 'LINUX_ICON')
      ;;
**    Android)
      OS_ICON=$(print_icon 'ANDROID_ICON')
      ;;**
    Solaris)
      OS_ICON=$(print_icon 'SUNOS_ICON')
      ;;
  esac
```

bhilburn:
@onaforeignshore -

> Don't forget that the OS_ICON is set in prompt_os_icon(), not in utilities...

Er, actually, it is set in `utilities`, now. See:
https://github.com/bhilburn/powerlevel9k/blob/master/powerlevel9k.zsh-theme#L886

Are you working off of an older branch, by chance?

> That way we can add support for 32-bit without changing 64-bit, and also use $ARCH globally if future segments need them. We can also add other architectures if required.

Excellent suggestion, @onaforeignshore! I actually tried to implement this, and discovered that while I'm on a 64-bit machine, my ZSH shell is only compiled to support 4-byte integers. I guess the Fedora packager for ZSH just decided that 4-bytes is fine, and rolled with it.

Since $ARCH isn't a good indicator, the only other way would be to directly test bit precision in the shell. This seemed like a bit overboard for a number that will rarely overflow 0xFFFFFFFF, though, so I left it as-is for now.

bhilburn:
This PR has been merged into `next`!

guixxx:
Hey guys! Sorry for the very late response. First of, thanks for adding this PR. :)

I've just updated my powerlevel9k to the master branch, and I got this error:

`powerlevel9k_prepare_prompts:3: number truncated after 7 digits: FFFFFFFF`

Which refers to [this line](https://github.com/bhilburn/powerlevel9k/blob/master/powerlevel9k.zsh-theme#L1298):

`_P9K_TIMER_START=0xFFFFFFFF`

That hex notation represents the maximum **unsigned** integer value in 32-bit systems. Since zsh scripts don't have the luxury of unsigned integers, we need to change that value to the maximum **signed** integer in 32-bit, which is `0x7FFFFFFF` or `2147483647`. I'd also suggest changing [this line](https://github.com/bhilburn/powerlevel9k/blob/master/powerlevel9k.zsh-theme#L1336) to the hex notation as well (it looks better :D).

Thanks guys.

PR #482, opened 2017-04-15
Attempt to fix powerlevel9k on systems where the perl fallback fails

belak:
This is a hacky workaround I've been playing with on OSX when messing with the prezto integration. readlink doesn't have -f and perl appears to return an empty string... so I fell back to python which may be relatively slow but it should be effective.

Please feel free to leave suggestions... I'm not really happy with how this looks at the moment, even if it works.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -45,6 +45,9 @@ fi
 if [[ -L "$POWERLEVEL9K_INSTALLATION_PATH" ]]; then
   # If this theme is sourced as a symlink, we need to locate the real URL
   filename="$(realpath -P $POWERLEVEL9K_INSTALLATION_PATH 2>/dev/null || readlink -f $POWERLEVEL9K_INSTALLATION_PATH 2>/dev/null || perl -MCwd=abs_path -le 'print abs_path readlink(shift);' $POWERLEVEL9K_INSTALLATION_PATH 2>/dev/null)"
+  if [[ -z "$filename" ]]; then
+    filename="$(python -c 'import os,sys;print(os.path.realpath(sys.argv[1]))' $POWERLEVEL9K_INSTALLATION_PATH)"
+  fi
 elif [[ -d "$POWERLEVEL9K_INSTALLATION_PATH" ]]; then
   # Directory
   filename="${POWERLEVEL9K_INSTALLATION_PATH}/powerlevel9k.zsh-theme"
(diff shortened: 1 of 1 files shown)


onaforeignshore:
On OS X this part should work `realpath -P $POWERLEVEL9K_INSTALLATION_PATH`. It does for me

onaforeignshore:
I installed the `coreutils` using `brew install coreutils` for this to work

belak:
However, that requires an additional package which is never mentioned in the documentation. This attempts to work around the issue if you don't have coreutils installed.

That being said, I do appreciate the workaround. It's nice to have something that works until there's some sort of fix merged in.

onaforeignshore:
If you want a pure macOS approach, try `stat -f%N $POWERLEVEL9K_INSTALLATION_PATH`

dritter:
@belak Thanks for the fix.
I agree that `coreutils` should not be a dependency for P9K. Could you try the following code:
```zsh
0=${(%):-%N}
dir=${0:A:h}
```
This piece of code comes from [this comment](https://github.com/bhilburn/powerlevel9k/pull/263#issuecomment-233740883). I always wanted to test it, but never found time for it.. It may be a good way to get rid of a lot of the old cruft. :)

bhilburn:
Thanks for the comments, @dritter and @onaforeignshore! And yes, agreed, we can't add `coreutils` as a dependency. This seems like a valuable fix though, @belak. Let us know what you find from the test @dritter suggested!

belak:
That's actually very similar to what I tried initially... however, when loaded with prompt_powerlevel9k_setup, the usage of that is inside a function (an autoloaded prompt_powerlevel9k_setup which gets put into the fpath) and $0 ends up being the pwd, not the location of the script. We actually had this issue with prezto as well and ended up storing the result of ${0:h} as ZPRESTODIR so it could be used inside the function itself.

belak:
Another option may be changing the options for readlink because that's found on a default OSX install as well.

dritter:
@belak what is a bit odd to me is that the `whence -v $0` trick does not work for you.. This is the only (more or less reliable) way of finding the source file of a defined function in ZSH (without using external tools). It should be available if you are using ZSH > 5.0.8.
See [source](https://github.com/bhilburn/powerlevel9k/blob/99e59bc83a7b20287306209cecbe732d5c21c46a/powerlevel9k.zsh-theme#L29-L38)

Hmm. Maybe it is a chicken-egg problem. We ask in [line 29](https://github.com/bhilburn/powerlevel9k/blob/99e59bc83a7b20287306209cecbe732d5c21c46a/powerlevel9k.zsh-theme#L29) if `prompt_powerlevel9k_setup` is defined as function. The actual function definition happens much later in the script. Could you change that line to `whence -w $0`?

belak:
I'm actually closing this in favor of #494 because it removes all external dependencies to read symlinks.

@dritter My issue wasn't with getting the filename. That part worked fine. POWERLEVEL9K_INSTALLATION_PATH ended up getting set to /Users/belak/.antigen/bundles/sorin-ionescu/prezto/modules/prompt/functions/prompt_powerlevel9k_setup using `whence $0` branch (this is following the set up from the prezto instructions of this repo). The issue was with resolving that symlink.

dritter:
@belak Ah. I see. Apologies for getting you wrong.

PR #483, opened 2017-04-16
Set tick to color name green instead of color code

changyuheng:
So the tick will not be too bright in light background color scheme such
as Solarized Light.

screen shot 2017-04-15 at 11 06 21 pm
screen shot 2017-04-15 at 11 06 43 pm


onaforeignshore:
You can override the color in your configuration file (usually `.zshrc`) using the following lines:
```
POWERLEVEL9K_STATUS_OK_BACKGROUND=""						# (note: only, if verbose is not false)
POWERLEVEL9K_STATUS_OK_FOREGROUND=""						# (note: only, if verbose is not false)
```

bhilburn:
@changyuheng - Great idea! And you're right, using that color code does make it a bit obnoxious if the terminal is setting different default colors.

Thanks so much for contributing to P9k, and welcome to the project =)

Issue #312, opened 2016-08-31
[svn] Change color of icons


shibumi:
Hello,
Is there any chance to change the color of the icons in the svn powerline?
Here is a screenshot: 

![yellow...](https://paste.archlinux.de/iWz/)

As you can see I can't see the icons because the icons and the latest svn number is yellow.
Any chance to change this or do I need to change my terminal color theme?

I don't have this issue with the git powerline there is everything and the icons are black on yellow.

bhilburn:
Ah, okay, I now see what you were talking about in your original PR. Not being a user of the Awesome font series myself, I didn't see this.

I'll play with it. I'm sure we can fix it.

shibumi:
@bhilburn  Yes thats what I've meant. I had no idea how to fix it myself.. I tried injecting some colorcodes into the line but it has always just broke my powerline.

shibumi:
Hey can you explain me how I can the color? I would do it my self. Would also be awesome if you could merge the newest changes with svn in the git master branch.

bhilburn:
@shibumi - I just spent a bit of time looking at this, and I honestly have no idea what's happening. I created a branch where I made some changes (just simplified something), but I honestly am not sure how this could be happening.

If anyone else has insight, here, that would be helpful.
https://github.com/bhilburn/powerlevel9k/tree/debug-svn-icon-color

shibumi:
@bhilburn 
I will have a look over weekend.. hope I can help.

bhilburn:
@shibumi - I'm assuming this is still an issue, yes?

shibumi:
@bhilburn yep. It's still an issue. Hadn't time yet to look over it. Hopefully sunday or Monday

bhilburn:
Fixed by #484!

PR #484, opened 2017-04-16
[SVN] removed %b identifier to fix color encoding

shibumi:
This is a workaround for https://github.com/bhilburn/powerlevel9k/issues/312 I've just added custom svn prompts instead of using our primary ones. The problem is the `%b`-identifier. When we use the `%b`-identifier for svn it seems to break our color-encoding. I guess they use some chars in there or some custom color encoding that will  break ours. The `%b`-identifier will give for SVN the following output:
`<branch-name>:<revision number of branch>`

When I use `%b` everything after `<branch-name>` will be yellow..I tried to fix the color with it. But maybe someone else has more luck.

It's sad that we lose the branch-name for this feature. But I think most people use only one branch with svn and we still display the revision-number. So it shouldn't be that bad. Let's focus on `git`. :)

The only thing what I would still miss for SVN is a nicer icon. I thought about using the `VCS_BRANCH_ICON` for this or we wait for `nerdfonts`-support.. I guess they have some cool icon for svn.

tl;dr: The svn colors are fixed now, I just removed the`%b`-identifier from the svn-bar.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1145,6 +1145,12 @@ powerlevel9k_vcs_init() {
   zstyle ':vcs_info:hg*:*' get-bookmarks true
   zstyle ':vcs_info:hg*+gen-hg-bookmark-string:*' hooks hg-bookmarks
 
+  # For svn, only
+  # TODO fix the %b (branch) format for svn. Using %b breaks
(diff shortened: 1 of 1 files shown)


bhilburn:
Nice work, @shibumi! #312 has been around for a while, now, and when I looked at it my head imploded.

Thanks for the great contribution =)

PR #486, opened 2017-04-18
Fix double expansion of prompt variables

### LINKED ISSUE #445: Possible exploit? unescaped shell meta characters
I just recently switched to using this theme and happened to read about this exploit: https://github.com/njhartwell/pw3nage

I tested it on powerlevel9k and confirmed that it is vulnerable due to not escaping meta chars where they are displayed in the prompt (in this case the branch name of the git repo).

Steps to reproduce: 
1. clone the repo: https://github.com/njhartwell/pw3nage
2. cd into the repo

The shell script included in the repo will execute due to the branch name `$(./pw3n)`

belak:
Fixes #445 

Because PROMPT and RPROMPT are in double quotes, they expand the shell commands inside them. Because prompt_subst is set, whatever these commands expand to will also be expanded when the prompt is displayed.

This fixes the problem by using single quotes, ensuring that they are only expanded once, on prompt display.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -664,7 +664,7 @@ prompt_dir() {
             break;
           fi
         done
-        
+
         local packageName=$(jq '.name' ${pkgFile} 2> /dev/null \
           || node -e 'console.log(require(process.argv[1]).name);' ${pkgFile} 2>/dev/null \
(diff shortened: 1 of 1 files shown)


bhilburn:
@belak - Nailed it! Thank you so much for taking the time to fix this and submit a PR. We really appreciate it.

Welcome to P9k, by the way! =)

Issue #459, opened 2017-03-26
Improve context segment


onaforeignshore:
I would like to suggest that the context segment be split into host and user segments for the following reasons:
· host and user can be colored separately
· we can add an icon to each (optional)
· host can be made to change color/icon for local and remote hosts
· user can be made to change color/icon for local user/root and remote user/root
· it is easier for the end user to specify which to show than to learn the correct format required by POWERLEVEL9K_CONTEXT_TEMPLATE

onaforeignshore:
I have added prompt_user and prompt_host to my repository. At the moment it does the following:

prompt_user
· changes color & icon based on user / root

prompt_host
· changes color & icon based on local / ssh

Any comments?

onaforeignshore:
<img width="1609" alt="screen shot 2017-03-26 at 14 49 47" src="https://cloud.githubusercontent.com/assets/704406/24330636/f16d101c-1233-11e7-9d06-2d5ee54ef8e1.png">

dritter:
Thanks for your input @onaforeignshore . The stakeholders of such a change would be @bhilburn and @shibumi .

Btw. Your prompt looks nice!

DanielGGordon:
That prompt is a beauty.

bhilburn:
Hey @onaforeignshore - this is a really great idea! I'm re-opening this because I don't want to lose the item. 

Can you file a PR against `next` from your fork? It would be great to start working on getting this merged. One item, though, is that we'll need to maintain the existing monolithic `context` so as not to break existing prompts. So, we would have three segments in total - `context` (current), and then two more that are individual.

shibumi:
@bhilburn so what's the difference to our current solution? Instead of a `ssh` prompt on the right we have an custom icon in front of the hostname + font-color changes. I definitly like this idea! The prompt looks awesome with the icons. I also like the style that he broke with the 'standard' `USER@HOST` prompt and did something like `HOST>USER>PWD`.

@onaforeignshore can you provide your custom `.zshrc` or changes? Would be nice to see it. I am curious now :) 

Maybe you guys have also an idea for an icon for the `virtualization-detection`.. I thought about lightning first but we have this icon already for root.

bhilburn:
@shibumi - The piece of this that's different from what we currently have are a `USER` and a `HOST` segment, which are separate from the `context` segment. Everything else that @onaforeignshore has done in that screenshot is configuration of existing features :) (I think!)

[...]

PR #489, opened 2017-04-19
User host

onaforeignshore:
As requested #459


bhilburn:
Awesome! Thanks, @onaforeignshore!

Couple of items:
1) Would you mind re-filing against `next`?
2) We need to get these segments added to the README so that users know about them
3) Can you also add them to the CHANGELOG?

PR #492, opened 2017-04-19
added host and user segments

onaforeignshore:


bhilburn:
@onaforeignshore - Thanks so much for submitting this!

It looks like you may have accidentally included `sh2md` in this PR, though. It looks like you added it as a submodule in the commit. You'll need to recreate the changeset without submodule and re-submit, I think. Unfortunately, simply adding another commit to this PR won't work because there will then exist one commit in P9k's tree where `sh2md` is in-tree.

Issue #487, opened 2017-04-19
Prompt is not working properly


maddie:
`nerdfont-complete` mode, installed with Prezto.

It looks like this with latest commit:

```
$(print_icon MULTILINE_SECOND_PROMPT_PREFIX)                                            $RPROMPT_PREFIX$(build_right_prompt)$reset_color$RPROMPT_SUFFIX
```

After reverting to 01a5c6a3 the prompt is normal again..

qifei9:
I meet the same issue, with `awesome-patched` mode.

bhilburn:
@maddie @qifei9 - Thanks for reporting this! I merged #486 to fix a security bug in #445, which looks like it may have broken something?

@maddie @qifei9 - Just to double-check:

1) You are using Prezto
2) You are using multi-lined prompt
3) You are using `nerdfont-complete`

My guess is that only #2 matters, but I want to confirm.

We'll want to move quickly to fix this since it is in master. Simply reverting the merge isn't a great option, as that re-introduces the security issue. 

@belak - Any input here would be awesome. @dritter your feedback will be important, as well.

Strayer:
I have the same issue. Using powerlevel9k directly with zplug, no special fonts either. I do have a multi-lined prompt though. Went back to 37174ea for now.

qifei9:
@bhilburn 
- I use zplug
- I use multi-lined prompt, but this issue also occurs in single-lined prompt
- I use `awesome-patched`

bhilburn:
@maddie @Strayer @qifei9 - Thanks for the comments. I have reverted the merge commit. Please checkout the `HEAD` of `master` and confirm that it's now working?

@belak @dritter - Reopening #486 so we can continue the discussion there.

qifei9:
@bhilburn 
It works at 99e59bc83a7b20287306209cecbe732d5c21c46a (current `HEAD` of `master`) and 01a5c6a3a10ff1fea9782607a431ac07652546dc.
The issue occurs at 2817b45f050e753b636ffb471413fef24b193f67 and ae259ff5940420389c6f7175d6f5a844362b56ae.

bhilburn:
@qifei9 - Thanks for confirming! 

I'd like to get one more confirmation that `HEAD` is now fixed. @Strayer, @maddie - Can you test & confirm as soon as you get the chance?

Strayer:
Master seems to work fine for me now on three machines, thanks for the quick fix!

[...]

PR #493, opened 2017-04-19
Preliminary fix for p3wnage exploit

wadkar:
* TODO:
- [ ] Need testing by others
- [ ] Does it break other `prompts`?
- [ ] Review by p3wange reporter or original author
- [ ] Remove commented code
- [ ] Thoughts on good names for the local variables such as `${PROMPT_P1}`?
- [ ] Fix coding style. I use `${}` inside double quotes for variable reuse
- [ ] Should those variables be _non-local_ to be consistent with overall coding style? 

* Details:

The construction of `${RPROMPT}` and `${PROMPT}` uses shell expansion
via `$()` that I believe is at the root of exploit.

I have extracted out the relevant expansions into a local variable, and
the construction of the prompt variables reuses those strings.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1282,8 +1282,13 @@ powerlevel9k_prepare_prompts() {
   _P9K_TIMER_START=99999999999
 
   if [[ "$POWERLEVEL9K_PROMPT_ON_NEWLINE" == true ]]; then
-    PROMPT="$(print_icon 'MULTILINE_FIRST_PROMPT_PREFIX')%f%b%k$(build_left_prompt)
-$(print_icon 'MULTILINE_SECOND_PROMPT_PREFIX')"
+    local PROMPT_P1="$(print_icon 'MULTILINE_FIRST_PROMPT_PREFIX')"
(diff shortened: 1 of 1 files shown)


wadkar:
TODO:
- [ ] Remove commented code
- [ ] Thoughts on good names for the local variables such as `${PROMPT_P1}`?
- [ ] Fix coding style. I use `${}` inside double quotes for variable reuse
- [ ] Should those variables be _non-local_ to be consistent with overall coding style?

bhilburn:
Nice work, @wadkar!

It would be great to have this tested by the folks that discovered the original breakage in #487: @Strayer @maddie @qifei9 

Original reporter of the exploit was: @sethdeckard, with confirmation from @theSoenke

Would like to have the code reviewed and tested by @dritter and @shibumi.

dritter:
Unfortunately this does not fix it on omz. I was able to reproduce the vulnerability in the test-vm again (with `vagrant-omz` user).

belak:
I believe `setopt prompt_subst` will cause this to trigger, in the same way that OMZ does without requiring the use of OMZ.

wadkar:
#496 supersedes this, closing.

PR #494, opened 2017-04-19
Use method built in to zsh to read symlinks

belak:
I'm not sure why I didn't think of this before, but there's no point in doing all this realpath/perl/whatever magic when there's something built in to zsh which does all this for us. 
The only question is whether to use the A or P modifier.

http://zsh.sourceforge.net/Doc/Release/Expansion.html#Modifiers

This supercedes #482.

CC @dritter


dritter:
LGTM. Thanks @belak 
`A` or `P` modifier seems to make no difference in my tests. :)

bhilburn:
Merging with @dritter's approval.

Thanks for the improvement, @belak! Much appreciated =)

PR #495, opened 2017-04-19
Set PROMPT_SUBST right before setting PROMPT

dritter:
This should fix substituting the prompt in all variations.
I've tested it in the VM with antigen, antibody, omz, prezto, prezto-community, zim, zplug, zplugin, zpm, zulu. And all in combination with `PROMPT_SUBST` and pw3nage.

I would appreciate other people testing this thoroughly. It might be that I haven't seen some settings/etc.

// cc @belak

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1298,6 +1298,8 @@ powerlevel9k_prepare_prompts() {
   # Reset start time
   _P9K_TIMER_START=0xFFFFFFFF
 
+  setopt PROMPT_SUBST
+
   if [[ "$POWERLEVEL9K_PROMPT_ON_NEWLINE" == true ]]; then
     PROMPT='$(print_icon 'MULTILINE_FIRST_PROMPT_PREFIX')%f%b%k$(build_left_prompt)
 $(print_icon 'MULTILINE_SECOND_PROMPT_PREFIX')'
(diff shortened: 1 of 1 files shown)


powerlevel9k.zsh-theme
```diff
@@ -1298,6 +1298,8 @@ powerlevel9k_prepare_prompts() {
   # Reset start time
   _P9K_TIMER_START=0xFFFFFFFF
 
+  setopt PROMPT_SUBST
+
   if [[ "$POWERLEVEL9K_PROMPT_ON_NEWLINE" == true ]]; then
     PROMPT='$(print_icon 'MULTILINE_FIRST_PROMPT_PREFIX')%f%b%k$(build_left_prompt)
```
belak: Also, unrelated to this PR, but I'm fairly sure this line is broken (my fault). It would probably be better to use double quotes and escape the `$(build_left_prompt)`

dritter: You are referring to the multiple single quotes, right? I have done such quoting style on other places and it seems to work correctly.. I could straighten it a bit by setting `MULTILINE_FIRST_PROMPT_PREFIX` in double quotes..

belak: Yeah, I guess I could be wrong though... if anything this set of PRs is telling me I need to go back and re-read a large portion of the documentation to make sure I know what I'm talking about.

dritter: Me too. I wish the ZSH documentation would be more accessible. 😄

dritter:
Closed in favour of #496

PR #496, opened 2017-04-20
Fix some oddities with prompt options

belak:
Firstly, `prompt_opts` needs to be set so `prompt_subst` and other prompt related options will be set consistently when this prompt is used. We also need to set them manually for when the script is executed directly (or when prompt_powerlevel9k_setup is called directly). This workaround is based off the same workaround in pure (https://github.com/sindresorhus/pure/blob/3a08df333d1a32b5be610f3cb688d5c773b8c650/pure.zsh#L397) which was based off the code from the prompt function in zsh, as seen below:

```
prompt () {
  local prompt_opts

  set_prompt "$@"

  (( $#prompt_opts )) &&
      setopt noprompt{bang,cr,percent,subst} "prompt${^prompt_opts[@]}"

  true
}
```

This introduces a complication... lots of prompts do this:

```
  setopt LOCAL_OPTIONS
  unsetopt XTRACE KSH_ARRAYS
```

I assume when starting off, powerlevel9k may have copied this from another prompt. However, because of the above workaround relating to `prompt_opts`, we cannot use local_options, so we need to look at removing the other options which have been set locally here.

Other parts of powerlevel9k currently rely on KSH_ARRAYS to not be set globally in the shell, and even some builtin functions start falling over if this is set, so I'm fairly sure this can be removed. In a similar fashion, it doesn't make a ton of sense to disable XTRACE locally for only prompt_powerlevel9k_setup, so it should be safe to disable that as well.

Sorry for the wall of text, but I felt like this was needed to explain where all these weird changes came from. On a side note, this fixes the issues some people were having with the pw3nage fix because it ensures prompt_subst will be set when using the powerlevel9k theme.

This supercedes #495

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1336,10 +1336,14 @@ prompt_powerlevel9k_setup() {
   # Maximum integer on 32-bit CPUs
   _P9K_TIMER_START=2147483647
 
+  # The prompt function will set these prompt_* options after the setup function
+  # returns. We need prompt_subst so we can safely run commands in the prompt
+  # without them being double expanded and we need prompt_percent to expand the
+  # common percent escape sequences.
   prompt_opts=(subst percent)
 
-  # borrowed from promptinit, sets the prompt options in case pure was not
-  # initialized via promptinit.
+  # Borrowed from promptinit, sets the prompt options in case the theme was
+  # not initialized via promptinit.
   setopt noprompt{bang,cr,percent,subst} "prompt${^prompt_opts[@]}"
 
   # Display a warning if the terminal does not support 256 colors
@@ -1372,12 +1376,6 @@ prompt_powerlevel9k_setup() {
     'longstatus'      'status'
   )
   print_deprecation_warning deprecated_segments
-
-  setopt prompt_subst
-
-  setopt LOCAL_OPTIONS
-  unsetopt XTRACE KSH_ARRAYS
-  setopt PROMPT_CR PROMPT_PERCENT PROMPT_SUBST MULTIBYTE
 
   # initialize colors
   autoload -U colors && colors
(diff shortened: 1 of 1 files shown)


powerlevel9k.zsh-theme
```diff
@@ -1336,6 +1336,12 @@ prompt_powerlevel9k_setup() {
   # Maximum integer on 32-bit CPUs
   _P9K_TIMER_START=2147483647
 
+  prompt_opts=(subst percent)
+
+  # borrowed from promptinit, sets the prompt options in case pure was not
```
dritter: > in case pure was..

Maybe: "in case the theme was.."

powerlevel9k.zsh-theme
```diff
@@ -1336,6 +1336,12 @@ prompt_powerlevel9k_setup() {
   # Maximum integer on 32-bit CPUs
   _P9K_TIMER_START=2147483647
 
+  prompt_opts=(subst percent)
```
dritter: Probably we should explain this line a bit more.

powerlevel9k.zsh-theme
belak: Whoops, good catch

powerlevel9k.zsh-theme
belak: Yeah, as far as I can tell it's not documented anywhere. :'(

dritter:
Great explanation! Thanks for your hard work!


I have not tested it yet, but I will tonight. :)

dritter:
Just tested it in antibody, antigen, omz, prezto, prezto-community, zgen, zim, zplug, zplugin, zpm, zulu.
Works like a charm! 👍

bhilburn:
Looks good, @belak! I really appreciate you taking the time to lay out an explanation =)

One thing I want to confirm: so, to fix the pw3nage exploit, we need this PR as well as your previous PR, #486, right?

belak:
Yes, it looks like my original PR was never reverted on this branch so that should be easy, but I'm not sure about master.

bhilburn:
You're right... it wasn't. I'm highly confused as to how that happened, actually, because I reverted it on `master` and then merged `master` into `next`.  But, sure enough, `next` is still on single quotes. WTF, git.

@belak - Thanks for continuing to work on this, and thank you, @dritter for your testing and input.

PR #497, opened 2017-04-20
host and user segments

onaforeignshore:


onaforeignshore:
This should be clean now... 🤞

shibumi:
@dritter @bhilburn can we get this merged for the next release? Or did you write the changelog already?

bhilburn:
@shibumi - I wanted to keep v0.6.3 as minimal as possible. Really, if I could have only included the fix for the pwn3nage exploit in it, I would have. I like to give new features a bit more time to bake =)

powerlevel9k.zsh-theme
```diff
@@ -567,6 +567,61 @@ prompt_context() {
...
+  if [[ "$POWERLEVEL9K_ALWAYS_SHOW_CONTEXT" == true ]] || [[ "$USER" != "$DEFAULT_USER" ]] || [[ -n "$SSH_CLIENT" || -n "$SSH_TTY" ]]; then
+    if [[ $(print -P "%#") == '#' ]]; then
+      user_state=(
+        "STATE"               "ROOT"
+        "CONTENT"             "${POWERLEVEL9K_USER_TEMPLATE}"
+        "BACKGROUND_COLOR"    "${DEFAULT_COLOR}"
+        "FOREGROUND_COLOR"    "yellow"
+        "VISUAL_IDENTIFIER"   "ROOT_ICON"
+      )
+    elif [[ "$POWERLEVEL9K_ALWAYS_SHOW_USER" == true ]]; then
```
bhilburn: I'm a little confused about the use of `ALWAYS_SHOW_CONTEXT` and `ALWAYS_SHOW_USER`, here. How are they different, and why are both used from this segment? What happens if `ALWAYS_SHOW_CONTEXT` is set to false but `ALWAYS_SHOW_USER` is set to true?

I think it might be more straight-forward to only use the `ALWAYS_SHOW_USER` flag in the the `user` segment, and add it to the README to describe what it does. Thoughts?

onaforeignshore:
What is the functionality we want? Do we want to hide this segment if `DEFAULT_USER` is a specific value? Then `ALWAYS_SHOW_USER` will override that anyway.

README.md
```diff
@@ -436,6 +467,29 @@ segment, as well:
...
+You can also override the icons by setting
+```
+POWERLEVEL9K_USER_ICON="\uF415" # 
+POWERLEVEL9K_ROOT_ICON="#"
+```
+
+| Variable | Default Value | Description |
+|----------|---------------|-------------|
+|`DEFAULT_USER`|None|Username to consider a "default context" (you can also set `$USER`).|
+|`POWERLEVEL9K_ALWAYS_SHOW_USER`|false|Always show the username, but conditionalize the hostname.|
```
bhilburn: It seems like the end of this description doesn't make sense, "...but conditionalize the hostname". I'm guessing this is just a copy/paste error?

bhilburn:
Left a minor comment in a review, above.

> What is the functionality we want? Do we want to hide this segment if DEFAULT_USER is a specific value? Then ALWAYS_SHOW_USER will override that anyway.

I think we should have the same behavior as the `context` segment. It only prints if $USER != $DEFAULT_USER, or if it's a remote session, or if 'ALWAYS_SHOW_USER' is set.

bhilburn:
Hey @onaforeignshore! I'm really excited about getting this one merged =)

Did you have any questions or want help implementing the `$DEFAULT_USER` bit we discussed, above?

I think a lot of people are really going to appreciate this feature.

onaforeignshore:
Anything else required before merge?

bhilburn:
> Anything else required before merge?

I have a comment left in a review that's still open about the README - did you see that one? That's relatively minor, though, and actually I'm planning to do some README work when I merge this so I would have fixed that on my own.

I just gave this a shot, though, and unfortunately the `user` segment is broken for me. You can see the broken rendering here:

broken_user

Looking back at the changelog, my guess is that the issue is that it is referring to the `user_state` when its in an uninitialized state (because the top-level conditional didn't fire). I suspect we should be able to trigger a similar bug with the `host_state` by playing with that a bit more. Thoughts?

README.md
onaforeignshore: Yes. I copied and pasted the `context` segment and missed that.

onaforeignshore:
I rechecked the logic - the segment rendering code should be inside the if statement - fixed that. Also removed the checks for SSH as this is the user, not the hostname. See if you still get the error please, as it works fine on my Mac.

bhilburn:
Merged!

Excellent work, @onaforeignshore . Thanks for another great PR!

Issue #500, opened 2017-04-21
Final Testing for v0.6.3


bhilburn:
With the resolution of a potential security exploit (#445), I'd like to get a new release tagged so that our packagers can get it out to our users that install via their package manager. Before I push the tag, though, I'd like to get some good thorough testing on `master`.

 @Strayer @maddie @qifei9 - Since our previous fix for #445 broke your prompts, would you mind updating your branches to the HEAD of `master` and confirming everything is good?

@dritter @shibumi @wadkar @Tritlo @onaforeignshore - Would you guys mind using `master` for a bit and confirming that everything looks good?

shibumi:
@bhilburn  first issue that I see is:

Do i need to set the color of the icon manually? Seems like it has something todo with https://github.com/bhilburn/powerlevel9k/pull/483
![](https://paste.archlinux.de/Y2A/)

bhilburn:
@shibumi - So we were previously hard-spec'ing `green` as the color, which means that it wouldn't change with the terminal colors, making it unreadable if you were actually using terminal themes. The color you are seeing, there, is actually what your terminal is using for `green`. If you want it to be the original bright green, you'll need to either change your terminal profile or override the color setting.

shibumi:
@bhilburn 
Yep, that's what I have thought already. So seems like I need to look in the commit history to get the icon more bright again.

bhilburn:
@shibumi - You should be able to just do:
```
POWERLEVEL9K_STATUS_OK_FOREGROUND="green"
```

dritter:
`master` works for me! :)

shibumi:
I have found different issues:

1. The dir_writable icon looks wrong for me.
2. the `detect_virt` segment seems broken

![](https://paste.archlinux.de/YTpgLr/)

V1rgul:
use `POWERLEVEL9K_DIR_WRITABLE_FORBIDDEN_FOREGROUND='black'`

shibumi:
@V1rgul Thanks I have mentioned this in the wiki now: https://github.com/bhilburn/powerlevel9k/wiki/Stylizing-Your-Prompt

wadkar:
Thank for this, I am on `master` (with `nerdfont-complete`) and its working fine on macOS El Capitan and Ubuntu 16.04

bhilburn:
Okay, the release has been tagged and pushed!

https://github.com/bhilburn/powerlevel9k/releases/tag/v0.6.3

PR #502, opened 2017-04-23
[detect-virt] quickfix for unnecessary information in prompt

shibumi:
this fixes the broken prompt mentioned in https://github.com/bhilburn/powerlevel9k/issues/500


bhilburn:
Thanks so much for the quick-fix, @shibumi! Nice work!

PR #503, opened 2017-04-23
Bring back space after branch indicator in default mode

Serede:
I don't understand why it got removed, it looks so ugly D:

### DIFF
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -265,7 +265,7 @@ case $POWERLEVEL9K_MODE in
       VCS_TAG_ICON                   ''
       VCS_BOOKMARK_ICON              $'\u263F'              # ☿
       VCS_COMMIT_ICON                ''
-      VCS_BRANCH_ICON                $'\uE0A0'              # 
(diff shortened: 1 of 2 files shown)


bhilburn:
Hey @Serede!

Hah, it actually didn't have a space there for the entire history of the project, I added it at one point, and then removed it due to popular demand (or backlash, I suppose), hah.

Anyway, the best way to add this back should be to simply customize the icon in your zshrc, actually. Per:
https://github.com/bhilburn/powerlevel9k/wiki/Stylizing-Your-Prompt#icon-customization

If you put the following in your `.zshrc`:
```
POWERLEVEL9K_VCS_BRANCH_ICON=$'\uE0A0 '
```

It should do what you need without modifying the code. And, actually, looking at these changes, I would actually expect that this introduced a  double-space after the branch icon - did it not?

Serede:
Thanks @bhilburn!

I didn't find that setting anywhere, that's why I made this pull request.
Anyway, it does what I wanted indeed, so sorry for any inconvenience.

Greetings.

bhilburn:
Glad to hear you're all set, @Serede!

PR #508, opened 2017-04-28
Set necessary prompt expansion options as late as possible

### LINKED ISSUE #506: Custom icons in prompt broken
I am desperate, dunno if it's a bug in powerlevel9k or oh-my-zsh, but looks like powerlevel9k. Happened today out of seemingly nowhere (I might have git pulled all my plugins yesterday)

http://i.imgur.com/cZ2wZMC.png

dritter:
Set PROMPT_SUBST and PROMPT_PERCENT right before we set PROMPT and
RPROMPT variables. This should fix the prompt even if you did not set PROMPT_SUBST yourself (or the used ZSH framework).

I don't know why I did not found this issue when I tested #496 , and it is still a mystery to me how PROMPT_SUBST is inherited to later called functions..

This should fix #506 

FYI: printing icons like `POWERLEVEL9K_DIR_PATH_SEPARATOR=' $(print_icon "LEFT_SUBSEGMENT_SEPARATOR") '` as @bhilburn explains in https://github.com/bhilburn/powerlevel9k/issues/506#issuecomment-298022553 can't be used any more, so this is not a fix for that.

I haven't had much time to test it carefully, so please double check it!
Btw. I'll have a look at the failing tests when I am back from vacation..

@onaforeignshore Could you confirm this fixes your issue?

//cc @belak

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1290,6 +1290,12 @@ powerlevel9k_preexec() {
 
 set_default POWERLEVEL9K_PROMPT_ADD_NEWLINE false
 powerlevel9k_prepare_prompts() {
+  # The prompt function will set these prompt_* options after the setup function
+  # returns. We need prompt_subst so we can safely run commands in the prompt
+  # without them being double expanded and we need prompt_percent to expand the
+  # common percent escape sequences.
+  setopt PROMPT_SUBST PROMPT_PERCENT
+
   RETVAL=$?
 
   _P9K_COMMAND_DURATION=$((EPOCHREALTIME - _P9K_TIMER_START))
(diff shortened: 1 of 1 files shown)


onaforeignshore:
It fails on these lines.
```zsh
# Borrowed from promptinit, sets the prompt options in case the theme was
# not initialized via promptinit.
setopt noprompt{bang,cr,percent,subst} "prompt${^prompt_opts[@]}"
```
Should we move them or delete them?

belak:
This change doesn't seem right. The user shouldn't be messing with prompt_ settings and this code ensures they've been been set properly.

belak:
Anyway, this might work if that set of code was removed as @onaforeignshore... But it seems like a hacky work around

belak:
For clarification, the reason prompt_opts is used is because of this function from zsh which is defined when promptinit is called:

```
prompt () {
  local prompt_opts

  set_prompt "$@"

  (( $#prompt_opts )) &&
      setopt noprompt{bang,cr,percent,subst} "prompt${^prompt_opts[@]}"

  true
}
```

The workaround we're using for when a user doesn't use the `prompt` function is lifted directly from here. There's a chance that might not be working as expected?

dritter:
Ah. I see. Thanks for clarification @belak .

I close this PR, as we should investigate more why @onaforeignshore s configuration does not work as expected.

PR #513, opened 2017-05-03
Make sure prompt_cr is set

### LINKED ISSUE #511: [Regression: 0.6.3] Visual artefacts with quick consecutive commands
I just upgraded from v0.6.2 to v0.6.3, and I've noticed some visual artefacts when I type rapidly after the previous command. For example, in v0.6.3, I type `l` (an alias for ls) twice quickly.
```
 /path > l
foo bar
l /path >l
```
You can see the extra `l` at the beginning of the second prompt, and the shift right of the prompt. This is a fairly facile example, but the artefacts sometimes result in very garbled text, forcing me to cancel the current command.

I can make the above occur ~50% of the time if I alternate between `l` and `Enter` as fast as possible. In contrast, I downgraded to v0.6.2, and can no longer replicate this. Perhaps 20% of the time I can type it fast enough that the prompt isn't drawn yet, but it writes on its own line, and doesn't cause artefacts.
```
 /path > l
foo bar
l
 /path > l
```
 If you cannot replicate this issue, I'm happy to test a few specific commits, if you can point me in the right direction. Cheers.

belak:
Fixes #511

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1339,7 +1339,7 @@ prompt_powerlevel9k_setup() {
   # returns. We need prompt_subst so we can safely run commands in the prompt
   # without them being double expanded and we need prompt_percent to expand the
   # common percent escape sequences.
-  prompt_opts=(subst percent)
+  prompt_opts=(subst percent cr)
 
   # Borrowed from promptinit, sets the prompt options in case the theme was
   # not initialized via promptinit.


bhilburn:
Tested and confirmed it works. Thanks so much for the quick response, @belak! Well done =)

wadkar:
@belak - any helpful URL/blog post on `prompt_opts`. Simple keyword search on the interweb did not yield pointers to correct manual. I would love to read and learn about it!

dritter:
@wadkar Here you go: http://zsh.sourceforge.net/Doc/Release/Options.html

belak:
Thanks @dritter. That will go over all the options available (look for ones starting with `prompt_`). `prompt_opts` itself is an oddity that doesn't appear to be documented anywhere but is used in most prompts. That's a part of the `prompt` function.

dritter:
@wadkar ah. yes. `prompt_opts` is used by the `promp` function. That is the explanation why we must set it as global variable and why `LOCAL_OPTS` must be "off". Have a look [here](https://github.com/zsh-users/zsh/blob/21b195562a54005b4341a037d2c02078787b7f2a/Functions/Prompts/promptinit#L171-L180) to see what happens.
Line 177 is the interesting one: it unsets all prompt related options and sets them again based on the contents of `prompt_opts`.

I hope that explains it a bit.

wadkar:
Ah, I see. This is fascinating. I think this entire _settings_ gymnastic deserves some documentation. Is there something we can do/submit upstream for the sake of posterity?

PR #515, opened 2017-05-04
status: add option for support pipes

### LINKED ISSUE #514: Capture exit codes for pipes, not just the last command of a pipe.
The current implementation only shows the exit code for the last command in a command with pipes.

Instead, I would suggest showing all the exit codes.

Current behavior (assuming `PIPE_FAIL` isn't set):

``` sh
$ true | false | false | true # Shows checkmark
$ true | false | false # Shows 1 ↵
```

Instead, the `status` segment should show:

```
$ true | false | false | true # Shows 0|1|1|0 ↵
$ true | false | false # Shows 0|1|1 ↵
```

Potentially, `0` could be a checkmark.

To do this, you need to capture `$pipestatus[@]`.

docwhat:
Added new option (POWERLEVEL9K_STATUS_SHOW_PIPESTATUS) that shows
the status for all piped commands, not just the right-most.

Closes #514

### DIFF
--- a/README.md
+++ b/README.md
@@ -463,7 +463,7 @@ This segment shows the return code of the last command.
 |----------|---------------|-------------|
 |`POWERLEVEL9K_STATUS_VERBOSE`|`true`|Set to false if you wish to not show the error code when the last command returned an error and optionally hide this segment when the last command completed successfully by setting `POWERLEVEL9K_STATUS_OK_IN_NON_VERBOSE` to false.|
 |`POWERLEVEL9K_STATUS_OK_IN_NON_VERBOSE`|`false`|Set to true if you wish to show this segment when the last command completed successfully in non-verbose mode.|
-|`POWERLEVEL9K_STATUS_SHOW_PIPESTATUS`|`false`|Set to true if you wish to show the exit status for all piped commands.|
+|`POWERLEVEL9K_STATUS_SHOW_PIPESTATUS`|`true`|Set to true if you wish to show the exit status for all piped commands.|
 
 ##### ram
 
--- a/README.md
+++ b/README.md
@@ -463,6 +463,7 @@ This segment shows the return code of the last command.
 |----------|---------------|-------------|
 |`POWERLEVEL9K_STATUS_VERBOSE`|`true`|Set to false if you wish to not show the error code when the last command returned an error and optionally hide this segment when the last command completed successfully by setting `POWERLEVEL9K_STATUS_OK_IN_NON_VERBOSE` to false.|
 |`POWERLEVEL9K_STATUS_OK_IN_NON_VERBOSE`|`false`|Set to true if you wish to show this segment when the last command completed successfully in non-verbose mode.|
+|`POWERLEVEL9K_STATUS_SHOW_PIPESTATUS`|`false`|Set to true if you wish to show the exit status for all piped commands.|
 
 ##### ram
 
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1019,7 +1019,7 @@ prompt_ssh() {
 # Status: return code if verbose, otherwise just an icon if an error occurred
 set_default POWERLEVEL9K_STATUS_VERBOSE true
 set_default POWERLEVEL9K_STATUS_OK_IN_NON_VERBOSE false
-set_default POWERLEVEL9K_STATUS_SHOW_PIPESTATUS false
+set_default POWERLEVEL9K_STATUS_SHOW_PIPESTATUS true
 prompt_status() {
   local ec_text
   local ec_sum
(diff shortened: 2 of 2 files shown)


powerlevel9k.zsh-theme
```diff
@@ -1004,10 +1004,30 @@ prompt_ssh() {
 # Status: return code if verbose, otherwise just an icon if an error occurred
 set_default POWERLEVEL9K_STATUS_VERBOSE true
 set_default POWERLEVEL9K_STATUS_OK_IN_NON_VERBOSE false
+set_default POWERLEVEL9K_STATUS_SHOW_PIPESTATUS false
```
bhilburn: Nice work using our built-ins, especially for a first-time contributor =)

So, actually, I think I would like for this to default to being `true` - I think it's a really great improvement. You'll also need to update the docs in the README to reflect this.

README.md
```diff
@@ -414,6 +414,7 @@ This segment shows the return code of the last command.
 |----------|---------------|-------------|
 |`POWERLEVEL9K_STATUS_VERBOSE`|`true`|Set to false if you wish to not show the error code when the last command returned an error and optionally hide this segment when the last command completed successfully by setting `POWERLEVEL9K_STATUS_OK_IN_NON_VERBOSE` to false.|
 |`POWERLEVEL9K_STATUS_OK_IN_NON_VERBOSE`|`false`|Set to true if you wish to show this segment when the last command completed successfully in non-verbose mode.|
+|`POWERLEVEL9K_STATUS_SHOW_PIPESTATUS`|`false`|Set to true if you wish to show the exit status for all piped commands.|
```
bhilburn: Thanks for adding docs!

powerlevel9k.zsh-theme
```diff
@@ -1004,10 +1004,30 @@ prompt_ssh() {
...
+      ec_sum=$(( $ec_sum + $ec ))
+    done
+  else
+    # We use RETVAL instead of the right-most RETVALS item because
+    # PIPE_FAIL may be set.
+    ec_text=${RETVAL}
+    ec_sum=${RETVAL}
+  fi
+
+  if (( ec_sum > 0 )); then
```
bhilburn: Wouldn't we want this to check that it is not equal to zero, as it it did previously with `RETVAL`? Often return codes are negative.

docwhat:
> You'll need to close this PR and submit a new one, and Github doesn't allow you to change the branch you have filed against (I think).

Son of a Guinea Pig!  I just discovered that you can after manually closing and re-opening all the other PRs!

You just click "Edit" at the top (next to the title for the PR) and you can change it there.  As an admin, you should have also been able to do that.

powerlevel9k.zsh-theme
docwhat: 👍

powerlevel9k.zsh-theme
docwhat: The exit codes should _never_ be negative.  It is by definition an unsigned 8-bit int. The first 7 bits are for "normal" exit codes (0–127).  If the 8th bit is set then the process was killed by an exit code  For example:

``` sh
(sleep 10 ; a=$? ; echo "\$?: $a   8th bit: $(( a >> 7 ))   low bits $(( a & 127 ))" ) & sleep 1 ; kill -9 $(pidof sleep)
[1] 30696
$?: 137   8th bit: 1   low bits 9
[1]  + 30696 done       ( sleep 10; a=$? ; echo ; )
```

In the example:

- `$?` was 137
- The 8th bit was set.
- The low bits were `9`
- If we used a signed 8-bit int, this would be `-9` but shells don't work that way.

I couldn't find the `zsh` references for `$?` but I'm pretty sure it follow the posix standard.

I would be interested in finding out I'm wrong:  Do you have a command that returns a less-than-zero exitcode?

README.md
docwhat: 👍

powerlevel9k.zsh-theme
bhilburn: You're right - exit codes are always positive. I genuinely didn't know that!

bhilburn:
> Son of a Guinea Pig! After manually closing and re-opening all the other PRs, I discovered that you can change the base branch!
> 
> You just click "Edit" at the top (next to the title for the PR) and you can change it there. As an admin, you should have also been able to do that.

So, this would have worked, but wouldn't have been ideal. I could have changed the target of your merge from `master` to `next`, but because you based your original branch on `master` instead of `next`, if I had merged it into `next` it would have carried all of the other `master` commits that were parents of your branch, going back to the first common ancestor of `master` and `next`. 

Realistically, this wouldn't have done any damage to the working status, as I try to make sure changes on `master` gets merged into `next` regularly so that bug fixes propagate. The issue is that if we ever needed to revert this merge, or simply wanted to look at the diff created by it, it would have been polluted by the changes on master. In short, it would have worked, but would have made for a messy git history that probably would have been fine but I which I like to avoid =)

Thanks so much for re-creating your PRs. I really appreciate it! This is a really great contribution - thank you 💯

powerlevel9k.zsh-theme
docwhat: Some programs, especially older ones, cast the exit code to a signed int.  This lets you tell if it was killed by a signal (negative number) or normal exit code.

In modern programs you have explicitly cast that, but I'm sure some still do.

docwhat:
True, you couldn't have changed it, but _I_ can if I rebase my patch off the `next` branch.  I'll try to remember next time.

PR #516, opened 2017-05-04
nvm: improve default detection

docwhat:
This is also (marginally) faster than before.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -874,13 +874,16 @@ prompt_node_version() {
 # Node version from NVM
 # Only prints the segment if different than the default value
 prompt_nvm() {
-  [[ ! $(type nvm) =~ 'nvm is a shell function'* ]] && return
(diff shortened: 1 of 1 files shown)


docwhat:
As a side-effect of these changes, it won't trigger `nvm` if you're using the zsh-nvm plugin and have lazy loading set.

With the previous code, there was no point to setting lazy load because the prompt would load it immediately.

bhilburn:
Thanks for the PR, @docwhat!

Not being a `nvm` user, or even a Node.js developer, myself, I'm not knowledgeable enough to review the code. That said, a couple of items:

1) Would you mind re-submitting this PR against `next`?
2) What happens if `nvm` isn't loaded due to the lazy loading bit you mentioned above? Does the segment just not print (fail gracefully)?

I'd like to get @dritter's review on your new PR because I believe he is a Node.js user? ;)

docwhat:
Closing in favor of #526

PR #517, opened 2017-05-04
add function to disable powerlevel9k

docwhat:
Sometimes you have to disable your prompt. :-(


bhilburn:
Interesting idea, @docwhat! I can think of a few instances where this might be useful.

It seems like this might be a difficult way to use the functionality, though, because you would need to modify your P9k segments list, add this segment, and then reload your `.zshrc`, right? Am I misunderstanding the implementation?

It seems like it would be more useful if there was some way we could simply provide a function that turned it on/off, although realistically I don't know that that's possible with ZSH.

Also, this will need to be filed against `next` =)

PR #518, opened 2017-05-04
newline: add new segment

docwhat:
I'd appreciate any feedback on this.


bhilburn:
Wow! Very col idea, @docwhat! I think a lot of people will really like this feature.

A couple of items:
1) What happens if the user turns this on and also turns on `PROMPT_ON_NEWLINE`?
2) Can you file against `next`? ;)
3) We should be sure to add a bit about this to the Wiki once it is merged.
4) Is there any limit to the number of times you can use this segment in a prompt?

docwhat:
Closed in favor of #524

PR #524, opened 2017-05-13
"newline" segment

docwhat:
This works only for `POWERLEVEL9K_LEFT_PROMPT_ELEMENTS`.

It can't work for the right-hand prompt (`RPROMPT`) because multi-line is handled
via terminal trickery (moving the cursor and restoring it).

This is against the `next` branch as requested by @bhilburn on #518

@bhilburn asked:

> 1. What happens if the user turns this on and also turns on PROMPT_ON_NEWLINE?
> 1. Can you file against `next`? ;)
> 1. We should be sure to add a bit about this to the Wiki once it is merged.
> 1. Is there any limit to the number of times you can use this segment in a prompt?

1. It works, but it doesn't have the appropriate ansi art in first two characters... Let me see if I can fix that.
1. Done!
1. 👍 😃 
1. Nope.  It works all you want, though it it can look silly. :smiley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,7 @@ The segments that are currently available are:
 * [`command_execution_time`](#command_execution_time) - Display the time the current command took to execute.
 * [`todo`](http://todotxt.com/) - Shows the number of tasks in your todo.txt tasks file.
 * `detect_virt` - Virtualization detection with systemd
+* `newline` - Continues the prompt on a new line.
 
 ---------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


docwhat:
The newline segment will now show `├─` when using
`POWERLEVEL9K_PROMPT_ON_NEWLINE`

This can be overridden via `POWERLEVEL9K_MULTILINE_FIRST_PROMPT_PREFIX`.

dritter:
This is quite interesting. Thanks for that Idea!

I always thought about having an easy way to align segments not only horizontally, but also vertically. This is a way to do this on the left side. 👍 

Btw. I suggest renaming `MULTILINE_SECOND_PROMPT_PREFIX` to `MULTILINE_LAST_PROMPT_PREFIX` (or similar). "Second" is not quite right in this context.

bhilburn:
Thanks for another great PR, @docwhat! I appreciate the answers to my questions, and the updates / changes. I also liked @dritter's comment on the variable naming, so I made that change prior to merging.

This has been merged into `next` =)

PR #525, opened 2017-05-13
add function to disable powerlevel9k hooks

docwhat:
This disables the hooks.  If you try to change the prompt (e.g. `PROMPT='$ '`) then
it'll revert right back to powerlevel9k because the hooks rebuild the `PROMPT`s.

Usage:

``` sh
 OSX > ~/.zgen/bhilburn/powerlevel9k-master > pr/teardown ↑1 >
 ✓ > prompt_powerlevel9k_teardown
lithium% echo "look ma, no powerlevel9k prompt"
look ma, no powerlevel9k prompt
lithium% prompt_powerlevel9k_setup
 OSX > ~/.zgen/bhilburn/powerlevel9k-master > pr/teardown ↑1 >
 ✓ >
```

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1412,4 +1412,9 @@ prompt_powerlevel9k_setup() {
   add-zsh-hook preexec powerlevel9k_preexec
 }
 
+prompt_powerlevel9k_teardown() {
+  add-zsh-hook -D precmd powerlevel9k_\*
+  add-zsh-hook -D preexec powerlevel9k_\*
+}
+
 prompt_powerlevel9k_setup "$@"
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1415,6 +1415,8 @@ prompt_powerlevel9k_setup() {
 prompt_powerlevel9k_teardown() {
   add-zsh-hook -D precmd powerlevel9k_\*
   add-zsh-hook -D preexec powerlevel9k_\*
+  PROMPT='%m%# '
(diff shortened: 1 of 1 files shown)


docwhat:
Closes #517

docwhat:
@bhilburn asked on #517:

> Interesting idea, @docwhat! I can think of a few instances where this might be useful.
>
> It seems like this might be a difficult way to use the functionality, though, because you would need to modify your P9k segments list, add this segment, and then reload your `.zshrc`, right? Am I misunderstanding the implementation?
>
> It seems like it would be more useful if there was some way we could simply provide a function that turned it on/off, although realistically I don't know that that's possible with ZSH.
>
> Also, this will need to be filed against `next` =)

It isn't super useful, but better than nothing.  I explained the usage above.  Hopefully that explains it.

I really should file a feature request for ZSH's `prompt` to look for a `prompt_<name>_teardown` method when switching _away_ from a prompt.   This would complement `prompt_<name>_setup`.

docwhat:
Another approach would be what the shipped [`bart` prompt](https://github.com/zsh-users/zsh/blob/ed4e37e45c2f5761981cdc6027a5d6abc753176a/Functions/Prompts/prompt_bart_setup#L181-L187) does.

`prompt bart off` turns it off

bhilburn:
@docwhat - I like this! Great solution. I fixed a typo in the code that prevented `RPROMPT` from getting blown away before merging, and added a commit that calls out this new feature in the README (see the bottom of the README =) )

Thanks again for the PR! 👍

PR #531, opened 2017-05-21
swapped colors for virtualization detection

shibumi:
black on yellow was pain for the eyes


bhilburn:
Hah, fair point, @shibumi  =)

PR #535, opened 2017-05-25
Add Gopher icon to go_version segment

zikes:
The Gopher icon is provided by nerd-font.


bhilburn:
Thanks for another contribution, @zikes! Simple, effective =)

PR #538, opened 2017-05-30
add VPN IPs to prompt

golgoth31:
add function to get vpn ips

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1062,6 +1062,17 @@ prompt_symfony2_version() {
   fi
 }
 
+# prompt if vpn active
+prompt_vpn_ip() {
+  if [[ -n "$POWERLEVEL9K_VPN_IP_INTERFACE" && -x /usr/bin/nmcli ]]; then
+    for vpn_iface in $(/usr/bin/nmcli connection show --active |grep $POWERLEVEL9K_VPN_IP_INTERFACE | awk '{print $4}')
+    do
+      ip=$(ip -4 a show "$vpn_iface" | grep -o "inet\s*[0-9.]*" | grep -o "[0-9.]*")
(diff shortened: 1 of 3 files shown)


shibumi:
I like the idea of VPN prompts, but isn't there a better way as using `nmcli`? I know enough people that don't use `nmcli`.

onaforeignshore:
`nmcli` is also linux specific, so no support on macOS

onaforeignshore:
On macOS, VPN IP address can be found using: `ifconfig ppp0 | grep "inet " | grep -v 127.0.0.1 | awk '{print $2}'`. If VPN is inactive the result is `ifconfig: interface ppp0 does not exist`

golgoth31:
Hi, thank you
I don't have any macOS so I can't test it. I will try to improve it through ifconfig or ip command

golgoth31:
Done using "ip" command and no more nmcli.

powerlevel9k.zsh-theme
```diff
@@ -1062,6 +1062,18 @@ prompt_symfony2_version() {
   fi
 }
 
+set_default POWERLEVEL9K_VPN_IP_INTERFACE "tun"
+# prompt if vpn active
+prompt_vpn_ip() {
+  if [[ -x /usr/bin/nmcli ]]; then
+    for vpn_iface in $(/bin/ip tuntap | grep -e ^"$POWERLEVEL9K_VPN_IP_INTERFACE" | cut -d":" -f1)
```
shibumi: Don't use the absolut path here. The path to `ip` can change, depending on OS and distribution.

golgoth31:
Ok, I removed the full path for the "ip" command

shibumi:
Ok thanks, one more thing.. can you use the `next` branch instead of `master`? We work on the `next` branch for future releases and it would be also nice to have the commits `rebased` to one commit via `git rebase -i ...` :)

golgoth31:
see pull #539

PR #539, opened 2017-05-31
add vpn ip to prompt

golgoth31:
move from master to next branch

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ The segments that are currently available are:
 * `history` - The command number for the current line.
 * [`host`](#host) - Your current host name
 * [`ip`](#ip) - Shows the current IP address.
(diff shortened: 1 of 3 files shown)


shibumi:
@bhilburn What do you think of this? :)

bhilburn:
@golgoth31 - Nice work! This is a really great PR, and a fantastic first contribution to the project. Welcome to P9k =)

@shibumi @onaforeignshore - Thanks for providing review & feedback in #538!

onaforeignshore:
This prompt is Linux only at the moment and does not support macOS, which does not have `ip`. There is a way to get the info on macOS, as described in #538

bhilburn:
@onaforeignshore - Ah, this needs to be fixed, then, or at the very least tracked. Just opened #540 for this. Thanks!

Issue #540, opened 2017-06-02
`vpn` Segment Needs macOS Support


bhilburn:
Per @onaforeignshore's comment in #539: https://github.com/bhilburn/powerlevel9k/pull/539#issuecomment-305855991

The `vpn` segment currently does not support macOS. @onaforeignshore offered a way to implement this in #538:

>On macOS, VPN IP address can be found using: 
`ifconfig ppp0 | grep "inet " | grep -v 127.0.0.1 | awk '{print $2}'`. If VPN is inactive the result is ifconfig: interface ppp0 does not exist

This simply needs to be conditionalized into the `vpn` segment code. If someone with a Mac is interested in taking this on and testing it, that would be great.

Stretch goal: unit tests for this segment =)

guidoilbaldo:
Hi @bhilburn,

just sharing my thoughts on this. The solution proposed by @onaforeignshore will work only with PPTP VPNs, while using other common tools (for a Mac OS user), for example OpenVPN or Tunnelblick, will result in having net ifaces with different names - respectively tun0 and utun0.
I would think of something more generic, then I'll be glad to help with the troubleshooting :)
Thanks again for the great work everyone's doing here!

bhilburn:
@guidoilbaldo - That's good to know, @guidoilbaldo! Thanks so much for the input! It's much appreciated.

golgoth31:
Hi, I made a test on tunnelbrick and it works; the default value is "tun" but it can be overridden in zshrc.

bhilburn:
I am on a issue-cleaning spree. This issue hasn't had motion in quite some time, and so I am closing it out due to inactivity. If anyone would like to re-open it because they feel it is un-resolved and is something they care about, please feel free to re-raise the discussion!

PR #541, opened 2017-06-03
vpn prompt: change "ip" to "ifconfig" to be used on OSX

golgoth31:
Update of the vpn prompt to use /sbin/ifconfig instead of ip

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -917,13 +917,12 @@ prompt_ip() {
 set_default POWERLEVEL9K_VPN_IP_INTERFACE "tun"
 # prompt if vpn active
 prompt_vpn_ip() {
-  for vpn_iface in $(ip tuntap | grep -e ^"$POWERLEVEL9K_VPN_IP_INTERFACE" | cut -d":" -f1)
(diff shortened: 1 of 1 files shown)


bhilburn:
Thanks so much for the PR, @golgoth31! Sorry you went for a few weeks, there, without a response.

We've had a couple of other suggestions for how to handle this in #540 from @onaforeignshore and @guidoilbaldo. Unfortunately, not being a macOS user myself, I have no way of testing these.

If some of our macOS devs could review & test this, and leave comments, that would be fantastic `=)`

golgoth31:
Hi, I found a mac and made some test so, it seems that using ifconfig is working on both Linux and macOS. Of course, If some one knowing macOS well could test and validate it's much better !

bhilburn:
I don't want to let this one bitrot. Are there any other macOS users that wouldn't mind giving this a shot to verify all is well?

guidoilbaldo:
hey, had the chance to test the override in zshrc for other interfaces name (like "tun" for Tunnelblick") as @golgoth31 suggested. It works like a charm, so you can merge this branch :)

bhilburn:
Thanks for the confirmation, @guidoilbaldo! 

@golgoth31 - Great contribution, and thanks for the PR! Welcome to P9k as a contributor `=)`

PR #547, opened 2017-06-11
Fixed RVM segment for usernames with dashes

D-side:
Having a username with a dash, such as mine, `d-side`, currently results in RVM segment displaying the wrong thing, e. g. `side/.rvm/rubies/ruby` (for `MY_RUBY_HOME=/home/d-side/.rvm/rubies/ruby-2.4.1`).

This is a kind of a dirty fix for the issue, I just forced `awk` to grab the last column, instead of 2nd. I'm not particularly fond of this approach, but this does work for my cases and doesn't (probably) introduce additional breakage.

Additional suggestions welcome.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -990,7 +990,7 @@ prompt_rvm() {
   local gemset=$(echo $GEM_HOME | awk -F'@' '{print $2}')
   [ "$gemset" != "" ] && gemset="@$gemset"
 
-  local version=$(echo $MY_RUBY_HOME | awk -F'-' '{print $2}')
+  local version=$(echo $MY_RUBY_HOME | awk -F'-' '{print $NF}')
 
   if [[ -n "$version$gemset" ]]; then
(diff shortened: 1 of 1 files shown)


bhilburn:
@D-side - Thanks so much for the PR! And wow, nice edge case! `;)`

To be honest, I think your fix looks good. We are using `awk`, anyway, so your change, here seems reasonable. Especially since this is a PR against `master`, though (which is correct as it's a bugfix), we need to be absolutely sure it's solid before merging.

It would be great if some of our Ruby devs could test & confirm this. @dritter, perhaps?

dritter:
Sorry for the late review. I am not much an `awk` pro.

Before it printed the second part after a dash. Now it prints the last part (which is way better).
bildschirmfoto 2017-06-29 um 22 41 47

Nice work @D-side 👍

bhilburn:
@D-side - Excellent work! Thanks so much for the PR, and welcome to the P9k contributor circle!

PR #536, opened 2017-05-26
Added kubectl config's current context

jtaylor32:
I created this little addition locally and thought someone else might enjoy this. It's the current context of your `kubectl` config. Whatever your current is set to it will show in the prompt with the Unicode [Helm](http://www.fileformat.info/info/unicode/char/2388/index.htm) symbol.

Let me know if I need to add anything else or change the function!

PR #548, opened 2017-06-12
Provide support for a prompt option for kubernetes-cli.

derekbassett:
The current context is useful information much like docker_machine and current working directory.
The current context configured inside .kube/config directory points to the current k8s cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ The segments that are currently available are:
     * [`aws`](#aws) - The current AWS profile, if active.
     * `aws_eb_env` - The current Elastic Beanstalk Environment.
 * `docker_machine` - The current Docker Machine.
+* `kubernetes` - The current Kubernetes Cluster.
 
 **Other:**
 * [`custom_command`](#custom_command) - Create a custom segment to display the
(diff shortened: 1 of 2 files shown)


derekbassett:
Any idea on if this will be included?

zikes:
This seems to work similarly to https://github.com/bhilburn/powerlevel9k/pull/536, is that correct?

bhilburn:
Hey @derekbassett! Thanks so much for the PR! Sorry for the delayed response - catching up on things, now.

As @zikes pointed out, we just merged a new Kubernetes segment in #536 (into `next`). I'm not a Kubernetes user, so honestly I can't speak knowledgeably about what the differences are. Would you mind having a look at that PR, or checking out the HEAD of `next` and giving it a shot, and providing some insight?

derekbassett:
So I agree #536 is a better solution and didn't notice it when I look through the issues.  I will point to next and use it instead.  Feel free to Close this PR without merging.

bhilburn:
Sounds good. Thanks so much, @derekbassett! `=)`

PR #556, opened 2017-06-26
Fixed no auth info in kubecontext

derekbassett:
Fixed when no auth info is inside the contexts


derekbassett:
Going to rework and resubmit

bhilburn:
Roger!

PR #560, opened 2017-07-04
add option for the status segment : error in non verbose

AdrienHorgnies:
The following option values will make the segment to only show the error
code if any :

```sh
POWERLEVEL9K_STATUS_VERBOSE=false
POWERLEVEL9K_STATUS_ERROR_IN_NON_VERBOSE=true
POWERLEVEL9K_STATUS_OK_IN_NON_VERBOSE=false # it's the default
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,8 +412,8 @@ This segment shows the return code of the last command.
 
 | Variable | Default Value | Description |
 |----------|---------------|-------------|
-|`POWERLEVEL9K_STATUS_VERBOSE`|`true`|Set to false if you wish to not show the error code when the last command returned an error and optionally hide this segment when the last command completed successfully by setting `POWERLEVEL9K_STATUS_OK_IN_NON_VERBOSE` to false.|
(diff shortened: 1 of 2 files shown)


AdrienHorgnies:
I didn't do test yet but I wanted to PR to ask you your opinion on this.

bhilburn:
@AdrienHorgnies - Thanks so much for the PR! It's been a while since this segment has had any work done to it - thanks for making improvements!

Regarding your proposal in your second comment, I see what you are going for, and I'm always a fan of simplification. The key to this will be doing it in a way that is backwards-compatible. Breaking existing configurations is something we try very, very hard not to do. What we could do, though, is implement your suggestion in a way that still supports the old way, but going forward only document your simplified approach.

Are you up for taking a shot at that?

AdrienHorgnies:
Yes, I should be able to do something backward compatible. There will be some condition into the if clause that will be there for the sake of backward compatibility.

I'll do that when I'm at home.

AdrienHorgnies:
I'm closing this pull request because it is into **master** and your documentation asks to do such PR into **next**. I'm opening a new PR correctly set up : #573

PR #562, opened 2017-07-08
Cleaned up installation directory detection

docwhat:
- Used ZSH built-in magic to get the installation directory.
- Works with symlinks.
- No longer calls out to external program 'dirname'.
- Fixed issue with spaces in the path.
- No longer leak the variables 'filename' and 'script_location'.

If this fails to work, then ZSH itself is broken.  The error handling is no longer needed.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -17,74 +17,57 @@
 ################################################################
 
 ## Turn on for Debugging
+#PS4='%s%f%b%k%F{blue}%{λ%}%L %F{240}%N:%i%(?.. %F{red}%?) %1(_.%F{yellow}%-1_ .)%s%f%b%k	'
 #zstyle ':vcs_info:*+*:*' debug true
 #set -o xtrace
 
 # Try to set the installation path
-if [[ -n "$POWERLEVEL9K_INSTALLATION_PATH" ]]; then
-  # If an installation path was set manually,
-  # it should trump any other location found.
-  # Do nothing. This is all right, as we use the
-  # POWERLEVEL9K_INSTALLATION_PATH for further processing.
-elif [[ $(whence -w prompt_powerlevel9k_setup) =~ "function" ]]; then
-  # Check if the theme was called as a function (e.g., from prezto)
-  autoload -U is-at-least
-  if is-at-least 5.0.8; then
-    # Try to find the correct path of the script.
-    POWERLEVEL9K_INSTALLATION_PATH=$(whence -v $0 | sed "s/$0 is a shell function from //")
-  elif [[ -f "${ZDOTDIR:-$HOME}/.zprezto/modules/prompt/init.zsh" ]]; then
-    # If there is an prezto installation, we assume that powerlevel9k is linked there.
-    POWERLEVEL9K_INSTALLATION_PATH="${ZDOTDIR:-$HOME}/.zprezto/modules/prompt/functions/prompt_powerlevel9k_setup"
+if [[ -n "$POWERLEVEL9K_INSTALLATION_DIR" ]]; then
+  p9k_directory=${POWERLEVEL9K_INSTALLATION_DIR:A}
+else
+  if [[ "${(%):-%N}" == '(eval)' ]]; then
+    if [[ "$0" == '-antigen-load' ]] && [[ -r "${PWD}/powerlevel9k.zsh-theme" ]]; then
+      # Antigen uses eval to load things so it can change the plugin (!!)
+      # https://github.com/zsh-users/antigen/issues/581
+      p9k_directory=$PWD
+    else
+      print -P "%F{red}You must set POWERLEVEL9K_INSTALLATION_DIR work from within an (eval).%f"
+      return 1
+    fi
+  else
+    # Get the path to file this code is executing in; then
+    # get the absolute path and strip the filename.
+    # See https://stackoverflow.com/a/28336473/108857
+    p9k_directory=${${(%):-%x}:A:h}
   fi
-else
-  # Last resort: Set installation path is script path
-  POWERLEVEL9K_INSTALLATION_PATH="$0"
 fi
-
-# Resolve the installation path
-if [[ -L "$POWERLEVEL9K_INSTALLATION_PATH" ]]; then
-  # If this theme is sourced as a symlink, we need to locate the real URL
-  filename="${POWERLEVEL9K_INSTALLATION_PATH:A}"
-elif [[ -d "$POWERLEVEL9K_INSTALLATION_PATH" ]]; then
-  # Directory
-  filename="${POWERLEVEL9K_INSTALLATION_PATH}/powerlevel9k.zsh-theme"
-elif [[ -f "$POWERLEVEL9K_INSTALLATION_PATH" ]]; then
-  # Script is a file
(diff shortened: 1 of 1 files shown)


dritter:
Here are my testcases:

1. Does this work on all ZSH frameworks in the [VM](https://github.com/bhilburn/powerlevel9k/tree/next/test-vm) (especially `prezto`, as this loads themes differently).
2. Does this work with [BSD](https://github.com/bhilburn/powerlevel9k/tree/next/test-bsd-vm)?
3. Does this work with spaces in the path?
4. Does this work with symlinks?
5. Does this work in old ZSH versions? <- We'll should define a minimum version..

Thanks for testing! 👍

docwhat:
- [ ] Does this work on all ZSH frameworks in the VM (especially prezto, as this loads themes differently).
- [ ] Does this work with BSD? (Does macOS count?)
- [ ] Does this work with spaces in the path?
- [ ] Does this work with symlinks?
- [ ] Does this work in old ZSH version X.Y.Z (need number)?

docwhat:
Paired with @dritter: Antigen uses `eval` (WTF?!?!?) to run theme plugins.  This will _not_ work with my code.

I'll re-enable the ability to manually override the installation directory.  I should also be able to halt and offer a message if using `eval` to make a nice UX.

docwhat:
I filed an issue with Antigen about the `eval` use: zsh-users/antigen#581

docwhat:
@dritter:

I did all the tests from the [above comment](https://github.com/bhilburn/powerlevel9k/pull/562#issuecomment-314192992) except the BSD one because I couldn't get the VM to start (it went in a weird loop).

However, it works on Darwin (macOS) and it makes use of _no_ external programs; everything is built-in to zsh.

Any other testing you want?

docwhat:
I cleaned up my commit and re-pushed it, @dritter.

Also note that I pushed up some more changes for the docker stuff.  I'll see about turning it into a PR.  I have a few more cleanups and I want to see about adding more of the frameworks you have in your personal fork.

docwhat:
It looks like the Antigen hack won't be needed in v2.1.0! W00T! zsh-users/antigen#581

dritter:
Awesome @docwhat ! Good job!

dritter:
Just tested it under BSD. It works 🎉  Even with a nasty relative symlink.
Thanks @docwhat :)

bhilburn:
@docwhat - This is really great work. As someone that is driven somewhat mad by complex code, this PR's diff makes me really happy. Thanks so much for taking the time to put this together, take feedback, test it, and revise as needed.

Special thanks to @dritter, too, for making himself available to team with @docwhat! Nice teamwork `=)`

I'm ready to merge this as soon as possible. It sounds like if we merge before the 2.1.0 release, though, we're going to break things for all of our Antigen users. I tried to figure out from the Antigen project when 2.1.0 will release, and I'm really confused.

Their tracking board talks about v2.0.2 and v2.3.0: https://trello.com/b/P0xrGgfT/antigen
Their Milestones only show v2.3.0, scheduled for this Saturday: https://github.com/zsh-users/antigen/milestones
Their `develop` branch is building v2.2.1: https://github.com/zsh-users/antigen

Does anyone else have a clearer picture on this?

docwhat:
> I'm ready to merge this as soon as possible. It sounds like if we merge before the 2.1.0 release, though, we're going to break things for all of our Antigen users. I tried to figure out from the Antigen project when 2.1.0 will release, and I'm really confused.

Nope.  It has special case code for < v2.1.0 and v2.1.0 works using the same code paths as everyone else.  It should be safe to merge now!

I'm putting together a docker test environment thingy ([`test-in-docker`](https://github.com/docwhat/powerlevel9k/blob/test-docker/test-in-docker)) that I'll submit shortly.
It's been letting me test _lots_ of frameworks.

I have a few more to pull from @dritter's async branch (he did it in a VM which is a little more cumbersome, but works too).

dritter:
> I'm ready to merge this as soon as possible. It sounds like if we merge before the 2.1.0 release, though, we're going to break things for all of our Antigen users. I tried to figure out from the Antigen project when 2.1.0 will release, and I'm really confused.

The trick is done [here](https://github.com/bhilburn/powerlevel9k/pull/562/files#diff-ea5a405f0331722fc10a832209ffe20bR28). We now check if `"${(%):-%N}"` holds `(eval)` instead of the path. If so, and if `$0` is `-antigen-load`, we try to use `${url}` (set by antigen)..

powerlevel9k.zsh-theme
```diff
@@ -17,74 +17,57 @@
...
-  print -P "Please specify by setting %F{blue}POWERLEVEL9K_INSTALLATION_PATH%f (full path incl. file name) at the very beginning of your ~/.zshrc"
-  return 1
-else
-  print -P "%F{red}Script location could not be found! Maybe your %F{blue}POWERLEVEL9K_INSTALLATION_PATH%F{red} is not correct?%f"
-  return 1
+  if [[ "${(%):-%N}" == '(eval)' ]]; then
+    if [[ "$0" == '-antigen-load' ]] && [[ -r "${PWD}/powerlevel9k.zsh-theme" ]]; then
+      # Antigen uses eval to load things so it can change the plugin (!!)
+      # https://github.com/zsh-users/antigen/issues/581
+      p9k_directory=${url}
```
dritter: Isn't that URL a real URL depending on how the user installed the theme?
Like if he does an `antigen install https://github.com/bhilburn/powerlevel9k` this url would be put in that variable, no? [This](https://github.com/zsh-users/antigen/blob/1d212d149d039cc8d7fdf90c016be6596b0d2f6b/src/lib/parse-args.zsh#L13) looks like.

docwhat: URL may or may not be a real URL.  It's a path for local installs.

I don't think powerlevel9k would work via http(s).

Also, I thought I changed that to `PWD`?  Let me fix that.

docwhat: Fixed.

FYI: `url` and `PWD` should be equal for a local install.

Do you want me to test it with http(s)?

dritter: Well. I think the local installation is the least common way to install p9k with antigen. Our [installation instructions for antigen](https://github.com/bhilburn/powerlevel9k/wiki/Install-Instructions#option-4-install-for-antigen) say to install it via `antigen theme bhilburn/powerlevel9k powerlevel9k`. It may be that antigen takes that argument and builds an url out of it.
But it is also possible to provide an url: `antigen theme https://gist.github.com/3750104.git agnoster` (taken from [here](https://github.com/zsh-users/antigen/wiki/Commands#antigen-theme)). So I think it would be better to test it, yes.

docwhat: I'll give it a try then.

bhilburn:
@docwhat @dritter - Thanks for the quick responses! And perfect, makes sense.

This is a really great PR, and I'm especially happy about how it was done. Thanks, all! :)

docwhat:
I don't remember where (or who asked) about trying this commit with the suggested syntax from the wiki; but it works.  I swapped out the `docker/antigen/zshrc` script with this:

``` zsh
source ~/antigen/antigen.zsh

antigen theme bhilburn/powerlevel9k@next powerlevel9k
antigen apply
```

It works fine. :-)

PR #566, opened 2017-07-10
Unset variables that interfere with named directories

cdlm:
If left defined, `~script_location` appears in the prompt when navigating to
powerlevel9k's checkout directory (due to `setopt autonamedirs`, it's a named
directory)

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -86,6 +86,10 @@ source $script_location/functions/colors.zsh
 
 source $script_location/functions/vcs.zsh
 
+# variables now unneeded: forget them
+# they could interfere with named directories
+unset filename script_location
+
 ################################################################
 # Color Scheme
 ################################################################


docwhat:
#562 also does this and more.

dritter:
@docwhat yes. You are right. Sorry, I was about to review your PR, but that requires some more testing. ;) Will do soon.

docwhat:
I've been using since I posted it.

If you give me some test cases I can try them out for you.

dritter:
@docwhat That would be awesome! I'll write them down in #562

bhilburn:
@cdlm - Thanks so much for the PR, and welcome to P9k! `=)`

@docwhat @dritter - Since we are in the process of merging #562, I want to keep things clean. You mention above that #562 also fixes this problem. Does that mean that this PR isn't necessary as it is a subset of #562?

docwhat:
Correct!

cdlm:
Yeah somehow I missed #562 before sending this PR, but it includes the same fix…

bhilburn:
No problem, @cdlm! I just merged #562, so we should be good, here.

Note that #562 was merged into `next`, not `master`, so you'll need to be on our `next` branch to see the change.

@cdlm - This was a good PR, and I look forward to your next contribution =)

Issue #565, opened 2017-07-09
Home folder shows as …/~ rather than ~ initially


unikitty37:
When I start a new shell, the home directory shows in the prompt as "…/\~" rather than just "\~". If I cd to another directory and then back to home, it shows correctly as "~".

![screen shot 2017-07-09 at 14 02 35](https://user-images.githubusercontent.com/139183/27994151-954c125a-64af-11e7-9597-33b6dd460184.png)

Relevant sections of my .zshrc:
```zsh
export POWERLEVEL9K_SHORTEN_DIR_LENGTH=2
export POWERLEVEL9K_SHORTEN_STRATEGY="truncate_folders"
```

Interestingly, this is happening on macOS 10.12.5 with zsh 5.2, but not on Linux with zsh 5.3.1 or zsh 5.1.1. Apart from differing colours, I have the same options in the .zshrc on all three systems.

unikitty37:
When I start a new shell, the home directory shows in the prompt as "…/~" rather than just "~". If I cd to another directory and then back to home, it shows correctly as "~".

![screen shot 2017-07-09 at 14 02 35](https://user-images.githubusercontent.com/139183/27994151-954c125a-64af-11e7-9597-33b6dd460184.png)

Relevant sections of my .zshrc:
```zsh
export POWERLEVEL9K_SHORTEN_DIR_LENGTH=2
export POWERLEVEL9K_SHORTEN_STRATEGY="truncate_folders"
```

Interestingly, this is happening on macOS 10.12.5 with zsh 5.2, but not on Linux with zsh 5.3.1 or zsh 5.1.1. Apart from differing colours, I have the same options in the .zshrc on all three systems.

bhilburn:
@unikitty37 - Hah, yup, that's bizarre. I'm using ZSH 5.2 on Linux and don't see this.

It's likely that I'll be a macOS user in a couple of weeks, but right now I don't have a mac available to me. It's really obnoxious that we can't easily set up test VMS (e.g., Vagrant, Docker) for macOS...

@dritter - Would you mind replicating @unikitty37's settings and testing to see if you can reproduce this?

unikitty37:
I'll leave all the oh-my-zsh stuff here in case it helps:

```zsh
export ZSH=$HOME/.oh-my-zsh
export ZSH_THEME='powerlevel9k/powerlevel9k'

export POWERLEVEL9K_MODE='nerdfont-complete'
[...]

PR #568, opened 2017-07-12
Fix empty delimiter

### LINKED ISSUE #558: truncate_from_right not working on BSD?
When using the truncate_from_right strategy on a FreeBSD system, sed returns an error and there is no dir string displayed.
`sed: 1: "s@(([^/]{1})([^/]{0}))[ ...": RE error: empty (sub)expression`

zshrc is attached. The problem goes away when using truncate_middle... 

[zshrc.txt](https://github.com/bhilburn/powerlevel9k/files/1111889/zshrc.txt)

dritter:
Even if the delimiter is empty, the min length should be 1. This works on OSX and BSD.

Fixes #558

### DIFF
--- a/test/segments/dir.spec
+++ b/test/segments/dir.spec
@@ -72,6 +72,26 @@ function testTruncationFromRightWorks() {
   unset POWERLEVEL9K_SHORTEN_STRATEGY
 }
 
+function testTruncationFromRightWithEmptyDelimiter() {
+  POWERLEVEL9K_SHORTEN_DIR_LENGTH=2
+  POWERLEVEL9K_SHORTEN_DELIMITER=""
+  POWERLEVEL9K_SHORTEN_STRATEGY='truncate_from_right'
+
+  FOLDER=/tmp/powerlevel9k-test/1/12/123/1234/12345/123456/1234567/12345678/123456789
+  mkdir -p $FOLDER
+  cd $FOLDER
+
+  assertEquals "%K{blue} %F{black}/tmp/po/1/12/123/12/12/12/12/12/123456789 %k%F{blue}%f " "$(build_left_prompt)"
+
+  cd -
+  rm -fr /tmp/powerlevel9k-test
+
+  unset FOLDER
+  unset POWERLEVEL9K_SHORTEN_DIR_LENGTH
+  unset POWERLEVEL9K_SHORTEN_DELIMITER
+  unset POWERLEVEL9K_SHORTEN_STRATEGY
+}
+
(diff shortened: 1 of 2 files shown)


test/segments/dir.spec
```diff
@@ -72,6 +72,26 @@ function testTruncationFromRightWorks() {
   unset POWERLEVEL9K_SHORTEN_STRATEGY
 }
 
+function testTruncationFromRightWithEmptyDelimiter() {
+  POWERLEVEL9K_SHORTEN_DIR_LENGTH=2
```
docwhat: Use `local` for the `POWERLEVEL9K_*` variables and `FOLDER`; then you don't need `unset` at the end.

test/segments/dir.spec
```diff
@@ -72,6 +72,26 @@ function testTruncationFromRightWorks() {
...
+  POWERLEVEL9K_SHORTEN_DELIMITER=""
+  POWERLEVEL9K_SHORTEN_STRATEGY='truncate_from_right'
+
+  FOLDER=/tmp/powerlevel9k-test/1/12/123/1234/12345/123456/1234567/12345678/123456789
+  mkdir -p $FOLDER
+  cd $FOLDER
+
+  assertEquals "%K{blue} %F{black}/tmp/po/1/12/123/12/12/12/12/12/123456789 %k%F{blue}%f " "$(build_left_prompt)"
+
+  cd -
```
docwhat: Can we move the `cd -` (return to original directory) into a teardown or even the setup function?

awensaunders:
Works on my FreeBSD system currently.

test/segments/dir.spec
bhilburn: Hm. I wonder if moving this to the setup function would actually help fix #565?

bhilburn:
Thanks for the quick bug fix, @dritter, and thanks for the review, @docwhat. 

@awensaunders - Happy to hear it! Once we get the review comments sorted out, we'll proceed :)

test/segments/dir.spec
dritter: Yep. I agree. Don't know why I didn't do this in the first place... To reduce merge conflicts, I would want to start that in the async branch.

test/segments/dir.spec
dritter: This is something I have already done in the `async` branch. To reduce merge conflicts, I leave it like this here.

bhilburn:
Thanks for the review responses, @dritter! All looks well `=)`

PR #569, opened 2017-07-12
Added tests and replaced a sed call

docwhat:
This was partially an experiment on my part; I wanted to see
what would be involved in (safely) replacing some sed calls.

The rational is that forks out to sed are slower than pure ZSH.

It isn't a huge difference on Linux, but each fork can take a while even in
the latest Windows.

...and as I said, it was an experiment.  I'm interested in optimizing some things.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -811,7 +811,7 @@ prompt_dir() {
   fi
 
   if [[ "${POWERLEVEL9K_HOME_FOLDER_ABBREVIATION}" != "~" ]]; then
-    current_path="$( echo "${current_path}" | sed "s/^~/${POWERLEVEL9K_HOME_FOLDER_ABBREVIATION}/")"
+    current_path=${current_path/#\~/${POWERLEVEL9K_HOME_FOLDER_ABBREVIATION}}
   fi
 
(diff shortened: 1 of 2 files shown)


bhilburn:
I really like this idea. `sed` is great and all, but I do like to avoid calls to external programs when possible. The only downside I can think of is that `sed` is more readable than ZSH regex, in my opinion. We can mitigate this with comments where necessary, though.

I'm all for moving ahead with this. @dritter, what are your thoughts?

docwhat:
I'm actually more confused by `sed` when it does more than `s/a/b/g`.  I just never bothered to learn it since `perl` was so available and much more powerful and readable.

The whole `zshexpn(1)` man page should be required reading... then re-read it.  It is horrible as far as documentation but the features it has are incredible and once you know how to read modifiers, etc. it all becomes much clearer.

Read the `test-in-docker` code for (I hope) clear examples of using a bunch of features like that.

Feel free to assign issues or make suggestions for me to add comments to cryptic ZSH code.

bhilburn:
@docwhat - Sounds good to me. And, going forward, I'm happy to merge any such changes that remove external calls to `sed` and simplifies the code =)

Thanks again, by the way, for your contributions and technical guidance 👍

PR #570, opened 2017-07-13
test-in-docker: show known issues

docwhat:
Certain frameworks don't work with the containers for reasons such as ZSH version or git versions.


bhilburn:
This is pretty slick. Nicely done =)

Thanks for continuing to improve the Docker work, @docwhat!

PR #573, opened 2017-07-15
Feature/status cross ok option set

AdrienHorgnies:
The purpose of this PR is to be able to have the error code but no icon when there is no error.
We strictly win one more possibility : 

|        | ERROR CODE           | CROSS ICON  |
| ------------- |:-------------:|:-----:|
| OK ICON      | present | present |
| HIDDEN SEGMENT     | **new** |   present |

I've made the new option set such as his default values doesn't affect old option set, no matter what the old option set values are. To do so, I've used logical operator properties : 
`true && condition` == `condition` where `true` is the default value of the new option set and `condition` is the old option set value.

So, to use the new option set, the old option set must be at his default values and to use the old option set, the new option set must be at his default values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -505,8 +505,8 @@ This segment shows the return code of the last command.
 
 | Variable | Default Value | Description |
 |----------|---------------|-------------|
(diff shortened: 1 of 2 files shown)


bhilburn:
@AdrienHorgnies - Great! Thanks so much for taking the suggestions & feedback from #560, incorporating them into your work, and then filing this PR. This is excellent. You did an excellent job of preserving backwards compatibility while documenting the new options, as well.

Thanks, and welcome to P9k! `=)`

AdrienHorgnies:
@bhilburn I will print and frame this message.

You're very much welcome, I love using your work.

PR #580, opened 2017-07-26
Fixed Vi mode indicator issue

sei40kr:
**Changes**
* Fixed an issue that Vi mode won't appear before the user changes the mode. (This change shows INSERT when failed to get current Vi mode)
* ~Added nice-exit-code plugin. This can be activated with the flag `POWERLEVEL9K_STATUS_NICE_EXIT_CODE` and shows a signal name of the last status code.~

PR #581, opened 2017-07-27
status: show signal name

docwhat:
This makes signal exit status easy to understand.  Instead of just
showing exit code "137", you now see "KILL(-9)".

Based on #580 by @sei40kr (thanks!)


docwhat:
I can add a flag if you want, but this is so sexy, I don't think it needs it. 😎

bhilburn:
@docwhat - Fantastic!

I would like for this new mode to be the default, but I'd still like to have the old mode available for those that want it. Would you mind adding a flag to show the true error code, but defaulting it to `false`?

docwhat:
Done, @bhilburn

bhilburn:
Nicely implemented, @docwhat, and this is a really fantastic idea, @sei40kr! This is a really neat feature `=)`

PR #582, opened 2017-07-27
Fix for #575 - RVM segment version reporting

rjorgenson:
This is the solution I was able to come up with =]

PR #583, opened 2017-07-27
rvm: handle -head, etc. more gracefully.

### LINKED ISSUE #575: RVM segment doesn't parse for -head version of ruby
With the recent fix to the RVM segment it now only prints the last column in the awk separation, for the -head version of ruby(and possibly others) this just results in the version being reported as 'head' which isn't terribly useful =] How does the below look? This should split out the path reported and print just the ruby version which we can then parse out as we wish. The below just does what the previous segment did, but we could expand on it if we want to display more/less information?

```shell
echo $MY_RUBY_HOME | awk -F'/' '{print $NF}' | awk -F'-' '{print $2}'
```

docwhat:
Closes #575

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1070,13 +1070,10 @@ prompt_rspec_stats() {
 
 # Ruby Version Manager information
 prompt_rvm() {
-  local gemset=$(echo $GEM_HOME | awk -F'@' '{print $2}')
-  [ "$gemset" != "" ] && gemset="@$gemset"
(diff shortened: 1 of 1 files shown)


rjorgenson:
I like the way you handled this better =]

bhilburn:
@docwhat @rjorgenson - Thanks for working together on this in #575, and then in #582 and here! I think your discussion yielded the best result, and this looks great 👍 

I had wanted to put this on `master` since it's a bit of a bugfix, but I'm going to try and push out a release, shortly, so I'll just go ahead and merge for now.

Thanks again!

PR #584, opened 2017-07-27
Update OK_ICON weight to match FAIL_ICON

### LINKED ISSUE #576: Status icons use inconsistent font bolding
I have following the following setup:

```
POWERLEVEL9K_STATUS_VERBOSE=false
POWERLEVEL9K_STATUS_OK_IN_NON_VERBOSE=true
```

When the exit status is OK, it returns a normal check mark (✓); when there's an error, it returns a heavy ballot X (✘). It'd be better if both were normal weight ones (✓, ✗) or heavy ones (✔, ✘).

giogziro95:
Fixes #576.


bhilburn:
Thanks, @giogziro95! Even the little things matter, and this is great.

Welcome to P9k `=)`

PR #592, opened 2017-08-08
Implemented functionality from dir_writable to dir

### LINKED ISSUE #588: Combine prompt elements "dir" and "dir_writable"
Would it make sense to combine the two prompt elements "dir" and "dir_writable", so that a read-only directory would change the color of the path (to red)?

tippl:
When variable POWERLEVEL9K_DIR_SHOW_WRITABLE is set to true, dir prompt
will have new state NOT_WRITABLE.

This state will be set when user doesn't have permission to write in
current working directory as in prompt dir_writable.

When mode that allows icons is set, LOCK_ICON is used.

Closes #588.

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,9 +402,11 @@ You can also customize the leading tilde character when you are in `$HOME` using
 # Double quotes are important here!
 POWERLEVEL9K_HOME_FOLDER_ABBREVIATION="%F{red} $(print_icon 'HOME_ICON') %F{black}"
 ```
-If you want to see, when you do not have write permissions on the current folder, you can set `POWERLEVEL9K_DIR_SHOW_WRITABLE=true`.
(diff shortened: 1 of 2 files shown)


bhilburn:
Nice work, @tippl! I'm really excited to see someone take up the torch from #588 :)

A few code review comments:

1. We need to make sure that `POWERLEVEL9K_DIR_SHOW_WRITABLE` is set to a default value. Check out some of the other segments, and you can see how we use a P9k utility function `set_default` to do that before the function.
2. Be sure to document this change in the README! We want users to know about this cool new feature you are adding =)
3. Since this is a new feature, it actually needs to get merged into the `next` branch rather than master (https://github.com/bhilburn/powerlevel9k/wiki/Developer's-Guide). Would you mind rebasing your changes onto `next` and then re-submitting the PR? Let us know if you need help with this!

I'm really excited about getting this in. Thanks for taking it on! =)

tippl:
@bhilburn This should address all your comments. I was thinking of adding the lock icon to the table at the start of the `dir` section, but i had no idea on how to upload an image to github.

bhilburn:
Excellent work, @tippl! Nice work incorporating the review feedback and updating the PR. This is a great addition to the project.

Welcome to P9k as a contributor `=)`

PR #594, opened 2017-08-10
Add back the prompt_sp option for zsh >= 5.4.1

belak:
In 5.4.1, this option was reset between prompts, so to retain the
previous default behavior, this should be added back.

This was the change which caused it to be reset: https://github.com/zsh-users/zsh/commit/43e55a9bcd2c90124a751f2597d2f33cb6e3c042#diff-bb10d67e7a8561b66a53a805f3c77a40R233

This is the documentation on the option itself: http://zsh.sourceforge.net/Doc/Release/Options.html#Prompting

Note that this makes it a little harder to unset this option (as it needs to be done after setting the prompt) but that seems fine, as all the other prompt options are already this way.


bhilburn:
Thanks, @belak! I'm glad you are staying on top of the relevant changes in the ZSH upstream `=)`

PR #596, opened 2017-08-11
Add TRUNCATE_TO_UNIQUE shortening dir strategy

### LINKED ISSUE #589: Truncate path elements to a unique minimum
While P9K offers different ways how to truncate the path elements, it doesn't check whether the resulting collapsed path is unique (copy&paste it to another terminal and TAB it to a unique directory). I didn't find a solution to this problem until [I asked at SO](https://stackoverflow.com/questions/45263799/collapse-directories-in-zsh-prompt-in-a-unique-way) and received an answer I consider a good one.

I put it already into my `.zshrc` (credits for the `spwd` function go to Vartan Christopher Simonian at SO):

    spwd() {
      paths=(${(s:/:)PWD})

      cur_path='/'
      cur_short_path='/'
      for directory in ${paths[@]}
      do
        cur_dir=''
        for (( i=0; i<${#directory}; i++ )); do
          cur_dir+="${directory:$i:1}"
          matching=("$cur_path"/"$cur_dir"*/)
          if [[ ${#matching[@]} -eq 1 ]]; then
            break
          fi
        done
        cur_short_path+="$cur_dir/"
        cur_path+="$directory/"
      done

      printf %q "${cur_short_path: : -1}"
      echo
    }

    POWERLEVEL9K_CUSTOM_SDIR="spwd"
    POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(status context custom_sdir dir_writable)`

Maybe this is an idea for another `POWERLEVEL9K_SHORTEN_STRATEGY`? It doesn't substitute the $HOME directory, but that shouldn't be a problem to implement.

eumiro:
Fixes #589


bhilburn:
Nice work, @eumiro! I made just a couple of small changes, including a link to the SO post where the logic came from for the truncation strategy. I also changed the target branch to `next`, which is where new features get merged into; we try to keep `master` as stable as possible, which means keeping churn low `=)`

Thanks so much for taking the idea from #589, and not only implementing it, but documenting it *and writing a test case for it*. Excellent work, and welcome to P9k!

eumiro:
Thank you for helping me to get this feature into my favorite zsh extension and for explaining me how the project works. Keep up the great work!

PR #599, opened 2017-08-12
similar to PR #472, fixing for awesome font mappings

pfrybar:
This is basically https://github.com/bhilburn/powerlevel9k/pull/472 but with the requested changes. I spent some time today getting this working for me and thought I would share my results. This creates a new icon set called `awesome-mapped-fontconfig` so it won't break for any existing users. This relies on the user sourcing the font mapping files in their shell startup scripts, as recommended by: https://github.com/gabrielelana/awesome-terminal-fonts#font-maps. It does a simple check to ensure the fonts maps were sourced, and logs a warning if they weren't.

### DIFF
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -152,6 +152,77 @@ case $POWERLEVEL9K_MODE in
       SSH_ICON                       '(ssh)'
     )
   ;;
+  'awesome-mapped-fontconfig')
+    # mapped fontconfig with awesome-font required! See
+    # https://github.com/gabrielelana/awesome-terminal-fonts
+    # don't forget to source the font maps in your startup script
+    # Set the right locale to protect special characters
+    local LC_ALL="" LC_CTYPE="en_US.UTF-8"
+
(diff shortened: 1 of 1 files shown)


bhilburn:
Nice work, @pfrybar! This is exactly what we needed. I appreciate you finishing up the work from #472 and then testing it out.

I know this took a while - many months since @V1rgul first put together the commits that started the effort. Fonts are painful, but I think we ended up with the best solution, here.

I really want to thank @V1rgul for pushing the idea and providing guidance, @gabrielelana for updating the AwesomeFont project docs based on our needs, and @dritter for the review feedback in the previous two PRs. 

Great PR, @pfrybar - especially for a first contribution! Welcome to P9k!

Thanks, all! And, obviously, please file bugs / new PRs if you see issues `=)`

bhilburn:
(By the way, I'll update the wiki to tell users to use `awesome-mapped-fontconfig` once we make our next release, which I would like to do relatively quickly behind v0.6.4)

PR #602, opened 2017-08-14
Updated braces for hyperlink

varundey:


bhilburn:
@varundey - Thanks for the quick fix! Even little things like this matter =)

varundey:
@bhilburn glad to be of help. Keep up the good work with P9K ;)

PR #603, opened 2017-08-17
add ability to force rbenv being shown

sjoeboo:
I set a global rbenv, but do always want the prompt segment to be shown, so this adds a simple `POWERLEVEL9K_RBENV_ALWAYS` to force the segment even if its the global version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,6 +402,10 @@ It figures out the version being used by taking the output of the `rbenv version
 * If `rbenv` is not in $PATH, nothing will be shown.
 * If the current Ruby version is the same as the global Ruby version, nothing will be shown.
 
+Variable | Default Value | Description |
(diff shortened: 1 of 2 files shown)


bhilburn:
@sjoeboo  - Thanks so much for the contribution! A few review comments:

1) We need to make sure that 'POWERLEVEL9K_RBENV_ALWAYS` is set to a default value. Check out some of the other segments, and you can see how we use a P9k utility function `set_default` to do that before the function.
2) Since this is a new feature, it actually needs to get merged into the `next` branch rather than `master` (https://github.com/bhilburn/powerlevel9k/wiki/Developer's-Guide). Would you mind re-targeting this PR to `next` once you have added the `set_default` call?

Thanks so much! Looking forward to getting this merged.

PR #608, opened 2017-08-27
Fix average selection in load segment.

tippl:
Removed duplicate code selecting which load average to use.
Fixed load average selection being in OSX/BSD part of if, making it not work
on non OSX/BSD systems.
Optimized actually getting the load average.


bhilburn:
This is, indeed, much cleaner. Thanks for taking care of this, @tippl. Much appreciated `=)`

PR #616, opened 2017-09-05
Changed "less than" to "fewer than".

trashbat:


bhilburn:
@trashbat - Well done 👍 ;) #WordCrimes

Thanks for the fix, and welcome to #P9k!

Issue #617, opened 2017-09-08
Github status icon and branch name missing space


runmad:
<img width="233" alt="screen shot 2017-09-08 at 1 22 30 pm" src="https://user-images.githubusercontent.com/708741/30229950-ddbce65e-9498-11e7-87ef-0fb0e5f2f0f7.png">

See screenshot. I have the same setup on another Mac, using iTerm2, however the spacing between the arrow icon and the branch name isn't there on one computer. Looking through docs, I see how to add the different segments but not customize their appearance specifically.

runmad:
Update: I just pulled the latest of this repo on the other machine. Before I pulled it looked like this:

<img width="285" alt="screen shot 2017-09-08 at 5 39 09 pm" src="https://user-images.githubusercontent.com/708741/30235340-e5046bc0-94bc-11e7-877c-1c0375d01645.png">

After I pulled:

<img width="289" alt="screen shot 2017-09-08 at 5 41 23 pm" src="https://user-images.githubusercontent.com/708741/30235341-f2472d68-94bc-11e7-8fe4-d87f3fbfa84b.png">

I kind of like that spacing in the previous version.

SConaway:
Agreed.

bhilburn:
Closing this so that we can focus the conversation in the thread over in #618 `=)`

bhilburn:
@SConaway @runmad - So, for now, you can simply overwrite the VCS_BRANCH_ICON string to add a space. We're still discussing #618.

`POWERLEVEL9K_VCS_BRANCH_ICON=$'\uF126 '

Note the space at the end, there? That'll do exactly what you need with zero code changes.

bhilburn:
@SConaway @runmad - Okay, I just reverted that change, actually, and added the space back. You should be good if you pull master.

runmad:
Thanks, @bhilburn

PR #618, opened 2017-09-09
Add space between VSC_BRANCH_ICON and branch name.

diraol:
FIX bhilburn/powerlevel9k#617

### DIFF
--- a/functions/vcs.zsh
+++ b/functions/vcs.zsh
@@ -51,7 +51,7 @@ function +vi-git-remotebranch() {
     remote=${$(git rev-parse --verify HEAD@{upstream} --symbolic-full-name 2>/dev/null)/refs\/(remotes|heads)\/}
     branch_name=$(git symbolic-ref --short HEAD 2>/dev/null)
 
-    hook_com[branch]="$(print_icon 'VCS_BRANCH_ICON')${hook_com[branch]}"
+    hook_com[branch]="$(print_icon 'VCS_BRANCH_ICON') ${hook_com[branch]}"
     # Always show the remote
     #if [[ -n ${remote} ]] ; then
     # Only show the remote if it differs from the local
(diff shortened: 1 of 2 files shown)


bhilburn:
@diraol - Thanks so much for submitting this PR! I apologize it's taken me so long to respond - the last month has been a bit crazy for me.

@diraol @runmad - This issue has actually come up a few times. When I first released P9k, we actually did *not* have a space, then I added it a little bit ago and it remained for a few releases before enough people complained that I pulled it back out again.

The primary issue with putting the space *in* the code is that it makes it difficult for those that don't want it to remove it. Going the opposite direction, though, if there is *not* a space in the code, it's very easy to add one by modifying the `VCS_BRANCH_ICON` text by appending a space.

If you look at your changeset, for example, with the space built-in, there, there is no way for someone to remove it other than modifying the P9k code directly.

I personally prefer having a space, to be honest, and I would like to settle on a solution that works well for everyone.

@dritter @rjorgenson @docwhat - Curious about your thoughts on this topic.

docwhat:
I think leaving the space out is best. Unicode 9 solves a lot of problems with characters that need double space.

diraol:
@bhilburn what about adding a new variable just like `POWERLEVEL9K_SHORTEN_DELIMITER`?

Something like `POWERLEVEL9K_VCS_BRANCH_ICON_RSPACE`. It can be a 'boolean' variable or just a string, which defaults to `""` (empty string) and we can set it to `" "`.

docwhat:
I think supporting Unicode 9 is better. It automatically makes wide glyphs take two spaces in mono spaced fonts.

diraol:
@docwhat I do not have any idea what is the necessary effort to support Unicode 9. Any tips on that?

Also, why not supporting the latest release of Unicode (10) instead of (9)?

docwhat:
https://github.com/jquast/wcwidth has some info. As well as iterm2 for Mac. 

Basically: emoji can be double width in 9. This makes things interesting if you need to know how wide text is.

bhilburn:
Hm, so it's not really clear to me what actually needs to be done in P9k to support UTF-9.

@diraol - So I just put the space back. I actually started receiving e-mails from folks about this, and they raised the good point that, previously,  the default was to have the space. I've returned it back to the old default, and anyone who doesn't want the space can remove it by resetting the icon.

PR #638, opened 2017-10-01
Added Dropbox segment

TimothyEarley:
Adds a simple dropbox segment that shows the dropbox symbol when in a dropbox directory. Also shows the status when syncing. Requires dropbox-cli to be installed.

I wrote it because I sometimes start editing in a directory without realising I'm in a dropbox folder. So while for me it is quite usefull, I don't know if others will need it.


bhilburn:
@TimothyEarley - Thanks so much for submitting this new segment, and documenting it! I really appreciate it.

I made some minor changes, removing the Dropbox icon for non-Awesome fonts. Otherwise, everything looks good.

Nice PR, and welcome to P9k! `=)`

PR #640, opened 2017-10-05
Corrected awesome-fontconfig lock icon

simonvpe:
Simple correction (see comment)


bhilburn:
Yup! Thanks for the catch and the fix, @simonvpe!

Welcome to P9k, by the way `=)`

Issue #632, opened 2017-09-23
Can "Rust" be removed from `rust_version` prompt?


chhschou:
The `rust_version` prompt has a `Rust` at the front. Since it already has a rust icon at the back, the `Rust` word in the prompt is redundant and takes up valuable real estate. 

![image](https://user-images.githubusercontent.com/429067/30772491-7a9a2522-a0b0-11e7-8bd5-6e92f30b3f63.png)

Happy to contribute a PR.

bhilburn:
@chhschou - Fair point! A PR would be great =)

I think it would be good to move the Rust icon to be before the version. That way, for folks who aren't using a font with an expanded icon set (and thus don't have the Rust icon), the 'icon' can be set to `Rust` and they will still get the same effect.

chhschou:
Ahh, good call on supporting folks without the extended icons. I'll put something in for this.

chhschou:
I just did it, but moving the icon to the front feels out of place wrt to other prompts that have icons at the back. 
![image](https://user-images.githubusercontent.com/429067/30996497-c5516e1a-a51d-11e7-962b-67e03eab3dbf.png)

I still think removing word `Rust` form the front is a good idea, it is then consistent with other prompts. Should it be `(rust)` or `Rust` if no icon? Both styles exist in code today.

bhilburn:
Closing this to track in #641. I know that PR has been languishing a bit, but I haven't forgotten about it =)

PR #641, opened 2017-10-05
#632: Indicate `Rust` or `(rust)` for rust_version segment only if ic…

chhschou:
…on is unavailable

Hi @bhilburn, please let me know what you think.

### DIFF
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -77,7 +77,7 @@ case $POWERLEVEL9K_MODE in
       VCS_GIT_GITLAB_ICON            $'\uE20E '             #
       VCS_HG_ICON                    $'\uE1C3 '             # 
       VCS_SVN_ICON                   '(svn) '
-      RUST_ICON                      ''
+      RUST_ICON                      '(rust)'
       PYTHON_ICON                    $'\ue63c'             # 
       SWIFT_ICON                     ''
       GO_ICON                        ''
@@ -290,7 +290,7 @@ case $POWERLEVEL9K_MODE in
       VCS_GIT_GITLAB_ICON            ''
       VCS_HG_ICON                    ''
       VCS_SVN_ICON                   ''
-      RUST_ICON                      ''
+      RUST_ICON                      'Rust'
       PYTHON_ICON                    ''
       SWIFT_ICON                     'Swift'
       GO_ICON                        'Go'
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1095,9 +1095,9 @@ prompt_root_indicator() {
 prompt_rust_version() {
   local rust_version
   rust_version=$(rustc --version 2>&1 | grep -oe "^rustc\s*[^ ]*" | grep -o '[0-9.a-z\\\-]*$')
-
+  
   if [[ -n "$rust_version" ]]; then
-    "$1_prompt_segment" "$0" "$2" "208" "$DEFAULT_COLOR" "Rust $rust_version" 'RUST_ICON'
+    "$1_prompt_segment" "$0" "$2" "208" "$DEFAULT_COLOR" "$rust_version" 'RUST_ICON'
   fi
 }
 # RSpec test ratio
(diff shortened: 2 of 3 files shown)


bhilburn:
Thanks, @chhschou! In looking at the code, here, you're right - this needs to get fixed.

Your PR looks good, but in looking at the code, it looks like there is even more we could fix. So, check this out:

* `awesome-patched`: You are fixing this with with `(rust`) [here](https://github.com/bhilburn/powerlevel9k/pull/641/files#diff-b6c6e7997da6b6ad40e6e917acafa27cR80)
* `awesome-fontconfig`: Mapped to `E6A8` [here](https://github.com/bhilburn/powerlevel9k/pull/641/files#diff-b6c6e7997da6b6ad40e6e917acafa27cL151)
* `nerdfont-complete`: Mapped to `E7A8` [here](https://github.com/bhilburn/powerlevel9k/pull/641/files#diff-b6c6e7997da6b6ad40e6e917acafa27cL222)
* `powerline-patched`: You fix this with `Rust` [here](https://github.com/bhilburn/powerlevel9k/pull/641/files#diff-b6c6e7997da6b6ad40e6e917acafa27cR293)

I have a couple of questions:

1) Is there a reason you used `(rust)` in one place and `Rust` in the other?
2) For the most part, `awesome-patched` and `awesome-fontconfig` can use the same codepoints. Are we sure that these need to be different?
3) So, I pulled up the `awesome-fontconfig` font map, and guess what? `E6A8` isn't anywhere in there. [Check it out](https://github.com/gabrielelana/awesome-terminal-fonts/blob/master/fonts/fontawesome-regular.sh) -- can you comment, @gabrielelana?
4) It's also strange to me that the Nerdfonts codepoint is only just slightly different from the Awesome map, `E6A8` vs `E7A8`. -- can you comment, @ryanoasis?

chhschou:
@bhilburn I have a pretty basic understanding of `awesome-patched` and `powerline-patched`. I used both `(rust)` and `Rust` following the patterns surrounding each change. If there is a preference for one over the other I'm happy to comply. I did ask a question in #632 but it wasn't answered so I thought the choice wasn't important. Personally, I like `(rust)` better, the parentheses helps me see that the icon is unavailable.

gabrielelana:
@bhilburn 

> So, I pulled up the awesome-fontconfig font map, and guess what? E6A8 isn't anywhere in there. Check it out -- can you comment, @gabrielelana?

Because the rust icon is not part of the [fontawesome](http://fontawesome.io/icons/) icon font but is part of [devicons](http://vorillaz.github.io/devicons/#/dafont) icon font and in fact you can find the font code point in the [devicon font map](https://github.com/gabrielelana/awesome-terminal-fonts/blob/master/fonts/devicons-regular.sh)

> For the most part, awesome-patched and awesome-fontconfig can use the same codepoints. Are we sure that these need to be different?

for `awesome-fontconfig` always use the variables defined in the [font map](https://github.com/gabrielelana/awesome-terminal-fonts/blob/master/fonts/devicons-regular.sh), so use `"\u$CODEPOINT_OF_DEVICONS_RUST"` instead of `$'\uE6A8'` if you do that then I can update the icon fonts, possibly shift icons around to avoid clashes and still don't break anything for anyone

ryanoasis:
@bhilburn 

> It's also strange to me that the Nerdfonts codepoint is only just slightly different from the Awesome map, E6A8 vs E7A8

As @gabrielelana has said that glyph is from [devicons](https://vorillaz.github.io/devicons/#/cheat) not font awesome. Nerd Fonts has shifted the codepoints range due to conflicts with other glyph fonts: `E600-E6C5 → E700-E7C5`.

It occurs to me the shifting isn't clear on the [readme](https://github.com/ryanoasis/nerd-fonts#devicons) of Nerd Fonts currently :disappointed: 

Also we somewhat recently added variables for mapping the codepoints as well, e.g. [devicons map](https://github.com/ryanoasis/nerd-fonts/blob/master/bin/scripts/lib/i_dev.sh) though honestly I am not sure how `P9k` would incorporate them.

functions/icons.zsh
```diff
@@ -290,7 +290,7 @@ case $POWERLEVEL9K_MODE in
       VCS_GIT_GITLAB_ICON            ''
       VCS_HG_ICON                    ''
       VCS_SVN_ICON                   ''
-      RUST_ICON                      ''
+      RUST_ICON                      'Rust'
```
dritter: At some point we should agree on parenthesis or not.. But that is not the issue in this PR.

PR #643, opened 2017-10-07
Fix incomplete vpn segment name

terencode:
The segment function is called *prompt_vpn_ip()* thus the segment needs to be called "vpn_ip".


bhilburn:
So it does! Thanks so much for the fix, @terencode! It is much appreciated, and welcome to P9k! =)

terencode:
@bhilburn My pleasure to contribute to this awesome project ;)

PR #647, opened 2017-10-12
OSX bug fix for prompt_public_ip

sbutler2901:
The OSX date command does not have the -r option for getting the last modification time of a file. The stat command is being used instead.


bhilburn:
@sbutler2901 - Great catch, and thanks for the fix! Nice work, and welcome to P9k! `=)`

Issue #555, opened 2017-06-26
Kubecontext does not work with unauthenticated clusters.


derekbassett:
Kubecontext on next does not correctly pull the namespace when no AUTHINFO set for the cluster.

bhilburn:
Not being a kubernetes user myself, I'm not sure how to go about addressing this.

Any thoughts from our `kubecontext` experts? @jtaylor32

jtaylor32:
I will look into this. This makes sense -- @derekbassett if you could give me some context around what sort of Kubernetes config problems you're running into I'd love to try and replicate this so I can fix it!

bhilburn:
@derekbassett - Just a ping, here, on @jtaylor32's question, above!

derekbassett:
My situation has several of clusters not having SSL or AUTHINFO turned on. 
if I were to do a `kubectl config get-contexts` the command would print out something like this:

CURRENT | NAME |                   CLUSTER |                    AUTHINFO |                 NAMESPACE
----------|-------|----------------------|------------------------|----------------------
|| dev-int|                 dev-int|
*|         minikube|                minikube|                    minikube
||         sit-qa | sit-qa || mynamespace

My initial thought was to use 
`local k8s_namespace=$(kubectl config get-contexts --no-headers | grep '*' | awk '{print $NF}')`
As that would pick up the last value normally the namespace but in the case of minikube in the case of minikube or any other cluster with no namespace defined it would print out either AUTHINFO or CLUSTER not optimal but is what I've been using for a while in my own branch.

I'm not sure without a Kubernetes Client change it is even possible to consistently get the current namespace.

jtaylor32:
Ahh I see what you mean. 99.9% of the time I am always in namespaces other than default. I do get the same error you are saying but my prompt would look something like `minikube/` and it just doesn't have a namespace after. Maybe we can detect if namespace is not set to default to `default`. What do you think @derekbassett?

bhilburn:
@derekbassett - Just a ping on @jtaylor32's question, above!

derekbassett:
I investigated this a small bit. Without a change to Kubectl I don't see a way to reliably do this. When the namespace is default it just leaves the column empty.  If the authentication is unset then it will also leave the namespace unset. Most of the clusters I work with lack authentication.  Would you be up for generating a PR to Kubectl adding the ability to add namespace as a header as a command line flag?

jtaylor32:
[...]

PR #653, opened 2017-10-19
Fix current Kubernetes namespace parsing issues

captjt:
Added functionality to parse the current namespace regardless of config settings. This should address the issue [#555](https://github.com/bhilburn/powerlevel9k/issues/555).

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1410,20 +1410,22 @@ prompt_dir_writable() {
   fi
 }
 
-# Kubernetes Current Context
+# Kubernetes Current Context/Namespace
 prompt_kubecontext() {
   local kubectl_version="$(kubectl version --client 2>/dev/null)"
 
   if [[ -n "$kubectl_version" ]]; then
-    # Get the current Kubernetes config context's namespaece
-    local k8s_namespace=$(kubectl config get-contexts --no-headers | grep '*' | awk '{print $5}')
-    # Get the current Kuberenetes context
-    local k8s_context=$(kubectl config current-context)
-
-    if [[ -z "$k8s_namespace" ]]; then
-      k8s_namespace="default"
+    local cur_ctx=$(kubectl config view -o=jsonpath='{.current-context}')
+    cur_namespace="$(kubectl config view -o=jsonpath="{.contexts[?(@.name==\"${cur_ctx}\")].context.namespace}")"
+    # If the namespace comes back empty set it default.
+    if [[ -z "${cur_namespace}" ]]; then
+      cur_namespace="default"
     fi
-    "$1_prompt_segment" "$0" "$2" "magenta" "white" "$k8s_context/$k8s_namespace" "KUBERNETES_ICON"
+
+    # Get the current Kuberenetes context
+    local cur_ctx=$(kubectl config view -o=jsonpath='{.current-context}')
+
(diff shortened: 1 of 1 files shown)


powerlevel9k.zsh-theme
```diff
@@ -1410,20 +1410,22 @@ prompt_dir_writable() {
...
-# Kubernetes Current Context
+# Kubernetes Current Context/Namespace
 prompt_kubecontext() {
   local kubectl_version="$(kubectl version --client 2>/dev/null)"
 
   if [[ -n "$kubectl_version" ]]; then
-    # Get the current Kubernetes config context's namespaece
-    local k8s_namespace=$(kubectl config get-contexts --no-headers | grep '*' | awk '{print $5}')
+    local cur_ctx=$(kubectl config view -o=jsonpath='{.current-context}')
+    cur_namespace="$(kubectl config view -o=jsonpath="{.contexts[?(@.name==\"${cur_ctx}\")].context.namespace}")"
```
derekbassett: Do you know what version of Kubectl they added support for -o=jsonpath?

captjt: Hmmm.. The earliest official version on kubernetes.io is v1.5 and it supports `jsonpath` flag --> [kubectl config view](https://kubernetes.io/docs/user-guide/kubectl/v1.5/#-em-view-em-).

bhilburn:
@captjt - Thanks so much for submitting this, and thanks to @derekbassett for providing testing and review! It looks like the unit tests are failing with this:

```
ASSERT:expected:
    <%K{magenta} %F{white%}⎈%f %F{white}minikube/default %k%F{magenta}%f > 
but was:
    <%K{magenta} %F{white%}⎈%f %F{white}/default %k%F{magenta}%f >

ASSERT:expected:
    <%K{magenta} %F{white%}⎈%f %F{white}minikube/kube-system %k%F{magenta}%f > 
but was:
    <%K{magenta} %F{white%}⎈%f %F{white}/default %k%F{magenta}%f >
```

So it's failing because the `minikube` string disappeared. I'm not a kubernetes user (indeed, I know next to nothing about containers), but my guess, here, is that that's basically a username or namespace and `default` is some sort of instance? Regardless, is there value in having that there, and is it easy to fix?

captjt:
I can check this out. I don't remember writing these unit tests maybe they were added after we merged in the first take on kube context's feature. I'll take a look as soon as I can get the time. Thanks @bhilburn!!

powerlevel9k.zsh-theme
derekbassett: Thought it might be important to highlight that.

derekbassett:
Hey so I figured out what was going on.  The mock was not correctly updated.  Also @captjt it looks like you are calling kubectl three time rather then just two times.  I've got the fix in my repo https://github.com/derekbassett/powerlevel9k/tree/fix/kubernetes-prompt.  I thought I was able to PR it to your branch @captjt but my github foo is failing me.

captjt:
Going to close this already approved @derekbassett's PR [#661](https://github.com/bhilburn/powerlevel9k/pull/661). Thanks for taking the time to fix the tests and clean up the code! 😄 

cc. @bhilburn

derekbassett:
Your quite welcome! I wouldn’t have thought to look for config context view. Glad that we got it figured out.

PR #660, opened 2017-10-27
ISS-650 Fix PYTHONPATH appearing in prompt

brianmoran:
- Modify prompt_dir function to set current_path based on pwd and
replace $HOME with ~


brianmoran:
Before
before

After
after

bhilburn:
@brianmoran - Wow! Thanks so much for hunting down the problem and providing a fix! I really do greatly appreciate it, as well all of the rest of our users.

Congratulations on your first PR into the project, and welcome to P9k `=)`

PR #663, opened 2017-10-31
declared variables as 'local'

MarcHauptmann:
Prevents warning messages from zsh


bhilburn:
Thanks for not just ignoring the warnings and submitting a PR to fix them, @MarcHauptmann! I really do greatly appreciate it.

Welcome to P9k, and thanks for the great PR! `=)`

Issue #561, opened 2017-07-07
Always show # of Jobs in Verbose - Even 1


Hacksaurus:
Hello!
I was wondering if there was a way so that powerlevel9k can show the # of jobs staring with #1 and then continue on as normally?

EDIT:
![screenshot_2017-07-07_20-51-36](https://user-images.githubusercontent.com/24832107/27988511-41d705a8-63d8-11e7-911f-4b0e62f3ff58.png)
What I am trying to achieve is where that it will show "1" when even there is only 1 background job.

Thank you!

Hacksaurus:
Very easy fix all you have to do is edit the powerlevel9k.zsh-theme file in the section:
```bash
# Segment to indicate background jobs with an icon.
set_default POWERLEVEL9K_BACKGROUND_JOBS_VERBOSE true
prompt_background_jobs() {
  local background_jobs_number=${$(jobs -l | wc -l)// /}
  local wrong_lines=`jobs -l | awk '/pwd now/{ count++ } END {print count}'`
  if [[ wrong_lines -gt 0 ]]; then
     background_jobs_number=$(( $background_jobs_number - $wrong_lines ))
  fi
  if [[ background_jobs_number -gt 0 ]]; then
    local background_jobs_number_print=""
    if [[ "$POWERLEVEL9K_BACKGROUND_JOBS_VERBOSE" == "true" ]] && [[ "$background_jobs_number" -gt 0 ]]; then
      background_jobs_number_print="$background_jobs_number"
    fi
    "$1_prompt_segment" "$0" "$2" "$DEFAULT_COLOR" "cyan" "$background_jobs_number_print" 'BACKGROUND_JOBS_ICON'
  fi
}
```
All you have to do is change the value from this line `[[ "$background_jobs_number" -gt 1 ]]`
 to zero (`[[ "$background_jobs_number" -gt 1 ]]`)
![screenshot_2017-07-09_18-27-57](https://user-images.githubusercontent.com/24832107/27999669-5e2c3cee-64d4-11e7-91be-a171048ece41.png)
I was able to achieve exactly what I wanted to do

Other stuff:
Also if you wanted to show the number even if there is no background jobs all of you have to do is change the code so its matches this:
 `[[ "$background_jobs_number" -gt 1 ]]`
`[[ background_jobs_number -gt -1 ]]`
`local background_jobs_number_print="0"`
and you get something like this:
[...]

PR #666, opened 2017-11-05
Allow showing the number of jobs even if there is <= 1 job present.

ParthKolekar:
Allow showing the number of jobs even if there is <= 1 job present.

Solves #561 better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,7 @@ your `~/.zshrc`:
 | Variable | Default Value | Description |
 |----------|---------------|-------------|
 |`POWERLEVEL9K_BACKGROUND_JOBS_VERBOSE`|`true`|If there is more than one background job, this segment will show the number of jobs. Set this to `false` to turn this feature off.|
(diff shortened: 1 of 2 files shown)


ParthKolekar:
@bhilburn 

This is a tiny aesthetic commit that does not change default behaviour, can this be merged?

bhilburn:
@ParthKolekar - Thanks so much for submitting this PR! I made some minor changes, including cleaning up some language, and changed the variable name from `_EVEN_ONE` to `_ALWAYS` to be more consistent with our other fields. Note that I am merging this into `next`, instead of `master`, as it's technically a new feature and we try to keep `master` clear of anything except bug fixes between releases. 

Thanks for putting this together! I especially appreciate you updating the docs, as well. Pull & switch over to the next branch to use your code in mainline, and welcome to P9k! `=)`

ParthKolekar:
I actually use prezto... So I guess I need to search for how to set a git submodule to this branch.... 😅 

Thanks for merging.

bhilburn:
@ParthKolekar - git submodules are actually exactly like normal git repos, except their "HEADS" are usually detached... I'm not a prezto user myself, but you should be able to just `cd` into the directory and do a `git checkout -b next -t origin/next` and be good, hah :)

PR #667, opened 2017-11-07
Add support for both AWS_PROFILE and AWS_DEFAULT_PROFILE

akranga:
`AWS_PROFILE` env var has been respected by AWS client as well as `AWS_DEFAULT_PROFILE`. See here: http://docs.aws.amazon.com/cli/latest/userguide/cli-multiple-profiles.html

This small change adds `AWS_PROFILE` in coalesce with `AWS_DEFAULT_PROFILE`

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -287,7 +287,7 @@ prompt_anaconda() {
 
 # AWS Profile
 prompt_aws() {
-  local aws_profile="$AWS_DEFAULT_PROFILE"
+  local aws_profile="${AWS_PROFILE:-$AWS_DEFAULT_PROFILE}"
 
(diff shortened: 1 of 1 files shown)


Loki22:
Yes please add this. I think AWS_PROFILE is the correct one anyway, at least for most things CLI related that use common libraries. The link below is the latest docs and they do not even reference AWS_DEFAULT_PROFILE, so perhaps that was deprecated.

http://docs.aws.amazon.com/cli/latest/userguide/cli-environment.html

akranga:
Waiting for PR to be merged

bhilburn:
Thanks so much for submitting this fix, @akranga! I really appreciate it, and thanks for linking to the relevant docs!

Welcome to P9k `=)`

PR #676, opened 2017-11-14
Kubernetes segment shorter if possible

### LINKED ISSUE #675: Don't print kubernetes segment's context if namespace is equal
I usually name my contexts after namespaces, I see no reason to see `spam/spam`. Simple check to print only one of these if they are equal would be great!

iScrE4m:
Fixes #675


bhilburn:
@iScrE4m - So sorry for the delayed response, here! Thanks so much for submitting this! This is a great usability improvement `=)`

PR #680, opened 2017-11-22
pyenv promt segment uses $PYENV_VERSION environment variable"

### LINKED ISSUE #679: Active virtualenv name is not correct for pyenv
When I activate multiple versions by `pyenv global devtools2 devtools3 3.2 2.7`, the promt segment shows that I've activated the first (`devtools`).
I think it should not show any thing here because they are global environment.
The issue can be fixed by using `PYENV_VERSION` environment variable. Check the document [here](https://github.com/pyenv/pyenv#choosing-the-python-version).
And I'm willing to fix that in a pull request.

dinhnv:
Fix #679

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1379,18 +1379,10 @@ prompt_virtualenv() {
 }
 
 # pyenv: current active python version (with restrictions)
-# More information on pyenv (Python version manager like rbenv and rvm):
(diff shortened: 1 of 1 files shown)


bhilburn:
Good catich, @dinhnv, and thanks so much for submitting the fix! I really appreciate it.

Welcome to `P9k` as a contributor! This is a great PR =)

PR #682, opened 2017-11-25
Use color names to help themes. +fix

Artistan:
use color names by default to allow dark/light themes to adjust colors based on names.
Also fix a bug with POWERLEVEL9K_DIR_OMIT_FIRST_CHARACTER 
 - truncates the $packageName when using 'truncate_with_package_name
 - added tests for both true and false for the omit option


Artistan:
not sure why
`The command "test/segments/rust_version.spec" exited with 1.`

Artistan:
Yep @diraol , that .idea folder/files was a mistake. I am used to php projects that have the .idea as a default in the .gitignore due to a lot of devs using Idea IDEs.  I have removed that and updated the last of the tests to check for the named color vs the 0-255 color codes.

Artistan:
The 'prompt_os_icon' color being hard coded number was messing with multiple iterm themes that used named colors (black/white) for editing the theme. This update fixes that in multiple other areas that used 0-255 colors. This help make it more readable for code and for theme settings.

Artistan:
Added a fix for POWERLEVEL9K_DIR_OMIT_FIRST_CHARACTER truncating the $packageName when  using 'truncate_with_package_name'

Artistan:
also optimized to use $current_path rather than calling `pwd | sed -e "s,^$HOME,~,"` multiple times.

Artistan:
needed to test the truncate_with_package_name with or without POWERLEVEL9K_DIR_OMIT_FIRST_CHARACTER

PR #685, opened 2017-11-27
vcs: Disable alias of grep to fix performance penalty

kenhys:
If user defined alias of grep command with '-r' or '--recursive'
option, it causes a slow response of command line. It is not
unexpected behavior of vcs info.

This commit explicitly disables alias of grep to suppress side
effects.


bhilburn:
@kenhys - Wow! This is great! This is a subtle issue with a big impact. Thanks so much for upstreaming the fix! I really appreciate it.

Welcome to P9k `=)`

PR #686, opened 2017-11-27
Fixed Issue with whitespaces in path.

Corwynt:
Whitespaces in Path lead to an error in ´prompt_dir()`. To fix this, the IFS variable which defines the split-chars for assignments is temporarily set to an empty string.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -695,10 +695,10 @@ set_default POWERLEVEL9K_DIR_PATH_SEPARATOR "/"
 set_default POWERLEVEL9K_HOME_FOLDER_ABBREVIATION "~"
 set_default POWERLEVEL9K_DIR_SHOW_WRITABLE false
 prompt_dir() {
-  local tmp=$IFS
+  local tmp="$IFS"
   local IFS=""
   local current_path=$(pwd | sed -e "s,^$HOME,~,")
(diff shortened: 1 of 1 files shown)


Nuntis:
It's not a full fix (it doesn't show error msg but the path is not complete) but thank you anyway

Corwynt:
@Nuntis can you tell me which path doesn't work for you?

Nuntis:
Wow, my bad, it's my fault, I didn't reload my shell u_u
It works perfectly !
Thank you :D

pucicu:
Great! Thank you very much! Works well now.

bhilburn:
Hey @Corwynt! Thanks so much for submitting this PR! The `dir` segment is pretty heavily used, so I do want to make sure we get this bugfix merged.

Currently, this PR is breaking unit tests, and it looks like it may be due to the use of `local`. Have a look:
https://travis-ci.org/bhilburn/powerlevel9k/builds/308097509

What version of ZSH did you test this with locally? I wonder if the behavior has changed.

Corwynt:
@bhilburn The Tests seem to run now, without an error. I was missing some quotes around a variable assignment

bhilburn:
@Corwynt - Thanks for the fix! Looks great.

Thanks again for putting this PR together, and welcome to P9k! `=)`

PR #661, opened 2017-10-27
Fix/kubernetes prompt

derekbassett:
Added functionality to parse the current namespace regardless of config settings. This should address the issue #555.

PR #693, opened 2017-12-05
Fix of k8s shorter segment

iScrE4m:
My latest PR #676 seemed to have been incompatible with #661 - I did my PR on top of master instead of next. This one is done on top of next and should fix the current not so great state of next (prints nothing). Sorry!

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1430,11 +1430,11 @@ prompt_kubecontext() {
 
     local k8s_final_text=""
 
-    if [[ "$k8s_context" == "k8s_namespace" ]]; then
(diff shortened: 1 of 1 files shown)


bhilburn:
Thanks for the fix, @iScrE4m! I'm actually surprised that #676 passed unit testing with that breakage. We clearly have a gap in test coverage that we need to fix.

Thanks again for the PRs, and for fixing this issue! Much appreciated =)

bhilburn:
Oh, wait, I understand what happened. The tests were fine on `master`, but when I re-targeted your branch target it didn't re-build against `next`, so I never saw the failure.

Good lesson learned, there.

PR #704, opened 2017-12-19
feature: add custom number of newlines after prompt

epiales:
I like to have a couple of newlines in between each of my prompts. This change would allow the user to set the number of newlines they prefer if they want more than the default of 1.


bhilburn:
Hey @epiales! Thanks so much for the PR! This is a great new feature! A few review items:

1. Since this is a new feature, it needs to be merged into `next`, not master. As long as you branched from `master` not very long ago, it wouldn't be that big of a deal for me to just re-target the PR for you, but as a rule of thumb you want to branch from `next` for feature branches and submit the PR against the same. See [this page of the wiki](https://github.com/bhilburn/powerlevel9k/wiki/Developer's-Guide#branches) for more.
2. I think we could improve the code by not using a `for` loop. It seems to me that we should be able to create a string with some quick & easy ZSH regex, which should be much faster than a for-loop. @dritter @rjorgenson - Any chance either of you are up for helping @epiales out with the regex?
3. Thanks for using the `defined` function properly! `=)`

powerlevel9k.zsh-theme
```diff
@@ -1519,7 +1519,14 @@ $(print_icon 'MULTILINE_LAST_PROMPT_PREFIX')'
   fi
 NEWLINE='
 '
-  [[ $POWERLEVEL9K_PROMPT_ADD_NEWLINE == true ]] && PROMPT="$NEWLINE$PROMPT"
+  if [[ $POWERLEVEL9K_PROMPT_ADD_NEWLINE == true ]]; then
+    defined POWERLEVEL9K_PROMPT_ADD_NEWLINE_COUNT || POWERLEVEL9K_PROMPT_ADD_NEWLINE_COUNT=1
```
Artistan: How about a one liner...
```bash
PROMPT="$(yes '\n' | head -n$POWERLEVEL9K_PROMPT_ADD_NEWLINE_COUNT)$PROMPT"
```

openam: `yes` example didn't work in my OSX machine.  How about

```
if [[ $POWERLEVEL9K_PROMPT_ADD_NEWLINE == true ]]; then
  PROMPT="$(printf "\n\ufeff"%.0s {1..${POWERLEVEL9K_PROMPT_ADD_NEWLINE_COUNT:-1}})$PROMPT"
fi
```

epiales: I tried a couple variations of the `yes` command  using `'\n'`, `'$NEWLINE'`, `'\ '`, `' '`, `''` but looks like the BSD version of `yes` doesn't play nicely.

@openam's suggestion is working for me and I'm also open to other alternatives.

Could even change it to

```zsh
[[ $POWERLEVEL9K_PROMPT_ADD_NEWLINE == true ]] && PROMPT="$(printf "\n\ufeff"%.0s {1..${POWERLEVEL9K_PROMPT_ADD_NEWLINE_COUNT:-1}})$PROMPT"
```

Artistan: That looks good to me.

bhilburn:
Let me know when you update the PR with new commits! =)

powerlevel9k.zsh-theme
bhilburn: Thanks for the reviews, @openam & @Artistan!

openam: I've been trying to run the option I suggested for a little bit. It doesn't appear to work in all terminals. I have had issues with it in the vscode integrated terminal, and my Android JuiceSSH client. It appears that the original for loop works fine in both of them. I think it should just be included as originally proposed, unless someone else has some good ideas.

bhilburn:
Hey @epiales! I don't want to lose track of this PR - it's a useful feature, and I appreciate the work you've put into it.

It sounds like the original suggestion from @Artistan won't quite work as it's not as portable as we need it to be. The `for` loop does seem a bit wasteful, though, and I would be surprised if there wasn't a way to do this with ZSH regex. Let's see if @docwhat or @rjorgenson have any thoughts for how to do it better.

epiales:
@bhilburn Thanks for the follow up. I've been trying to come up with a regex way but haven't been able to find anything that will repeat replacement _n_ number of times. I've tried a couple other variations but haven't found many solutions which work across different terminals. I'm still open to other suggestions to try. An alternative which works using `repeat`:
```zsh
if [[ $POWERLEVEL9K_PROMPT_ADD_NEWLINE == true ]]; then
    NEWLINES=""
    repeat ${POWERLEVEL9K_PROMPT_ADD_NEWLINE_COUNT:-1} { NEWLINES+=$NEWLINE }
    PROMPT="$NEWLINES$PROMPT"
fi
```

docwhat:
Have you tried the newline widget?  I did submit that? Right?

epiales:
@docwhat If I understand your widget correctly, I think that only continues segments on the next line?  I haven't tried extensively, other than I have `POWERLEVEL9K_PROMPT_ON_NEWLINE=true` and have modified the prefixes. I initially tried changing to
```zsh
POWERLEVEL9K_MULTILINE_FIRST_PROMPT_PREFIX='


'
```
and
```zsh
POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(newline newline newline os_icon status context dir vcs)
```
but those didn't work for me.

Here's an example of what I'm trying to achieve for posterity.

image

image

Artistan:
Note, the new line option also allows a special character.
How are you going to deal with that?
image

openam:
The [solution](https://github.com/bhilburn/powerlevel9k/pull/704#issuecomment-362027805) @epiales suggested works with the special characters.

screen shot 2018-02-01 at 7 55 54 am

bhilburn:
@epiales - I like the solution you suggested, and @openam has confirmed it handles special characters well. Can you update this PR with the change, and document the new feature in the README? As soon as you push the new commit(s), ping me here! I'm excited to get this merged into `next` =)

epiales:
@bhilburn I've updated the PR. I added a section to the README but it looks like most of the stylizing is defined on the [wiki page](https://github.com/bhilburn/powerlevel9k/wiki/Stylizing-Your-Prompt), will someone also need to update that?

bhilburn:
@epiales - Everything looks good! Sorry for the long turn-time on this. I really appreciate you taking the time to put this together and then incorporate feedback as we reviewed it (normally it doesn't take this long!). It's really great that you implemented a feature you wanted, and then upstreamed it! Well done, and welcome to P9k `=)`

PR #709, opened 2017-12-23
Add remote state to context prompt

tippl:
Adds `REMOTE` state to color prompt with in case of a shell over ssh.
This state has lower priority than `ROOT` state to not lose the privilege warning.


bhilburn:
@tippl - I like this. Well done. Your implementation is clean & simple, follows our coding style, and is submitted against our feature branch. Thanks for another great PR! `=)`

PR #713, opened 2018-01-02
Corrected VPN IP command

magicalraccoon:
As it stands, using `vpn` in .zshrc returns the error `bash: prompt_vpn: command not found...`.

`vpn_ip` is the correct term.


bhilburn:
OOPS. You're absolutely correct, @magicalraccoon. Thanks so much for submitting the fix! I really appreciate it. Welcome to P9k `=)`

PR #722, opened 2018-01-15
fix: Remove dependency of bc for load widget

lemarsu:
Hi,

I sometime, have hosts that do not have `bc` program installed. The load widget use bc for doing math that zsh can easily handle.

This PR replace math done with `bc` by math done with zsh.

Best regards.


bhilburn:
@lemarsu - Thanks so much! This is a great PR - I'm always a fan of reducing dependencies on outside programs. Welcome to P9k `=)`

PR #726, opened 2018-01-20
getColor() must always print result

docwhat:
This was causing colors specified as numbers to not be shown.


bhilburn:
Thanks so much for the quick fix, @docwhat! Much appreciated =)

PR #729, opened 2018-01-25
Added sudo state to prompt_user

Mikewl:
Before I discovered zsh, I was using oh-my-bash. What I missed from that was that if I still had elevated privileges that I could see that. This adds the state of `SUDO` to `user_prompt` which gives this capability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,7 +312,7 @@ main theme distribution so that everyone can use it!
 
 The `context` segment (user@host string) is conditional. By default, it will
 only print if you are not your 'normal' user (including if you are root), or if
-you are SSH'd to a remote host.
+you are SSH'd to a remote host. `SUDO` and `REMOTE_SUDO` states are also available to show whether the current user or remote user has superuser privileges.
 
 To use this feature, make sure the `context` segment is enabled in your prompt
 elements (it is by default), and define a `DEFAULT_USER` in your `~/.zshrc`.
@@ -580,6 +580,7 @@ You can also override the icons by setting:
 ```
 POWERLEVEL9K_USER_ICON="\uF415" # 
 POWERLEVEL9K_ROOT_ICON="#"
+POWERLEVEL9K_SUDO_ICON=$'\uF09C' # 
 ```
 
 | Variable | Default Value | Description |
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -258,6 +258,7 @@ case $POWERLEVEL9K_MODE in
       RIGHT_SUBSEGMENT_SEPARATOR     $'\uE0B3'              # 
       CARRIAGE_RETURN_ICON           $'\u21B5'              # ↵
       ROOT_ICON                      $'\uE614 '             # 
+      SUDO_ICON                      $'\uE0A2'              # 
       RUBY_ICON                      $'\uF219 '             # 
       AWS_ICON                       $'\uF270'              # 
       AWS_EB_ICON                    $'\UF1BD  '            # 
@@ -332,6 +333,7 @@ case $POWERLEVEL9K_MODE in
       RIGHT_SUBSEGMENT_SEPARATOR     $'\uE0B3'              # 
       CARRIAGE_RETURN_ICON           $'\u21B5'              # ↵
       ROOT_ICON                      $'\u26A1'              # ⚡
+      SUDO_ICON                      $'\uE0A2'              # 
       RUBY_ICON                      ''
       AWS_ICON                       'AWS:'
       AWS_EB_ICON                    $'\U1F331 '            # 🌱
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -27,6 +27,7 @@ case $POWERLEVEL9K_MODE in
       RIGHT_SUBSEGMENT_SEPARATOR     $'\uE0B3'              # 
       CARRIAGE_RETURN_ICON           $'\u21B5'              # ↵
       ROOT_ICON                      $'\uE801'              # 
+      SUDO_ICON                      $'\uF09C'              # 
       RUBY_ICON                      $'\uE847 '             # 
       AWS_ICON                       $'\uE895'              # 
       AWS_EB_ICON                    $'\U1F331 '            # 🌱
@@ -105,6 +106,7 @@ case $POWERLEVEL9K_MODE in
       RIGHT_SUBSEGMENT_SEPARATOR     $'\uE0B3'              # 
       CARRIAGE_RETURN_ICON           $'\u21B5'              # ↵
       ROOT_ICON                      $'\uF201'              # 
+      SUDO_ICON                      $'\uF09C'              # 
       RUBY_ICON                      $'\uF219 '             # 
       AWS_ICON                       $'\uF270'              # 
       AWS_EB_ICON                    $'\U1F331 '            # 🌱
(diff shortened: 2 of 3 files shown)


tippl:
Few comments:
- This is a new feature, you should PR into the `next` branch, not `master`.
- This might be hard to notice, but the `ROOT` state has color 011 which is yellow. (has been changed in `next` to the actual name) So the new state should probably also be by default yellow so it is up to the user to set their colors.
- `sudo -n uptime 2>&1 | grep -q "load"` The grep here is redundant calling of another process. If `sudo -n` fails to execute uptime it returns nonzero exit code, which is enough for if, so `elif sudo -n uptime >/dev/null 2>&1; then`

bhilburn:
@Mikewl - Thanks so much for submitting this! We love seeing first-time contributors.

@tippl's review comments are good. One of the things I'm struggling with here, I think, is how this feature really gets used. The `user` segment obviously has the root feature already, and since we aren't re-drawing the prompt as keys are being pressed, it will never update when it sees the word `sudo`. In short, when would this become active?

tippl:
@bhilburn The way I understand this feature. When you use sudo, after you enter password there is a time for which sudo won't ask for password again. This would be active during that timeout to notify you that sudo won't ask for a password.

Another thing i'm thinking is whether to add this feature to the `context` segment, as it would be probably useful there too.

Mikewl:
@tippl thanks for the comments! I will get to those in the morning. Your explanation is far better than mine for the feature. It is something I personally like and missed - why I added it!

It would make sense to add it to `context` as well I feel. Still need to play with that as it was always showing me `user@host` while I was logged in locally using iTerm2. I would have preferred it if there was an option to only show the hostname when remote.

tippl:
@Mikewl For `context` to show your username only you need to set 2 variables.
`DEFAULT_USER` to your username, this makes the context not show for you
`POWERLEVEL9K_ALWAYS_SHOW_USER` to true so it shows username even if it matches default.
It will show hostname when your username differs from default one or when remoted via ssh.

bhilburn:
@tippl @Mikewl - Thanks for the discussion! And yes, I understand now. Thanks for the explanation! And agreed, this is a very useful feature! There are a few things I want to make sure get done before we merge this and make it available =)

1. So I'm assuming that you chose `uptime` because it's relatively simple and easy to parse? If that's the case, I think it might be better to switch to something more basic. Since `uptime` reaches out to `/proc`, it's a bit expensive when utilized just to see if a command will run. I wonder if something like:

`sudo -n echo "true" | grep -q "true"` 

might serve the purpose better?

2. I think we need to use an icon other than the `ROOT_ICON`, which I'm worried people will find confusing given that they aren't *actually* root - especially the users that don't have their username shown as part of their prompt (including myself!)

3. And, of course, we need to add this to the README =)

@Mikewl - This is a really great feature, and I'm excited to get it merged. Are you up for handling the items, above?

tippl:
@bhilburn True with uptime being maybe the wrong command to use here, but again like i wrote in my first comment, the grep and parsing the output is redundant too. `sudo -n` already has nonzero return values for failed to execute command. Since this applies, all we need is to redirect the sudo error message to null and use even simpler command that doesn't write to stdout. Maybe just `true` command.

`sudo -n true 2>/dev/null`

Mikewl:
`sudo -n true 2>/dev/null` does work!

tippl:
The if/elif in context works like priority system. I think the `SUDO` should maybe have a higher priority as a state than `REMOTE`.

Considering the colors, i do wonder why we even set them in the state array, since all default fg/bg colors in `user` and `context` are the same. We could probably just remove that part and set the colors in calling `"$1_prompt_segment"`. Any comments on this? @bhilburn

Mikewl:
`context` is changed. I was just blind and missed a typo which was not causing the detection. Should I still move the priority higher?

Also: If we are choosing a new icon, should this be added to `root_indicator` as well maybe?

Mikewl:
Possibly use https://fontawesome.com/icons/lock-open?style=solid for sudo icon?

bhilburn:
@Mikewl - the unlocked icon works! Nice find.

@tippl - The colors are declared that way simply so that users can override them. In any segment with different "action states" we have to call them out like that so that the fields exist for each state.

So, `prompt_user` doesn't actually have an indication for `REMOTE` - `prompt_context` does (see [this line on `next` branch](https://github.com/bhilburn/powerlevel9k/blob/next/powerlevel9k.zsh-theme#L594)).

I don't think we should make this change to `prompt_root_indicator`, but there is a question of whether or not we add it to `prompt_context`, I think, which would be useful. One of the more difficult parts of this, which I think @tippl was alluding to, is that `sudo` status is orthogonal to being local or `REMOTE`. Really, we would want to be able to indicate `sudo` for both `DEFAULT` and `REMOTE` states. Are you up for taking a shot at that, @Mikewl?

Mikewl:
More than up for it.

Just busy hunting for the code points for the symbol right now. Then I'll quickly modify the docker file to point at my repo and get to trying out getting `remote_sudo` going over there.
I will honestly say that I have no idea how to get powerline9k going over ssh though so this may take me a little bit longer.

tippl:
It's not about getting it running over ssh. You just need a ssh server that has p9k installed. Unfortunately i don't think it's possible to run your local p9k on any ssh server you connect to, so it's for when you connect to your pc that has p9k from a remote pc. @Mikewl

bhilburn:
On my phone now reading this by email, but ah, yes, I think you are right
@tippl. What we are doing is noisy & clunky - your suggestion makes it
cleaner. Would need to test to be certain, but I'm pretty sure you're right.

Sent from my mobile.

On Feb 10, 2018 14:48, "tippl" <notifications@github.com> wrote:

@bhilburn <https://github.com/bhilburn> Yeah, i get that, but isn't color
overriding handled in the $1_prompt_segment?

context_states=(
    "ROOT"      "yellow"
    "DEFAULT"   "yellow"
    "REMOTE"    "yellow"
)

The way I see it this declaration is useless.

"$1_prompt_segment" "${0}_${current_state}" "$2" "$DEFAULT_COLOR" "yellow"
"${content}"
Doing this would work just as well, since color overrides are not handled
here user color changes still work. I just see no need to have a color
array for default colors, when all of them are the same.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/bhilburn/powerlevel9k/pull/729#issuecomment-364684412>,
or mute the thread

Mikewl:
Changes made, unsure if all the icon code points are correct. I know powerline and nerdfonts definitely are.
`REMOTE_SUDO` should work but is untested.

If there are no comments/changes then I can rebase and it should be ready to merge.

bhilburn:
@Mikewl - I know this PR has been a couple of months in the making, and I apologize for the long turn-time, but this is great work. I really appreciate that you took the time to implement a feature you wanted, submitted it as a PR, and then iterated with us as we reviewed, discussed, and provided feedback.

Great work, and thanks so much for putting in the effort. Welcome to P9k `=)`

Mikewl:
Thanks! The feedback was very useful. Glad to be contributing to a tool which I am using pretty much all day, everyday now.

onaforeignshore:
This segment "sort-of" works.

This line `sudo -n true` tests if a password is required for sudo. If the user is in the `sudoers` group and doesn't require to enter a password, then this is ALWAYS true, so the prompt will always be in this mode if the user isn't root.

bhilburn:
@onaforeignshore - You have a good point, here. Perhaps the best thing to do is to add an option that disables this feature so that people with that configuration don't get annoyed by it.

onaforeignshore:
This code will allow us to add a fourth state - for users that have sudo and don't require a password:
```zsh
has_sudo() {
  local sudoStatus
  sudoStatus=$(sudo -nv 2>&1)

  if (( $? == 0 )); then
    echo "sudo__passwd_set"
  elif echo $sudoStatus | grep -q '^sudo:'; then
    if sudo -n true 2>/dev/null; then
      echo "sudo__no_passwd_needed"
    else
      echo "sudo__needs_passwd"
    fi
  else
    echo "no_sudo"
  fi
}
```

PR #740, opened 2018-02-08
Fix vi-mode and support hiding segment when in insert mode

### LINKED ISSUE #725: Possible to make vi_mode work with ^X^V?
While using the regular emacs mode, you can temporarily enter vi mode by pressing ctrl+x ct+v.

This is really useful for moving the cursor around. It would be cool to be able to have the vi_mode indicator show up for NORMAL mode and maybe just hide it completely for INSERT mode.

Is that something that could be added?

ionsquare:
I believe [this commit](https://github.com/bhilburn/powerlevel9k/commit/520eed12482b276e630d66b01a3f5852c96b6b1f) stopped the vi_mode plugin from working (also see [this comment](https://github.com/bhilburn/powerlevel9k/issues/319#issuecomment-355870476) from issue #319).

This PR returns one of the parts that was removed in that commit to get vi_mode working again as suggested by the comment linked above.

It also adds a check to see if `POWERLEVEL9K_VI_INSERT_MODE_STRING` is an empty string and if so, it skips showing the segment entirely.

This works both when using vi mode directly and when temporarily entering vi mode from emacs mode with `^X^V`.

This fixes #725 and possibly #319.

### DIFF
--- a/README.md
+++ b/README.md
@@ -621,6 +621,8 @@ you are using the [ZSH Line Editor](http://zsh.sourceforge.net/Doc/Release/Zsh-L
 |`POWERLEVEL9K_VI_INSERT_MODE_STRING`|`"INSERT"`|String to display while in 'Insert' mode.|
 |`POWERLEVEL9K_VI_COMMAND_MODE_STRING`|`"NORMAL"`|String to display while in 'Command' mode.|
 
+To hide the segment entirely when in `INSERT` mode, set `POWERLEVEL9K_VI_INSERT_MODE_STRING=''`
+
 #### Unit Test Ratios
 
 The `symfony2_tests` and `rspec_stats` segments both show a ratio of "real"
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1366,6 +1366,7 @@ prompt_vi_mode() {
       "$1_prompt_segment" "$0_NORMAL" "$2" "$DEFAULT_COLOR" "default" "$POWERLEVEL9K_VI_COMMAND_MODE_STRING"
     ;;
     main|viins|*)
+      if [[ -z $POWERLEVEL9K_VI_INSERT_MODE_STRING ]]; then return; fi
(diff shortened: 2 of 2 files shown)


bhilburn:
Hey @ionsquare - Thanks so much for submitting this PR! As you might have seen in [my comment](https://github.com/bhilburn/powerlevel9k/issues/319#issuecomment-261554849) in #319, the reason I never merged #320 (and the reason I pulled the `zle` commands, to begin with), is that the performance impact of having them was *tremendously* negative. The VCS segment needs some speeding up in its current state, but those `zle` commands made it horrendous for non-Vi_mode users.

In testing this branch, I have to say I'm stunned - it actually **speeds up P9k**. It is obviously a subset of #320, but if this fixes #319 that would be fantastic, because it doesn't seem to slow anything down at all.

I'm testing this on macos right now where I see the speed-up. I'd like to test it on Linux, too, to be sure. It would be great if a few others wouldn't mind pulling this branch to give it a go, as well. Merging things into `master` with this broad of an impact can be really damaging if we make a mistake, so I want to be sure it's thoroughly tested.

Based on my tests so far, though, this is really excellent, @ionsquare!

@dritter @iScrE4m - Would you mind pulling this branch and giving it a go? I'll test on my Linux laptop in a bit.

iScrE4m:
I pulled it, then looked into how to test it and realized that since I've never used zle before, I'm probably not the best candidate for testing, sorry!

bhilburn:
@ionsquare - I want to get this into the mainline, and at the same time I'm still just a touch nervous of something going wrong if I merge it into `master` - touching the `zle` stuff can break things in bizarre ways.

So, I've merged it into `next`. We are doing a bunch of work on `next` right now, and I hope to tag a new release in fairly short order. This work will get tested along with the rest of that, and then will be part of the next release.

Thanks so much for contributing this, by the way. I'm still stunned that something so simple seems to have such a positive impact on the problem after so many of us have struggled to make a meaningful fix, here. Great work, and welcome to P9k `=)`

ionsquare:
Thanks for the update, I'm really glad I was able to help. I hope there aren't any unfortunate surprises as it gets tested out in `next`.

PR #741, opened 2018-02-10
Option to place icons on the left in the right prompt segments

onaforeignshore:
This PR allows a user to change the icon location on the right prompt segment to the left if desired by setting `POWERLEVEL9K_RPROMPT_ICON_LEFT=True`.


onaforeignshore:
Wrong PR

PR #742, opened 2018-02-10
Option to place icons on the left in the right prompt segments

onaforeignshore:
For the right prompt segments, if a user sets `POWERLEVEL9K_RPROMPT_ICON_LEFT=true`, the icon will be printed on the left instead of the right.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 test-vm/.vagrant
 *.swp
 .idea
+sh2md
--- a/README.md
+++ b/README.md
@@ -82,6 +82,15 @@ your `~/.zshrc`:
 POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(context dir rbenv vcs)
 POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(status root_indicator background_jobs history time)
 ```
+
+You can customize each prompt segment by setting the following variables:
+| Variable | Default Value | Description |
+|----------|---------------|-------------|
+|`POWERLEVEL9K_segment_FOREGROUND`|Global foreground color|A custom color from 0 to 255.|
+|`POWERLEVEL9K_segment_BACKGROUND`|Global foreground color|A custom color from 0 to 255.|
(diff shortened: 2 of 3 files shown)


onaforeignshore:
Since this segment deals with icons in the right prompt, I have added the option to make segments bold into this, otherwise, it would cause issues when merging this into next.

onaforeignshore:
Tests will have to be updated to expect the bold reset string `%b` in the result, which is why Travis is failing.

bhilburn:
@onaforeignshore - Thanks for submitting this PR! Looking through the changeset, it looks like this PR actually includes three different new features?

1. icons on the right
2. bold fonts
3. iTerm2 integration (thus being a superset of #743)

I suspect these items may have been done in a way that is not easily separable, which is why they all got swept up into this PR? If so, we'll need to make sure that the README and tests are updated for all of them, and obviously test all three of those features.

I am a bit confused about the bold fonts code in this PR, though, vs the bold fonts code in #744. Is one a super/subset of the other? Are they both required for the bold fonts to work?

onaforeignshore:
The bold set in #744 is ONLY for the last dir in the `prompt_dir` segment, so you get something like `~/code/**powerlevel9k**`, whereas this one bolds the whole segment `**~/code/powerlevel9k**` so overrides the other

onaforeignshore:
1 - Icons on the left of right segments creates a conditional in the code where you draw icon/text or text/icon
2 - If this is implemented, you need to bold both separately, which is why I included it here instead of against next, as you would have to update the code again when you applied either separately.
3 - slipped into this accidentally and isn't part of this, but if you want all 3 you can merge this and delete #743.

onaforeignshore:
I will update this (including tests) once #744 has been merged.

bhilburn:
@onaforeignshore - Okay, so it sounds like you want to put this PR into a holding pattern until #744 is resolved?

It might actually be best to close this one, to be honest. It's really best to handle new features / developments in separate PRs where possible, so we would want to do iTerm2 in #743, not here, which means this PR will conflict once that PR is merged.

onaforeignshore:
Waiting for #744 to be merged. Closing for now.

PR #743, opened 2018-02-10
Fix iTerm2 integration with PowerLevel9k

onaforeignshore:
This fix allows the user to see the "little blue triangle" that is created when iTerm2 shell integration is enabled.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1623,6 +1623,9 @@ $(print_icon 'MULTILINE_LAST_PROMPT_PREFIX')'
 NEWLINE='
 '
   [[ $POWERLEVEL9K_PROMPT_ADD_NEWLINE == true ]] && PROMPT="$NEWLINE$PROMPT"
+
+  # Allow iTerm integration to work
+  [[ $ITERM_SHELL_INTEGRATION_INSTALLED == "Yes" ]] && PROMPT="%{$(iterm2_prompt_mark)%}$PROMPT"
 }
 
 set_default POWERLEVEL9K_IGNORE_TERM_COLORS false


powerlevel9k.zsh-theme
```diff
@@ -1623,6 +1623,9 @@ $(print_icon 'MULTILINE_LAST_PROMPT_PREFIX')'
 NEWLINE='
 '
   [[ $POWERLEVEL9K_PROMPT_ADD_NEWLINE == true ]] && PROMPT="$NEWLINE$PROMPT"
+
+  # Allow iTerm integration to work
+  [[ $ITERM_SHELL_INTEGRATION_INSTALLED == "Yes" ]] && PROMPT="%{$(iterm2_prompt_mark)%}$PROMPT"
```
bhilburn: Does `ITERM_SHELL_INTEGRATION_INSTALLED` get set automatically by iTerm2? And it gets set to the string `"Yes"`, not `true`? I'm asking because I don't know, hah.

Same questions for `iterm2_prompt_mark`.

onaforeignshore: `ITERM_SHELL_INTEGRATION` gets set to `Yes` in the iTerm shell integration script, which you have to download and source separately after installing iTerm if you want that functionality. The instructions are on their website.

levbishop:
I think it's necessary to also add `$(iterm2_prompt_end)` to the end of `PROMPT` for the iterm2 command history to work properly.

This PR seems like perhaps the wrong approach though since the iterm2 integration does try to set PS1 properly by itself? See https://github.com/gnachman/iterm2-website/blob/master/source/shell_integration/zsh#L47 for the implementation and intended logic. Maybe the problem is sequencing of the `precmd_functions` and `preexec_functions`?

onaforeignshore:
As I understand the code...

```zsh
    iterm2_decorate_prompt() {
      # This should be a raw PS1 without iTerm2's stuff. It could be changed during command
      # execution.
      ITERM2_PRECMD_PS1="$PS1"
      ITERM2_SHOULD_DECORATE_PROMPT=""

      # Add our escape sequences just before the prompt is shown.
      if [[ $PS1 == *'$(iterm2_prompt_mark)'* ]] # <--- Test if iterm2_prompt_mark is already in the prompt
      then
        PS1="$PS1%{$(iterm2_prompt_end)%}" # <--- Add only iterm2_prompt_end
      else
        PS1="%{$(iterm2_prompt_mark)%}$PS1%{$(iterm2_prompt_end)%}" # <--- Add both.
      fi
    }
```

The reason we set it manually instead of letting the iTerm integration script set it, is because it sets it on the line preceding the prompt, and not the prompt line itself.

PR #744, opened 2018-02-10
PROMPT_DIR segment updates

onaforeignshore:
I have taken @dritter 's latest code from `async_all_the_segments` and merged it with the current code in `next` to create a starting point for working on updating this segment. He has already done some optimizations to speed up the code.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -724,6 +724,7 @@ set_default POWERLEVEL9K_HOME_FOLDER_ABBREVIATION "~"
 set_default POWERLEVEL9K_DIR_PATH_HIGHLIGHT_BOLD false
 prompt_dir() {
   local current_path="$(print -P '%~')"
+  local num_paths=(${(s:/:)current_path})
   local paths directory test_dir test_dir_length trunc_path threshhold
   (( ${#current_path} > 1 )) && paths=(${(s:/:)${current_path//"~\/"/}}) || paths=() # only split if not root/home folder
   if [[ -n "$POWERLEVEL9K_SHORTEN_DIR_LENGTH" || "$POWERLEVEL9K_SHORTEN_STRATEGY" == "truncate_with_folder_marker" ]]; then
@@ -738,7 +739,7 @@ prompt_dir() {
         current_path=$(truncatePath "$current_path" $POWERLEVEL9K_SHORTEN_DIR_LENGTH $POWERLEVEL9K_SHORTEN_DELIMITER)
       ;;
       truncate_with_folder_marker)
-        if (( ${#paths} > 0 )); then # root is an exception and won't have paths
+        if (( ${#num_paths} > 0 )); then # root is an exception and won't have paths
           local last_marked_folder marked_folder
           set_default POWERLEVEL9K_SHORTEN_FOLDER_MARKER ".shorten_folder_marker"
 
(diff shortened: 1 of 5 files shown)


onaforeignshore:
TODO: Add tests for `POWERLEVEL9K_DIR_PATH_SEPARATOR_FOREGROUND`

onaforeignshore:
Busy removing `sed` from `prompt_dir` wherever possible. Next step is to tackle `function truncatePathFromRight()` in `utilities.zsh` and the 3 calls to `sed` for separator coloring, and custom separator and home path

onaforeignshore:
Only 8 tests in `dir.spec` still failing. Will work on them this week

powerlevel9k.zsh-theme
```diff
@@ -718,21 +718,78 @@ prompt_command_execution_time() {
 # Dir: current working directory
 set_default POWERLEVEL9K_DIR_PATH_SEPARATOR "/"
 set_default POWERLEVEL9K_HOME_FOLDER_ABBREVIATION "~"
-set_default POWERLEVEL9K_DIR_SHOW_WRITABLE false
```
bhilburn: I see you've removed this field, but it's still in use in the code? Was it an accidental delete?

test/powerlevel9k.spec
```diff
@@ -20,7 +20,7 @@ function testJoinedSegments() {
   POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(dir dir_joined)
   cd /tmp
 
-  assertEquals "%K{blue} %F{black}/tmp %K{blue}%F{black}%F{black}/tmp %k%F{blue}%f " "$(build_left_prompt)"
+  assertEquals "%K{blue} %F{black}/tmp%b %K{blue}%F{black}%F{black}/tmp%b %k%F{blue}%f " "$(build_left_prompt)"
```
bhilburn: It looks like you've added bolding to all of the tests, is that right? We need to be sure that we are testing both bold & non-bold options. Really, what we care about is that the core functionality of the segment works, not that it works **bolded**. We can then test bolding separately without affecting all of the segment tests.

onaforeignshore: I've added tests at the end to test the bolding and coloring

onaforeignshore: All strings have `%b` as that doesn't really affect anything except being in the prompt string. Can code a way to remove it, but didn't think it was worth the increased code complexity. What do you think?

bhilburn: Er, I'm confused. Let me ask a different question: why were the previous tests modified to add `%b` to the assertion strings? I must just not be understanding some aspect of this.

onaforeignshore: Bolding is switched on with `%B` and off with `%b`. Since the `%b` doesn't affect the string, I declared `[[ POWERLEVEL9K_segment_BOLD == true ]] && bld='%B' || bld=''` in the code. I didn't do the same for bold off. So, `%b` *always* shows, but `%B` only shows when the segment is bold.

Make sense now?

onaforeignshore:
I have cleaned up the code by adding another bolding variable. This will remove the `%b` from the tests when not needed.

onaforeignshore:
Have replaced `truncatePathFromRight` with `truncatePath`. Have left both in the file for now in case truncatePathFromRight is being used elsewhere / current users

onaforeignshore:
Merged #736

bhilburn:
@onaforeignshore - Done.

PR #760, opened 2018-02-19
Option to Show Node Version Only In Project

weilbith:
As the `vcs` segment only appears in directories which are part of such a repository, it should be possible to set the segment `node_version` to appear only in _NodeJS_ projects.

## Functionality
By using a new variable called `POWERLEVEL9K_NODE_VERSION_PROJECT_ONLY` (which is set to `false` per default), the user can enable the described option. If so, the segment search for a project description file `package.json` from the current working directory upwards to the root. The search stops if a project root has been found. Afterwards the _NodeJS_ version is just requested if this option is disabled or a project is detected.<br>
Using this new variable and set it to `false` per default, the segment stay compatible to older configurations by the user.

## Documenation
The documentation has been already adjusted in the `README`, where a new section for this segment has been opened, the option explained and linked in the list of all segments at the top of the document.

## Testing
It is open to define tests. So far it has been practically tested on _Ubuntu 16.4_ with _oh-my-zsh_. It is not expected, that different OS and/or frameworks affect this behaviour.
This segments has no own test cases yet. To do so, an adjustment of the environment is required, to initialize a plain _NodeJS_ project and check the behaviour in- and outside of this directory.


weilbith:
I've decided to implement such a search functionality, cause I'm not aware of any different option to determine a _NodeJS_ project. Also a quick research do not offer such a possibility. As long as the current working directory is not that much nested, this search loop could be evaluated rly quick and should not represent an performance issue.

powerlevel9k.zsh-theme
```diff
@@ -1023,9 +1023,9 @@ prompt_load() {
   # Replace comma
   load_avg=${load_avg//,/.}
 
-  if [[ "$load_avg" -gt $(bc -l <<< "${cores} * 0.7") ]]; then
```
onaforeignshore: These are changes to `prompt_load` so muddies your PR. Would it be possible to do this in a separate PR if you want to suggest these changes, and also document what the changes are and why?

weilbith: That have to be a slip. Wasn't my intention to touch this. I will revert this immediately.

bhilburn:
Hey @weilbith - this looks great! I apologize for the long delay in responding to this PR. It's a great enhancement to the NodeJS segment, and I'm excited about getting it merged.

I think the point you raised in your initial post is a good one - this segment really needs its own tests, and we currently lack that. Are you willing to implement some test cases as part of this PR, @weilbith? It would be a great improvement for P9k :)

weilbith:
@bhilburn sure, it's my implementation so should provide the tests for it too. :)<br>
What r ur thoughts? What should I add? Simply a new test class in `test/segments/`?

bhilburn:
@weilbith - Yup! I would make a new `spec` file and just fill in some cases that will give us confidence that if someone touches something that impacts the NodeJS segment down the road, we'll catch any breakages before it gets merged.

weilbith:
Sry for beeing idle so long here. 
I just read my first tutorial for `shunit2` and it feels familiar. Just to make this sure: Running the shell unit tests is independent from the whole test-vm/test-bsd topic? A quick search in the whole repo where `shunit` is used make it look like they segment test classes are executed manually only. Did I miss something?

dritter:
> Running the shell unit tests is independent from the whole test-vm/test-bsd topic?

Yes. The test VMs are for manual tests if someone with BSD has a problem.

> A quick search in the whole repo where shunit is used make it look like they segment test classes are executed manually only. Did I miss something?

No. The tests are also executed on every push by Travis-CI. Have a look at `.travis.yml` in the top folder. You need to add your spec there as well.

weilbith:
@dritter
Thanks!
That make sense actually. ^^
Can I directly add the node package to travis environment? Cause it is required for my tests I guess.

weilbith:
That was more "am I allowed to" and not how can I..m

bhilburn:
@weilbith - Yes! You may add Node to travis for your testing =)

dritter:
Hmm. I am opposed to installing node on travis. This would be an integration test. For a proper integration test we need to test different node versions (the node output could change between versions). This would lead to quite some complexity and make the tests more fragile (install X packages in Y different versions on Travis).
My suggestion is to mock node away, as our contract with node is the output (no need for a real node version). As a bonus: It would be easy to mock/test different node versions.
I did a similar thing in the [go test](https://github.com/bhilburn/powerlevel9k/blob/afb7387abec505fb6565d31481c5f5ed724c2f08/test/segments/go_version.spec#L14-L34).

bhilburn:
@dritter is our test / travis / intergration expert, so I would like to go with his guidance, here. Let's do as @dritter suggests =)

weilbith:
@dritter @bhilburn 
Okay, accepted. But one point is left here: The project. Of cause I could do a whitebox test and simply do something like `touch package.json`, cause I now my own written extension is searching for this indicator. But this would not report us, if this doesn't work anymore. So if NodeJS should ever change their project structure definition, we don't get this. But as @dritter said, this lead to the trouble of working with different NodeJS versions and so forth.
So I fake both parts here, the version output and the project itself, where the segment should have an output. Correct?

dritter:
@weilbith You are right. This is the downside of this approach. But if NodeJS chooses a different file name it would be a major BC break for them. I count on the community that would bring that change to us and change the P9K accordingly. And, as said, we would have to support two versions of NodeJS, which would then need a special setup (if we test against real installations).

> So I fake both parts here, the version output and the project itself, where the segment should have an output. Correct?

Yes, you have to fake the node output and the node project (`package.json`). You can see a similar test here: [testTruncateWithPackageNameWorks](https://github.com/bhilburn/powerlevel9k/blob/4c919f1a2bca9d52271edebcbd5090423304afef/test/segments/dir.spec#L201-L231).

weilbith:
Obsolete by more recent PR #1219 since I never made it extend the tests. :skull:

PR #773, opened 2018-03-12
Fixed home abbreviation not being exclusive with circular navigation

conradhaupt:
When navigating to a sub-folder of $HOME and then back up to $HOME, the shortened directory incorrectly contains the shorten delimiter (`cd; cd Documents; cd ..;'` results in `.../~` as the directory shown). This only seems to occur if the shorten strategy environment variable is left blank.

To solve this I added a conditional check to apply the default switch statement case only if the current_path is not $HOME (i.e. not equal to "~").

Just in case this may be more specific to my configuration, my PowerLevel9K config is shown below:
```export DEFAULT_USER=conrad
export POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(context dir_writable dir ssh anaconda vcs)
export POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(status root_indicator time) 
export POWERLEVEL9K_STATUS_VERBOSE=true
export POWERLEVEL9K_SHORTEN_DIR_LENGTH=1
export POWERLVEL9K_BACKGROUND_JOBS_VERBOSE=true```


bhilburn:
@conradhaupt - Thanks so much for taking the time to debug this, implement a solution, and submit it as a PR! Other users were also affected by this bug, and you had to dive right into the meat of one of our most complicated segments to fix this. Thanks so much, and welcome to P9k! `=)`

conradhaupt:
@bhilburn It's an absolute pleasure to help out. powerlevel9k is a fantastic theme and I'm glad I could help resolve the bug.

PR #776, opened 2018-03-17
fix wrong current_path under certain circumstances

### LINKED ISSUE #731: wrong current_path under certain circumstances
Not sure why it sometimes happens and now always, but in a mounted directory i get something that looks like the variable name instead of the actualy pwd content.
No shortening option is enabled.

> ~current_path

actually it should look like this:

> /mnt/usb0/tmp

It has to do with `powerlevel9k.zsh-theme` via https://github.com/bhilburn/powerlevel9k/blob/master/powerlevel9k.zsh-theme#L812
If i remove the -P everything seems to work as expected

anthraxx:
Under certain circumstances like a mounted directory the variable literal of ```current_path``` is printed instead of the actualy pwd content.

-P in this case is wrongly used, what it dies: ```print the arguments to the input of the coprocess.``` which results in printing ```~current_path```.

This fixes #731

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -809,7 +809,7 @@ prompt_dir() {
         current_path="${cur_short_path: : -1}"
       ;;
       *)
-        current_path="$(print -P "%$((POWERLEVEL9K_SHORTEN_DIR_LENGTH+1))(c:$POWERLEVEL9K_SHORTEN_DELIMITER/:)%${POWERLEVEL9K_SHORTEN_DIR_LENGTH}c")"
+        current_path="$(print "%$((POWERLEVEL9K_SHORTEN_DIR_LENGTH+1))(c:$POWERLEVEL9K_SHORTEN_DELIMITER/:)%${POWERLEVEL9K_SHORTEN_DIR_LENGTH}c")"
       ;;
     esac
   fi


anthraxx:
This apperantly make a check fail, but i can't really spot why and what to adjust.
Maybe there is a side effect? However this -P definitivly breaks stuff so maybe something else has to be fixed too to make the test happy.
Some insights appreciated.

dritter:
Hmm. `man zshall` says for `print -P`:
> Perform  prompt  expansion  (see  EXPANSION OF PROMPT SEQUENCES in zshmisc(1)).  In combination with `-f', prompt escape sequences are parsed only within interpolated arguments, not within the format string.

And this is what the tests secures.

anthraxx:
Maybe, but it makes printing current directory fail for the case I described making it not show up at all. The other code paths don't contain this option so it makes me wonder what needs adjustment.

dritter:
Back in the beginning of the `prompt_dir` segment there were only a couple of truncation strategies. The default strategy was to truncate directories from left. This strategy is in charge, if `POWERLEVEL9K_SHORTEN_DIR_LENGTH` is set and `POWERLEVEL9K_SHORTEN_STRATEGY` is any other than the known ones.
What I want to say by that is, if you have set a length to truncate, P9K automatically thinks that you want to truncate the current path. That is how to get into that code path.

The output of all shorten strategies is captured in `current_path` for further processing. This includes omitting the first charater (the shorten delimiter) if the user wants to. This is what the test does: Changing the delimiter and omitting the first character. You can see the setup [here](https://github.com/bhilburn/powerlevel9k/blob/master/test/segments/dir.spec#L348-L371). The original PR for that is https://github.com/bhilburn/powerlevel9k/pull/394

anthraxx:
@dritter thanks for explaining, now the 1000 dollar question: How to fix the issue? :yum: 

In my case there is no shortening happening (path shorter then 100):
```
POWERLEVEL9K_SHORTEN_DIR_LENGTH=100
POWERLEVEL9K_SHORTEN_STRATEGY is not set
```

The result should definitivly not be a literal ```~current_path``` i think we can agree on that :joy_cat:

dritter:
`~current_path` looks definitely not right. :D

I think we still don't know enough about what happens in the code.
First of all, setting `POWERLEVEL9K_SHORTEN_DIR_LENGTH=100` to disable truncation seems not right. As said, setting the length, but not `POWERLEVEL9K_SHORTEN_STRATEGY` still triggers left truncation. So, it would be better do unset `POWERLEVEL9K_SHORTEN_DIR_LENGTH` completely.

What happens in the `prompt_dir` segment is that we construct a string that gets evaluated by ZSH: `'%101(c:…/:)%100c'`, which means to truncate 100 directories from left.

But how to debug it:
At first I would check if there is a general issue with ZSH settings by calling the command manually: `print -P '%101(c:…/:)%100c'`. This should work. If not, there is some other problem (like some special ZSH settings, or aliasing `print`).
This does not explain why an internal variable name is exposed, so to debug this I usually do:
1. Open a new shell and navigate to the right directory
2. Only use the problematic segment: `POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(dir); POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=()`
3. Trace your shell: `setopt xtrace` (disable later with `unsetopt xtrace`
With the step 3 you can follow the code through the execution. This should bring more light in the dark. It would be nice, if you could paste the last lines that start with `prompt_dir` here.

bhilburn:
Hey @anthraxx! Just checking in on this! I do really want to get this fix merged once we get it sorted out `=)`. Did you get a chance to try @dritter's advice?

fikovnik:
I'm also affected by this issue - in my case it just happens somewhat randomly - not connected to a mount. For example if I cd into any directory under `/var/lib` which is not owned by root I also see `~current_path` (is it somehow related to different user/group).

This fix (removing `-P` at line 941) fixes it for me.

PR #781, opened 2018-03-21
Fix Brightness Table

nikeee:
The markdown table seems to be broken. This PR fixes that.


bhilburn:
@nikeee - You are indeed correct! Our docs are really important, and this is an important fix - even if the changeset seems minor. Thanks so much for taking the time to put together this PR and submit it, and welcome to P9k! `=)`

Issue #711, opened 2017-12-28
Change term_colors logic?


tonylambiris:
Running into a weird issue on a FreeBSD 8.3 box I have a shell account on:
```
 export TERM=xterm-256color; echotc Co; tput colors
8
167
 ./colors.sh 
 export TERM=vt220; echotc Co; tput colors         
echotc: no such capability: Co
167
```
Not sure why zsh doesn't report a color capability when set to vt220 (the default if TERM isn't defined in the environment). Also not sure why it's reporting 167 colors for both termcaps. I don't have admin access so I can't do much in the way of digging much deeper into the system libraries and installed versions. As a fix would it make sense to use the following logic instead?

```
term_colors=0

if which tput &>/dev/null; then
	term_colors=$(tput colors)
else
	term_colors=$(echotc Co)
fi

if (( term_colors <= 16 )); then
    print -P "%F{red}WARNING!%f Your terminal isn't setup to use more than 16 colors!"
    print -P "You should put: %F{blue}export TERM=\"xterm-256color\"%f in your \~\/.zshrc"
    print -P "_before_ loading this theme in your \~\/.zshrc. In most terminal emulators, putting"
    print -P "%F{blue}export TERM=\"xterm-256color\"%f at the top of your \~\/.zshrc is sufficient."
fi
```

FWIW here is some version data and screenshots of running a local `colorgrid` function on both my local Linux box and the FreeBSD server:
```
 uname -a     
FreeBSD phalse.2600.COM 8.3-RELEASE-p3 FreeBSD 8.3-RELEASE-p3 #0: Mon Jun 11 23:52:38 UTC 2012     root@i386-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  i386
 zsh --version
zsh 4.3.17 (i386-portbld-freebsd8.3)
```

**Linux colorgrid**
![colorgrid-linux](https://user-images.githubusercontent.com/14058972/34413434-8d3159a2-ebb1-11e7-97d2-fc6f6d5366a7.png)

**FreeBSD colorgrid**
![colorgrid-freebsd](https://user-images.githubusercontent.com/14058972/34413438-915487a2-ebb1-11e7-8676-a1ba8a16fe7c.png)

bhilburn:
Hey @tonylambiris! Congratulations on being our first FreeBSD user to report an issue! If I could award badges on Github, I totally would, hah `=)`

[...]

PR #783, opened 2018-03-24
Added a segment that shows the Java version.

VanRoy:


bhilburn:
Hey @VanRoy! Thanks so much for the segment submission! This is a good one - we are rather short on Java segments!

onaforeignshore:
Already merged this into #771 - [5843e80](https://github.com/bhilburn/powerlevel9k/pull/771/commits/5843e804dd2c4f3f0bd6d6bb5a01891927d66b5a)

PR #789, opened 2018-03-26
workaround for ZSH status behavior

### LINKED ISSUE #749: `status` segment behavior with [[
`[[` doesn't fire the `status` segment:

```
 ❯ [[ 1 == 1 ]]                                                                                                        11:10:11  02.14.18  48% (2:09) ▅
 ✔  me@host  ~ 
 ❯ [[ 1 == 2 ]]                                                                                                        11:10:15  02.14.18  48% (2:09) ▅
 ✔  me@host  ~ 
 ❯ [[ 1 == 1 ]] ; echo $?                                                                                              11:10:18  02.14.18  48% (2:09) ▅
0
 ✔  me@host  ~ 
 ❯ [[ 1 == 2 ]] ; echo $?                                                                                              11:10:27  02.14.18  48% (2:09) ▅
1
 ✔  me@host  ~ 
 ❯
```

Also if you force a non-zero return with something else (`[` in this case) then future calls to true (zero  exit) calls of `[[` retain the previous status return code:

```
 ✔  me@host  ~ 
 ❯ [ 1 = 2 ]                                                                                                           11:14:13  02.14.18  47% (2:39) ▅
 ↵ 1  me@host  ~ 
 ❯ [[ 1 == 1 ]]                                                                                                        11:14:18  02.14.18  47% (2:39) ▅
 ↵ 1  me@host  ~ 
 ❯ [[ 1 == 1 ]]                                                                                                        11:14:25  02.14.18  47% (2:39) ▅
 ↵ 1  me@host  ~ 
```

The odd thing is that `echo $?` returns the correct exit code when you use `[[` but the status segment doesn't seem to see it.

docwhat:
`$pipestatus` is returning all zeros when using `[[  ]]` expressions
that are false.

This works around it by using `$status` (A.K.A. `$?`) when `$pipestatus`
has only 1 items.

Fixes #749

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1163,8 +1163,13 @@ prompt_status() {
   local ec
 
   if [[ $POWERLEVEL9K_STATUS_SHOW_PIPESTATUS == true ]]; then
-    ec_text=$(exit_code_or_status "${RETVALS[1]}")
-    ec_sum=${RETVALS[1]}
+    if (( $#RETVALS > 1 )); then
+      ec_text=$(exit_code_or_status "${RETVALS[1]}")
+      ec_sum=${RETVALS[1]}
+    else
+      ec_text=$(exit_code_or_status "${RETVAL}")
+      ec_sum=${RETVAL}
+    fi
 
     for ec in "${(@)RETVALS[2,-1]}"; do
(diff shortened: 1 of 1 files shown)


bhilburn:
Thank you very much, @docwhat! I really appreciate the discussion and testing on this one in #749, and you taking the time to submit the fix. Another superb PR `=)`

PR #791, opened 2018-03-27
Make variables local

dritter:
This is a rather small change. I recently stumbled upon the option `setopt WARN_CREATE_GLOBAL` that 🙄 warns about creating global variables. With this PR I made all unnecessary global variables local.

### DIFF
--- a/functions/colors.zsh
+++ b/functions/colors.zsh
@@ -35,11 +35,11 @@ function getColor() {
     fi
   else
     # named color added to parameter expansion print -P to test if the name exists in terminal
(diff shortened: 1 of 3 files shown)


AtomicCoding:
@dritter Ok thanks! Sorry!

dritter:
No need to apologize. It is a valid question ;)

bhilburn:
@dritter - Oh, awesome! Thanks for having a look at this and taking the time to fix this. Another good PR `=)`

Issue #735, opened 2018-02-03
Readme for shorten strategy for dir segment is ambiguous


dualscyther:
The current README says that the default value for `POWERLEVEL9K_SHORTEN_STRATEGY` is "None". It also says that the `POWERLEVEL9K_SHORTEN_DIR_LENGTH` is 2.
In the table which defines which each of the different strategies do, the default strategy is listed as "truncate whole directories from left".

The default behaviour could be interpreted as either:
- truncate the path such that it only shows 2 directories.
- don't perform any truncation

Currently, the latter option is what happens (which I agree is good default behaviour). What actually happens is that there is no path shortening unless:
- `POWERLEVEL9K_SHORTEN_STRATEGY` is set
  - Then `POWERLEVEL9K_SHORTEN_DIR_LENGTH` defaults to 2
- `POWERLEVEL9K_SHORTEN_DIR_LENGTH` is set
  - Then `POWERLEVEL9K_SHORTEN_STRATEGY` defaults to .. default (truncate directories)

This is a bit unintuitive.. A simple fix would just be to actually provide a `none` option and a `truncate_folder` option - rather than needing to set `POWERLEVEL9K_SHORTEN_DIR_LENGTH` in order to experience the default `truncate_folder` behaviour.

bhilburn:
Hey @dualscyther - Thanks so much for the feedback! I can see how the current docs might create confusion.

I agree, adding a `truncate_folder` option rather than `default` is more intuitive, and makes the `SHORTEN_DIR_LENGTH` field more immediately understandable. This is a fairly easy fix, as we can create the new option and ghost default to it, thus preserving backwards compatibility with configs.

Are you up for creating a PR for this?

(My apologies for the long delay in getting back to you, by the way!)

dualscyther:
Sure

bhilburn:
Closing this in favor of tracking it in #793

PR #793, opened 2018-03-31
Improve truncation strategy API

dualscyther:
Do not merge this yet as it is just a change to the README and is not actually implemented, this is just an RFC at the moment. This is in response to #735 

Breaking changes:

* Users that have only set `POWERLEVEL9K_SHORTEN_DIR_LENGTH` without setting `POWERLEVEL9K_SHORTEN_STRATEGY` will no longer experience a default strategy of directory truncation from the left. Instead, they must explicitly set `POWERLEVEL_9K_SHORTEN_STRATEGY` to `truncate_dirs`.

cc:
@bhilburn

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,20 +368,21 @@ Customizations available are:
 
 | Variable | Default Value | Description |
 |----------|---------------|-------------|
-|`POWERLEVEL9K_DIR_PATH_ABSOLUTE`|None|If set to `true`, will use absolute paths instead of home folder abbreviation `~`|
+|`POWERLEVEL9K_SHORTEN_STRATEGY`|See below|How the directory strings should be truncated. No truncation occurs by default. See the table below for more information.|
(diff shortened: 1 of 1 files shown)


bhilburn:
I like this approach. It simplifies the configuration, I think, and makes it more clear which options affect which behaviors.

@dritter - What are your thoughts, here?

bhilburn:
Hey @dualscyther! Were you interested in finishing up this PR and getting it merged! It would be great to have it merged into `next` for our next release. And, if we are going to break something, now is the time to do it.

@dritter - Thoughts?

dualscyther:
Hey @bhilburn I'm down to finish this PR, when is the next planned release? Sorry for the late reply :(

bhilburn:
Hey @dualscyther! Any chance you're still up for taking this on? It would be great to get this update made `=)`

dualscyther:
Hey, yes! I've just been a little busy with studies but I'll try to get on this :)

dualscyther:
@bhilburn Sorry I'm not actually sure where to make this change as it seems things have moved around a little bit. I assume I'm meant to be looking at the `next` branch?

dritter:
Yes, the `next` branch is the right branch to look at.

dualscyther:
@bhilburn  Might have to put this on hold unless someone can take over, as I've recently become a bit busy :(

PR #797, opened 2018-04-11
Fix wrapper slowness in vcs segment

guicassolato:
This prevents `git` command from passing through eventual aliases, commonly defined when Git wrappers (e.g. hub) are used.

### DIFF
--- a/functions/vcs.zsh
+++ b/functions/vcs.zsh
@@ -16,8 +16,8 @@ function +vi-git-untracked() {
       FLAGS+='--ignore-submodules=dirty'
     fi
 
-    if [[ $(git rev-parse --is-inside-work-tree 2> /dev/null) == 'true' && \
-            -n $(git status ${FLAGS} | grep -E '^\?\?' 2> /dev/null | tail -n1) ]]; then
+    if [[ $(command git rev-parse --is-inside-work-tree 2> /dev/null) == 'true' && \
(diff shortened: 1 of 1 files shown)


dritter:
Thanks @guicassolato for this PR (and congrats for the 400th PR ;) ).
I think this PR deserves a bit more explanation.

From the [docs](http://zsh.sourceforge.net/Doc/Release/Shell-Builtin-Commands.html#Shell-Builtin-Commands):
> command [ -pvV ] simple command
> The simple command argument is taken as an external command instead of a function or builtin and is  executed. If the POSIX_BUILTINS option is set, builtins will also be executed but certain special properties of them are suppressed. The -p flag causes a default path to be searched instead of that in $path. With the -v flag, command is similar to whence and with -V, it is equivalent to whence -v.

So the `command` builtin skips possible wrapper scripts and executes the command directly. If I have an alias for `git` (e.g. `alias git='echo "git-wrapper"; git'`; stupid, I know) and I execute `git status` in a repository, I'll get `git-wrapper` besides the git status output. But if I do `command git status` I'll get only the output of that command (without `git-wrapper`).

bhilburn:
This is excellent, @guicassolato! This will certainly improve a number of issues, including extremely confusing bug reports as well as likely speeding some things up. And thanks, @dritter, for explaining how it works!

AtomicCoding:
Can this be closed?

PR #798, opened 2018-04-12
Add vcs branch name truncating.

floor114:
These changes allow truncating of vcs branch name.

Before:
image

After:
image

@bhilburn, please take a look.


bhilburn:
@floor114 - This is excellent, and especially given @dritter's approval, I'm eager to get it merged!

PR #800, opened 2018-04-13
Fix table formatting in README

lifehackett:


bhilburn:
@lifehackett - Thanks so much for the fix! Clean docs are really important, and while this changeset is small, it's meaningful. Thanks for taking the time to submit this, and welcome to `P9k`!

PR #806, opened 2018-04-22
Edited virtualenv prompt to show python version

hiimdoublej:
Edited the prompt for virtualenv to always show current python version
If inside a virtualenv, it would show <virtualenv_name>:<python_version>
ex: django:3.6.0
if not inside any virtualenv, it would show <python_version>
ex: 3.6.0

Since I just recently adopted python3 from python2, I'm switch python versions often to maintain older projects. I would sometimes forget which version I'm at, but the pyenv prompt would only display versions set by `pyenv shell <version>`. I would have to type `python -V` to check for myself.
So I integrated persistent python version displaying along with the `virtualenv` prompt to make `virtualenv` prompt display python version as well ! I thought this could help some others, hence this pull request.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1385,10 +1385,11 @@ prompt_vi_mode() {
 prompt_virtualenv() {
   local virtualenv_path="$VIRTUAL_ENV"
   if [[ -n "$virtualenv_path" && "$VIRTUAL_ENV_DISABLE_PROMPT" != true ]]; then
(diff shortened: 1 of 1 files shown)


sambadevi:
Maybe add a modifier so that the old behaviour (segment only shows while in pyenv) is still accessible. I can imagine a lot of people like it that they only see the pyenv information if they are actually in a virtualenv

hiimdoublej:
Alright, I'm closing this one, I think a better approach to this problem is to add a new prompt elem where it will eval '''python -V''' so that the theme itself will be more customisable. I'll close this and open a new PR. Thanks @sambadevi

PR #808, opened 2018-04-23
Add new prompt: pyver

hiimdoublej:
Added new prompt ```pyver```, this prompt element will evaluate current python
version every prompt, so it will always display current working python
version, unlike the pyenv prompt where it would only show versions
exported to the ```$PYENV_VERSION``` variable. By adding this prompt element,
people like me can feel safe by having current python version printed on the
prompt 24/7 :D


sambadevi:
@hiimdoublej I don’t think the drink water prompt should be added to this PR (even though I like the idea 😄)

Could you create a new branch for the pyver segment and base the PR on that branch?

Just so that new commits on your master branch won’t conflict or get included into this PR

sambadevi:
Also, I think this change should go into `next` instead but I think that's up to @bhilburn

hiimdoublej:
Oops....the drink water prompt was not supposed to be here....LOL
Let me fix this.

hiimdoublej:
@sambadevi I've separated the two prompts in two branches, also reverted my previous commit, so this branch only contains the ```pyver``` prompt, thanks for the tip :)

sambadevi:
@hiimdoublej I don't seem to have the `pyver` plugin installed by default. Could you tell me the advantages of using `pyver` instead of using this code snippet:

```
#!/usr/bin/env zsh
if [ -v 'VIRTUAL_ENV' ]; then
 echo "$(python -c 'import sys; version=sys.version_info[:3]; print("{0}.{1}.{2}".format(*version))')"
fi
```
It checks if you're in a venv and prints the current python version. (Works with `python2` and `python3`)

If you use it as a custom segment in your `.powerlevelrc` (`.zshrc` or anywhere):
```
POWERLEVEL9K_CUSTOM_PYVER='~/.dotfiles/bin/pyver'
POWERLEVEL9K_CUSTOM_PYVER_BACKGROUND='107'
POWERLEVEL9K_CUSTOM_PYVER_FOREGROUND='white'
```

It looks like this:
Using Python2
Imgur
Using Python3
Imgur

hiimdoublej:
I think that the main difference between your code and mine is that, I personally uses ```pyenv``` to manage and switch between python versions, hence why I use ```pyenv version``` to output my current version and extract current python version from there.  
 I think your code is better in terms of evaluating current python version, It's a more straight up approach compared to mine ! With that in mind, would you want to submit your own pull request ? Because I'm thinking maybe I should close this PR and submit a new one with my code under the ```pyenv``` prompt, since my code evaluates current version using``` pyenv```.

Issue #747, opened 2018-02-14
Carrier return in promt line when Anaconda environment is active


sbmueller:
When I activate a virtual Anaconda environment, it's name is displayed in my promt. It seems since a few days, there is a carrier return at the end of the virtual environment name, which yields all elements right of the VE name to be displayed at the beginning of the row. Any ideas how that could happen? I'm on Windows 10.
![image](https://user-images.githubusercontent.com/7465688/36195579-3ca6d614-116e-11e8-869e-669f9c2fe7cb.png)

bhilburn:
Hey @sbmueller! Sorry for the delayed response, here.

And, uh, whoa. That's not great. Did this start happening after an update to P9k, or was it using an existing config / P9k version?

bhilburn:
@sbmueller - Whoa, crap. Would you mind bisecting to nail down the commit?

sbmueller:
Hmm I'm unlucky. When checking out a commit somewhere around early 2017, I even get a nasty conda error when using `source activate`:
```
 Traceback (most recent call last):
      File "C:\app\anaconda3\lib\site-packages\conda\cli\main.py", line 98, in main
        return activator_main()
      File "C:\app\anaconda3\lib\site-packages\conda\activate.py", line 581, in main
        print(activator.execute(), end='')
      File "C:\app\anaconda3\lib\encodings\cp1252.py", line 19, in encode
        return codecs.charmap_encode(input,self.errors,encoding_table)[0]
    UnicodeEncodeError: 'charmap' codec can't encode characters in position 16-17: character maps to <undefined>
```

bhilburn:
Errrr no. Definitely not. @sbmueller - What commit triggers the error? And the one right before it works fine?

I believe @dritter is our dev most familiar with the conda work. Any thoughts, here?

sbmueller:
Oh sorry, I wasn't precise. *All* the commits back to early 2017 have the carrier return problem. If I go back as far as early 2017 (052eba9c74b2153dbd70eb4ba829f258a596baf2 for instance) I even get the conda error message. So there is no commit that I can check out that works fine for me.
I think the problem is not within p9k, but if you @bhilburn or @dritter have any ideas what external impact can cause this, I'm happy to hear about it!

dritter:
> If I go back [...] I even get the conda error message.

Does that mean your conda variables `$CONDA_ENV_PATH$CONDA_PREFIX` contain a stack trace? If not, do they contain a line break?
[...]

PR #810, opened 2018-04-23
Fix newline issues in the anaconda segment

dritter:
This is a PR for #747 

It might be that the directory after `proxy` starts with an "n". This could lead to a newline as the string gets evaled..

@sbmueller Could you check if this branch works for you?
It is not a super good solution, because it just strips the newlines, but if the path contains backslashes it will still be broken eventually.. It might be a better idea to replace backslashes with slashes. What do you think @sbmueller ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,3 +29,4 @@ script:
   - test/segments/go_version.spec
   - test/segments/vcs.spec
   - test/segments/kubecontext.spec
+  - test/segments/anaconda.spec
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -281,7 +281,7 @@ prompt_anaconda() {
   # Caret notation: ^J is Newline (LF), ^M is Carriage Return (CR),
   # ^G is Bell (BEL), ^I is horizontal tab (TAB)
   # see https://en.wikipedia.org/wiki/ASCII#ASCII_control_characters
-  local SANITIZATION_CHARACTER_GROUP="[\^J\^M\^G\^I]"
+  local SANITIZATION_CHARACTER_GROUP="[\^J\^M\^G\^I"$'\n\r\a\t'"]"
   local _path="${CONDA_ENV_PATH//${~SANITIZATION_CHARACTER_GROUP}/}${CONDA_PREFIX//${~SANITIZATION_CHARACTER_GROUP}}/"
 
   if ! [ -z "$_path" ]; then
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -278,7 +278,7 @@ CURRENT_BG='NONE'
 prompt_anaconda() {
   # Depending on the conda version, either might be set. This
   # variant works even if both are set.
-  local _path=$CONDA_ENV_PATH$CONDA_PREFIX
+  local _path="${CONDA_ENV_PATH//\\n/}${CONDA_PREFIX//\\n/}"
   if ! [ -z "$_path" ]; then
     # config - can be overwritten in users' zshrc file.
     set_default POWERLEVEL9K_ANACONDA_LEFT_DELIMITER "("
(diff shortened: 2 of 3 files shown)


sbmueller:
@dritter thank you! This actually did something, but not solve the issue. There is now a newline after the environment name instead of a carrier return, meaning the end of the first line is displayed at the beginning of the second, beyond the cursor. I think if you regard carrier returns as well in your PR, this could work?

sbmueller:
I could fix it by using this answer https://stackoverflow.com/a/19347380
`${CONDA_ENV_PATH//[$'\t\r\n']}${CONDA_PREFIX//[$'\t\r\n']}`

dritter:
Good to know @sbmueller 
Initially I tried it with a character group as well (`[\\r\\n]`; double escaped here), but it didn't work. Neither did [ANSI-C](https://www.gnu.org/software/bash/manual/bash.html#ANSI_002dC-Quoting) style quoting work (`[$'\r\n']`). What worked was using the [Caret notation](https://en.wikipedia.org/wiki/Caret_notation) (`[\^M\^N]`). I am not sure it this is because I just execute tests (maybe shunit transforms the control characters), or if we are on different operating systems..

Could you verify that the caret notation works on your machine as well? That would be great!

sbmueller:
@dritter sorry, caret notation does not work for me... guess I'll just fork and use an own branch since I've found a personal solution.

dritter:
@sbmueller I think it is a good idea to have some sanitization in the core. And being a regular expression character class we could check for both types (whatever the difference is in ZSH).

Could you check my latest commit? That's hopefully the last thing I request from you - Sorry for all the noise.

sbmueller:
Sorry @dritter I think you're cutting away too much now. When no virtual environment is activated, the anaconda segment still displays `(/)` as environment name. When I `source activate base`, instead of `(anaconda3)`, the name is displayed as `(appanaconda3)` where `app` is the parent folder of my anaconda3 installation.

bhilburn:
So, by my read of this, this PR solves the original problem that @sbmueller had in #747, but effectively un-conditionalizes the segment. Is that right?

sbmueller:
@bhilburn on my (maybe special) setup, yes. I'm fine if we close the issue since it's solved for me now with my custom solution.

bhilburn:
@dritter - Do you think this issue needs a general fix? @sbmueller's issue was the first we've heard of it, if I'm not mistaken?

dritter:
@bhilburn Well. Maybe. 😆 I think it is mostly a good idea to trim special characters from input strings (pretty sure we never want to print a `BELL` sign..). What makes this task quite difficult is the different character sets and different character notations (that seems to make a difference). Question remains how we could generalise this.

@sbmueller Could you try again? Fingers crossed that I got it right this time. :) I now changed that `basename` is invoked before trimming the special characters..
 Btw. Could you do a quick check how `basename` behaves on windows? Does `basename "C:\app\anaconda3\lib\site-packages\redmond\newton\conda\cli"` output `cli`?
I think my tests are a bit off, because I test windows paths on mac with a unix basename.. The tests now ensure that not special character leak through the prompt, but the actual behaviour of the segment is to trim the path as well..

bhilburn:
@dritter @sbmueller - What's the status of this one? If it's not going to make it into the v0.7.0 codebase, I'd like to go ahead and close this one out.

sbmueller:
My custom solution fixed it for me:

> I could fix it by using this answer https://stackoverflow.com/a/19347380
> `${CONDA_ENV_PATH//[$'\t\r\n']}${CONDA_PREFIX//[$'\t\r\n']}`

Unfortunately, @dritter s code did never work for me on my machine. I would suggest to just close this since I'm happy and nobody else is experiencing this issue :)

bhilburn:
Okay, sounds good. Sorry for the headache, @sbmueller!

@dritter - I'm going to close this one out. If there is anything you want to capture here and pull in, feel free to re-open! =)

cgjosephlee:
`$CONDA_ENV_PATH` is deprecated and replaced by `$CONDA_DEFAULT_ENV` and `$CONDA_PREFIX` (see https://github.com/conda/conda/issues/2312).
Also as referred in [Anaconda 4.4.0 changelog](https://github.com/conda/conda/blob/master/CHANGELOG.md#440-2017-12-20), activate a default `base` environment when start up is recommended. Here I suggest to add a variable `set_default POWERLEVEL9K_ANACONDA_BASE "base"` to let users can hide `base` because it is alway activated.

PR #811, opened 2018-04-24
Update default public ip file name

shirts:
The default value for Public IP file in the readme was referencing `/tmp/p8k_...` rather than `/tmp/p9k_...`


dritter:
Good catch! Thx @kevinchandler

shirts:
Hopefully it's not the last 😉

PR #816, opened 2018-04-30
Add chruby options

opes:
Refactors and adds options for the chruby segment. Documentation updated accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,17 @@ Note that you can [modify the `_FOREGROUND`
 color](https://github.com/bhilburn/powerlevel9k/wiki/Stylizing-Your-Prompt#segment-color-customization)
 without affecting the icon color.
 
+##### chruby
+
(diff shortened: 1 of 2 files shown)


bhilburn:
@opes - Wow, this is a really well-done PR. Based on your use of P9k's internal functions and adherence to coding guidelines, I wouldn't have guessed that this was your first PR. Nice work!

As a new feature, we need to merge this into `next` rather than `master` to keep things stable. Unfortunately, I can't re-target this PR because it would pull in a bunch of commits from `master`. So, I tried doing the `master` merge manually first, and actually triggered several conflicts, at least one of which is in the Ruby segment code.

So, anyway, this is a long way of saying I need to sort this out, and then will merge this into `next`, hah.

opes:
Thanks for following up, @bhilburn! Let me know if there's anything I need to do on my end to assist.

PR #818, opened 2018-05-01
Add a Laravel version segment

dritter:
This adds a new Laravel version segment.

I always wanted to have support for Laravel. :)
But, not being a user of it, I would be glad if someone else could confirm that this works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,3 +29,4 @@ script:
   - test/segments/go_version.spec
   - test/segments/vcs.spec
   - test/segments/kubecontext.spec
+  - test/segments/laravel_version.spec
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## next
+
+### New Segment: `laravel_version`
+
+Displays the current laravel version.
+
 ## v0.6.4
 
 - `load` segment now has configurable averages.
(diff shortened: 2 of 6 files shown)


brandon14:
@dritter If this is the same code that is in the master branch, when activated on a directory that does not contain artisan (i.e. artisan is not installed globally and you are not inside the root Laravel project folder), it displays an error because calling `php artisan --version 2>/dev/null` will still allow the stdout to escape the `/dev/null` redirect and if php does not find artisan, it will output `Could not open input file: artisan` as the version. Maybe redirecting both stdout and stderr to `/dev/null` would work. Something like:

```bash
prompt_laravel_version() {
  local laravel_version="$(php artisan --version > /dev/null 2>&1)"
  if [[ -n "${laravel_version}" ]]; then
    # Remove unrelevant infos
    laravel_version="${laravel_version//Laravel Framework version /}"

    "$1_prompt_segment" "$0" "$2" "maroon" "white" "${laravel_version}" 'LARAVEL_ICON'
  fi
}
```

dritter:
Hi @brandon14 !

Thanks for your feedback. This is especially useful, as this segment is already in the latest release, and I couldn't test it. Do you mind creating a new pull request against `master`? That would be awesome.

@bhilburn the fix should be included in v0.6.6

brandon14:
@dritter I have a PR up here for it in master: https://github.com/bhilburn/powerlevel9k/pull/912

dritter:
This PR is alread in `master` and @brandon14 already provided some fixes. Closing.

PR #821, opened 2018-05-03
add `POWERLEVEL9K_VIRTUALENV_SHOW_VERSION` option to virtualenv segment

sambadevi:
As discussed in #808 
This PR adds a new option to the `virtualenv` segment.

If you add `POWERLEVEL9K_VIRTUALENV_SHOW_VERSION=true` to your `.zshrc` p9k will add your python version to your venv segment

It looks like this:
Using Python2
Imgur
Using Python3
Imgur

Thanks @hiimdoublej for suggesting this!

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1570,6 +1570,10 @@ prompt_virtualenv() {
   local virtualenv_path="$VIRTUAL_ENV"
   if [[ -n "$virtualenv_path" && "$VIRTUAL_ENV_DISABLE_PROMPT" != true ]]; then
     "$1_prompt_segment" "$0" "$2" "blue" "$DEFAULT_COLOR" "$(basename "$virtualenv_path")" 'PYTHON_ICON'
+
+    if [[ -v "POWERLEVEL9K_VIRTUALENV_SHOW_VERSION" && "$POWERLEVEL9K_VIRTUALENV_SHOW_VERSION" == true ]]; then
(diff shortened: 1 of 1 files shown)


dritter:
Thanks @sambadevi for this PR!

To be honest, I am a bit ambivalent with this PR. On one hand 👍 for creativity (I personally never thought about rendering another segment within one). But on the other side the segments are quite independent from each other, right? The Version segment is always shown if the variable is true.. So, if there would be a `python_version` segment that prints out the version, we would get the exact same result (if the colors match and the icon would be set to empty, of course), and it would be combinable with other python env segments..

sambadevi:
So @dritter your suggestion is creating a new segment `python_version` and wait for #771 to add conditionals? I can change this PR to be a separate segment, the code wouldn’t differ that much. But there won’t be any segment combination unless #771 gets merged.

dritter:
For the records: The successor of #771 is merged and it is possible to define own conditions for every segment. The condition should look like this `P9K_PYTHON_VERSION_CONDITION='[[ -n "${VIRTUAL_ENV}" ]]'`. Yes, this is basically a duplication of the condition in the `virtualenv` segment; that is the trade-off..
That a separate `python_version` segment makes sense is shown in #1053 .

bhilburn:
Per @dritter, the project's structure is completely changed in the `next` branch. It should be pretty easy to port this change over, though!

sambadevi:
I am closing this PR and will port this change to the new structure in `next`

PR #826, opened 2018-05-05
Fixed prompt_rust_version to avoid grep aliases

### LINKED ISSUE #803: rust_version doesn't display anything
I added `rust_version` to my `POWERLEVEL9K_LEFT_PROMPT_ELEMENTS` list but nothing appeared.

Looking at https://github.com/bhilburn/powerlevel9k/blob/master/powerlevel9k.zsh-theme#L1107 I see the command used but this doesn't seem to do anything.

Similar commands:
```shell
$ rustc --version
rustc 1.24.0-nightly (b98fd524e 2018-01-05)
$ rustc --version 2>&1 | grep -oe "^rustc\s*[^ ]*"
rustc 1.24.0-nightly
$ rustc --version 2>&1 | grep -oe "^rustc\s*[^ ]*" | grep -o '[0-9.a-z\\\-]*$'

```

Seems like the regex is off in the final grep?

TheDan64:
Fixes #803

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1104,7 +1104,7 @@ prompt_root_indicator() {
 # Print Rust version number
 prompt_rust_version() {
   local rust_version
-  rust_version=$(command rustc --version 2>&1)
+  rust_version=$(command rustc --version 2>/dev/null)
   # Remove "rustc " (including the whitespace) from the beginning
   # of the version string and remove everything after the next
   # whitespace. This way we'll end up with only the version.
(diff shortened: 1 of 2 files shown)


TheDan64:
Thanks for the review; I silenced the missing rustc error and tried to fix the test

dritter:
Hah. 🤦‍♂️ In our test we mock `rustc`. In the new code we avoid any wrapper scripts and call `command rustc`. So we need to find a good way to mock rustc anyways..

dritter:
Just for the record: This PR is related to #632

PR #830, opened 2018-05-08
add new option "POWERLEVEL9K_BATTERY_HIDE_ABOVE_THRESHOLD"

sambadevi:
Hides the battery segment if percentage hits 100%

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -487,7 +487,7 @@ prompt_battery() {
       [[ "${(t)POWERLEVEL9K_BATTERY_STAGES}" =~ "array" ]] && POWERLEVEL9K_BATTERY_ICON="$POWERLEVEL9K_BATTERY_STAGES[$offset]" || POWERLEVEL9K_BATTERY_ICON=${POWERLEVEL9K_BATTERY_STAGES:$offset:1}
     fi
   fi
-
+  # return if POWERLEVEL9K_BATTERY_HIDE_ABOVE_THRESHOLD is set and the battery percentage is greater or equal
   if [[ -v "POWERLEVEL9K_BATTERY_HIDE_ABOVE_THRESHOLD" && "${bat_percent}" -ge $POWERLEVEL9K_BATTERY_HIDE_ABOVE_THRESHOLD ]]; then
     return
   fi
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -488,28 +488,19 @@ prompt_battery() {
     fi
   fi
 
-  if [[ -v "POWERLEVEL9K_BATTERY_HIDE_FULL" && "$POWERLEVEL9K_BATTERY_HIDE_FULL" == true ]]; then
-      if [[ "$bat_percent" != "100" ]]; then
-          if [[ -n "$POWERLEVEL9K_BATTERY_LEVEL_BACKGROUND" ]] && [[ "${(t)POWERLEVEL9K_BATTERY_LEVEL_BACKGROUND}" =~ "array" ]]; then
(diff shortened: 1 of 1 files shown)


sambadevi:
If you add `POWERLEVEL9K_BATTERY_HIDE_FULL=true` to your p9k config the battery segment will be hidden if the percentage is equal to 100%

powerlevel9k.zsh-theme
```diff
@@ -488,14 +488,27 @@ prompt_battery() {
     fi
   fi
 
-  # override the default color if we are using a color level array
-  if [[ -n "$POWERLEVEL9K_BATTERY_LEVEL_BACKGROUND" ]] && [[ "${(t)POWERLEVEL9K_BATTERY_LEVEL_BACKGROUND}" =~ "array" ]]; then
-    local segment=$(( 100.0 / (${#POWERLEVEL9K_BATTERY_LEVEL_BACKGROUND} - 1 ) ))
-    local offset=$(( ($bat_percent / $segment) + 1 ))
-    "$1_prompt_segment" "$0_${current_state}" "$2" "${POWERLEVEL9K_BATTERY_LEVEL_BACKGROUND[$offset]}" "${battery_states[$current_state]}" "${message}" "BATTERY_ICON"
+  if [[ -v "POWERLEVEL9K_BATTERY_HIDE_FULL" && "$POWERLEVEL9K_BATTERY_HIDE_FULL" == true ]]; then
+      if [[ "$bat_percent" != "100" ]]; then
```
dritter: Wouldn't it be easier to inverse that condition and exit early?
Something like this:
```zsh
if [[ "${POWERLEVEL9K_BATTERY_HIDE_FULL}" == "true" && "${bat_percent}" == 100 ]]; then
  return
fi
```
That way we get rid of that indentations and the code duplication..

dritter:
Btw. Something itches my head: I always try to ~~reduce the amount of variables~~ make variables more  meaningful in P9K. So it might be worth it to rename that variable to `POWERLEVEL9K_BATTERY_HIDE_ABOVE_THRESHOLD` and make the value an integer. That way we would gain more flexibility without making the code more complex..

What do you think?

sambadevi:
Thanks! I didn’t think about the code duplication. I will take a look at this later on and update the PR accordingly

powerlevel9k.zsh-theme
```diff
@@ -488,28 +488,19 @@ prompt_battery() {
...
-      # override the default color if we are using a color level array
-      if [[ -n "$POWERLEVEL9K_BATTERY_LEVEL_BACKGROUND" ]] && [[ "${(t)POWERLEVEL9K_BATTERY_LEVEL_BACKGROUND}" =~ "array" ]]; then
-        local segment=$(( 100.0 / (${#POWERLEVEL9K_BATTERY_LEVEL_BACKGROUND} - 1 ) ))
-        local offset=$(( ($bat_percent / $segment) + 1 ))
-        "$1_prompt_segment" "$0_${current_state}" "$2" "${POWERLEVEL9K_BATTERY_LEVEL_BACKGROUND[$offset]}" "${battery_states[$current_state]}" "${message}" "BATTERY_ICON"
-      else
-        # Draw the prompt_segment
-        "$1_prompt_segment" "$0_${current_state}" "$2" "${DEFAULT_COLOR}" "${battery_states[$current_state]}" "${message}" "BATTERY_ICON"
-      fi
+  if [[ -v "POWERLEVEL9K_BATTERY_HIDE_ABOVE_THRESHOLD" && "${bat_percent}" -ge $POWERLEVEL9K_BATTERY_HIDE_ABOVE_THRESHOLD ]]; then
```
dritter: Could you add a comment before that line? Something like:
"Check if POWERLEVEL9K_BATTERY_HIDE_ABOVE_THRESHOLD is set and is exceeded by bat_percent, early exit."

sambadevi: Sure, I’ll add the comment

Issue #828, opened 2018-05-07
RVM prompt broken


HaoZeke:
# Details

Hi. I have a rather odd problem.

When I have the following setup in my `.zshrc`:

```bash
POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(dir newline rbenv virtualenv vcs)
```

I get this
![rbenvbug](https://user-images.githubusercontent.com/4336207/39730310-80565b40-527e-11e8-9598-b5a5d9675ac4.png)

Weirdly though when I have the following:

```bash
POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(dir newline rvm virtualenv vcs)
```

I get nothing after reloading the shell.

![rvmbug](https://user-images.githubusercontent.com/4336207/39730374-dfc5a306-527e-11e8-8d82-454d29af4aab.png)

# System Details

![sysdetails](https://user-images.githubusercontent.com/4336207/39730396-f8bd0d40-527e-11e8-8a7a-fbed55126967.png)

* My `powerlevel9k` setup is via `zplug`:

``` bash
zplug "bhilburn/powerlevel9k", use:powerlevel9k.zsh-theme, defer:3
```

I am using the [current master](https://github.com/bhilburn/powerlevel9k/commit/13b0e383a066c9c37d5cf71d4af521f9b73a173e), determined via:

```bash
$ git rev-parse HEAD
13b0e383a066c9c37d5cf71d4af521f9b73a173e
```

HaoZeke:
Yup, `rvm-prompt` returns the correct ruby version while `echo "${rvm_env_string/ruby-}"` is blank.
![rvmrbenv](https://user-images.githubusercontent.com/4336207/39792054-6a8d4222-535c-11e8-8904-fa6f6b71819b.png)

pmutch:
This also affects me

dritter:
@HaoZeke s fix landed in `master` with #1149 . Thanks for the fix @HaoZeke and everybody commenting here.

Closing for now.

PR #832, opened 2018-05-09
Fix rvm

HaoZeke:
With reference to https://github.com/bhilburn/powerlevel9k/issues/828#issuecomment-387598562


dritter:
Not being a ruby user, I suggest @rjorgenson and @natemccurdy for a proper review.
I always thought `rvm_env_string` is set by RVM itself, but that seems not to be the case.

bhilburn:
Unfortunately, we let this PR bitrot to the point that there is a conflict. Thankfully, it's relatively simple, as it's just from the color codes change.

I really don't know anything about Ruby, and would like to have at least one other knowledgeable reviewer check out the code. @rjorgenson - Any chance you're available to take a peek?

bhilburn:
I really don't want to let this one die! We do need a Ruby user to actually take a look, here, and make sure everything looks good, and we need to resolve conflicts. Going to ping gitter to try and find a Ruby user

HaoZeke:
I am willing to take another look to see if she's is still current but it'll still need a reviewer. 
________________________________

natemccurdy:
I use `rbenv` not `rvm`, so I'm not super familiar with any oddities this may cause, but at a quick glance this seems like a good change.

But I do have some questions :
* For @HaoZeke  - The `rvm-prompt` command by default will output something like ` ruby-1.9.1-p243`. Do you want all that as part of the prompt, or maybe *just* the version number and patch level? For example `rvm-prompt v p` would return `1.9.1-p243`. More example are here https://rvm.io/workflow/prompt
* For @bhilburn or @dritter Should there be some sanity checking that `rvm-prompt` exists and is available?  Maybe just redirect stderr: `$(rvm-prompt 2>/dev/null)`

powerlevel9k.zsh-theme
```diff
@@ -1365,10 +1365,10 @@ prompt_rspec_stats() {
 ################################################################
 # Segment to display Ruby Version Manager information
 prompt_rvm() {
-  local version_and_gemset=${rvm_env_string/ruby-}
-
-  if [[ -n "$version_and_gemset" ]]; then
-    "$1_prompt_segment" "$0" "$2" "grey35" "$DEFAULT_COLOR" "$version_and_gemset" 'RUBY_ICON'
+  local rvm_prompt
```
natemccurdy: @HaoZeke You should probably wrap this entire thing in `if [ $commands[rvm-prompt] ]` to prevent issues if the command doesn't exist.

Like this:
```shell
prompt_rvm() {
  if [ $commands[rvm-prompt] ]; then
    local rvm_prompt
    rvm_prompt=$(rvm-prompt v p)
    if [ "$rvm_prompt" != "" ]; then
      "$1_prompt_segment" "$0" "$2" "240" "$DEFAULT_COLOR" "$rvm_prompt $(print_icon 'RUBY_ICON') "
    fi
  fi
}
```

natemccurdy:
Looks good to me 👍

HaoZeke:
Sure, but I will have to get to it in a few days, I have exams right now...

dritter:
Closing this one in favor of #1149 .

PR #838, opened 2018-05-17
add cog symbol to dir segment if folder is /etc*

sambadevi:
This PR adds a new setting for the `dir` segment to style `/etc` and subfolders independently

It also adds a default Icon to use `\uF013` (only if you are using font-awesome or nerd-fonts config)

Options to change colors and icon:
```
POWERLEVEL9K_ETC_ICON='\uf0ad'
POWERLEVEL9K_DIR_ETC_BACKGROUND='57'
POWERLEVEL9K_DIR_ETC_FOREGROUND='white'
```

Example Screenshot:
bildschirmfoto 2018-05-17 um 12 54 51

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -909,10 +909,13 @@ prompt_dir() {
     "HOME"            "HOME_ICON"
     "HOME_SUBFOLDER"  "HOME_SUB_ICON"
     "NOT_WRITABLE"    "LOCK_ICON"
+    "ETC"             "ETC_ICON"
   )
   local state_path="$(print -P '%~')"
   local current_state="DEFAULT"
-  if [[ "${POWERLEVEL9K_DIR_SHOW_WRITABLE}" == true && ! -w "$PWD" ]]; then
+  if [[ $state_path == '/etc'* ]]; then
+    current_state='ETC'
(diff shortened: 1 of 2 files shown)


powerlevel9k.zsh-theme
```diff
@@ -909,10 +909,13 @@ prompt_dir() {
     "HOME"            "HOME_ICON"
     "HOME_SUBFOLDER"  "HOME_SUB_ICON"
     "NOT_WRITABLE"    "LOCK_ICON"
+    "ETC"             "ETC_ICON"
   )
   local state_path="$(print -P '%~')"
   local current_state="DEFAULT"
-  if [[ "${POWERLEVEL9K_DIR_SHOW_WRITABLE}" == true && ! -w "$PWD" ]]; then
+  if [[ $state_path == '/etc'* ]]; then
```
iScrE4m: I wonder if an option to turn this off should be present?

dritter: Well. You could override `ETC_ICON`..

iScrE4m: Sure, that's an option. I was going off of `DIR_SHOW_WRITABLE` but looking at the rest, there's not really a set precedent as I thought, as home behaves exactly like this, Sorry!

Issue #794, opened 2018-04-02
cropped icons


D1mon:
get_icon_names:
![icon](https://user-images.githubusercontent.com/3872485/38208099-9c614110-36b8-11e8-8ad0-298a3e65a704.png)

in terminal:
![cropped_icons](https://user-images.githubusercontent.com/3872485/38208123-b1eb66be-36b8-11e8-82bb-915b85c6b535.png)

two icons cropped (home directory and subdirectories), others seem to work fine.

OS: Manjaro
font: Hack (ttf-hack package)
terminal: Konsole (kde)
POWERLEVEL9K_MODE='awesome-patched'

dritter:
Hi @D1mon !

Yes, it seems like [Konsole does not handle double-width characters correctly](https://bugs.kde.org/show_bug.cgi?id=378124). We had reports about [this before](https://github.com/bhilburn/powerlevel9k/issues/453). The best workaround is to [add a whitespace to the icon](https://github.com/bhilburn/powerlevel9k/issues/231#issuecomment-194240147).
I hope that helps.

D1mon:
Yes, it helped. Thank you.

Issue #453, opened 2017-03-22
Awesome Powerline icons cut off


binxycraft:
This is probably not a powerlevel9k issue but I figured this would be the best place to start since someone may have an answer. I have recently installed powerlevel9k on 3 different computers. I am using Kayant's config as a starting point and like it so far. I am using SourceCodePro+Powerline+Awesome+Regular since it seemed popular. The problem is with the Awesome powerline icons. This happens on all 3 computers exactly the same. I am running Manjaro/KDE on all three and using Konsole as my primary terminal emulator. I have tried playing with every appearance setting I could find within Konsole but could not get this behavior to stop. I decided to try another terminal emulator (Terminator) and the font seemed fine there. I thought maybe it only affected Qt emulators so I downloaded QTerminal and I had the exact same issue as with Konsole. Weirdly enough the folder icon (outside of home folder) is fine but the other two are cut off. I know as a workaround I could disable these icons but I kind of like them so would rather get it fixed if anyone has any ideas.

![powerlevel9k](https://cloud.githubusercontent.com/assets/8657284/24179146/b80bea5e-0e83-11e7-8aa8-7341bb7df76e.jpeg)

D4v2z:
Hey man, the solutions : https://github.com/bhilburn/powerlevel9k/issues/231

dritter:
Ha! Thanks @D4v2z for finding that solution. I completely forgot about it.

Btw. I did a quick research this morning and found out that `Konsole` has some issues regarding double-width glyphs. Hard to say what other terminal emulators do different. The information that a glyph needs to be printed as double-width is coded somewhere in the font I guess..

binxycraft:
Perfect! Thank you. This is exactly the kind of workaround I was looking for. I must have been searching the wrong key words. Following the example I also added another entry for my other cut-off icon:
`POWERLEVEL9K_HOME_SUB_ICON=$'\UE18D '`
Using this page as reference:
https://github.com/bhilburn/powerlevel9k/blob/master/functions/icons.zsh

dritter:
Btw. , Just for completeness, easier than looking into `functions/icons. zsh` would have been to call `get_icon_names`. ;)

millsoft:
Hi, I did this but no home icon is being displayed, just a square. :(

![screenshot](https://user-images.githubusercontent.com/964209/32226936-2d398248-be4a-11e7-97a8-132581d98218.png)

dritter:
[...]

Issue #231, opened 2016-03-09
Context icon not displayed properly


vinaychandra:
As you can see, the font is installed correctly but the prompt is breaking in between. What is the solution for this probelm
![snapshot2](https://cloud.githubusercontent.com/assets/5670927/13630006/e23cf502-e602-11e5-8cc0-a2485ecbab07.png)

dritter:
What do you mean by saying the prompt breaks in between? Do you want a multiline-prompt? If so, you need to set `POWERLEVEL9K_PROMPT_ON_NEWLINE=true`.

vinaychandra:
Sorry for my english. I meant that the icon that is being displayed is broken. In the first line, the home icon is partially displayed. I want to set it right

dritter:
No need to apologize. I was to stupid to see the broken icon in the first place. :smile: 

The problem is quite complex, and has to do with how the terminal editor renders multi-width glyphs. So it is more an issue somewhere there (nothing we can really do about).

A quick solution _could_ be to render more whitespace after the icon.. Set `POWERLEVEL9K_HOME_ICON=$'\UE12C '` (note the whitespace after the codepoint). This codepoint is for `awesome-patched`-mode. If you use `awesome-fontconfig`, it should be `\UF015`.

It might also be a setting in your terminal emulator. Maybe there is somewhere a switch to treat all fonts as monospaced fonts..

Btw. which terminal emulator do you use? Is that terminix?

vinaychandra:
It is Konsole

dritter:
Okay. Good to know. But did giving the icon more whitespace work?

vinaychandra:
Yes. Giving the icon more whitespace worked. Thank you :smiley:

9ao9ai9ar:
@vinaychandra I patched the problem using this ugly code:
```
for key in ${(k)icons[@]}
do
    if [[ ! $key =~ 'SEPARATOR' ]]
    then
        icons[$key]=" ${icons[$key]} "
    fi
done
```
Add it after `antigen apply` in your `.zshrc`.

PR #854, opened 2018-06-01
Fix icons cut off in some terminal emulators

ytang:
For terminal emulators such as [suckless](https://st.suckless.org/), the most common scenario of icons being cut off is the virtual identifier before the segment content, even though there is a space between the icon and the text. The reason is that we add color only to the icon but not to the space next to it. To fix it, just add color to both the icon and its neighboring space. It allows the terminal emulator to render the icon correctly without introducing any visual side effects.

Here is a screenshot showing the cut-off before (the first line) and after (the second line) applying the patch:
powerlevel9k-fix-icons-cut-off

This is superior to existing workarounds mentioned in #231 #453 #794 since we do not need to waste an extra space.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -156,7 +156,9 @@ left_prompt_segment() {
   if [[ -n $6 ]]; then
     visual_identifier="$(print_icon $6)"
     if [[ -n "$visual_identifier" ]]; then
-      # Add an whitespace if we print more than just the visual identifier
+      # Add an whitespace if we print more than just the visual identifier.
+      # To avoid cutting off the visual identifier in some terminal emulators (e.g., Konsole, st),
+      # we need to color both the visual identifier and the whitespace.
       [[ -n "$5" ]] && visual_identifier="$visual_identifier "
       # Allow users to overwrite the color for the visual identifier only.
       local visual_identifier_color_variable=POWERLEVEL9K_${(U)1#prompt_}_VISUAL_IDENTIFIER_COLOR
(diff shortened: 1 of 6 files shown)


powerlevel9k.zsh-theme
```diff
@@ -156,12 +156,12 @@ left_prompt_segment() {
   if [[ -n $6 ]]; then
     visual_identifier="$(print_icon $6)"
     if [[ -n "$visual_identifier" ]]; then
+      # Add an whitespace if we print more than just the visual identifier
+      [[ -n "$5" ]] && visual_identifier="$visual_identifier "
```
dritter: Could you add a comment why there is a whitespace before ending the foreground color?

powerlevel9k.zsh-theme
```diff
@@ -240,12 +240,12 @@ right_prompt_segment() {
   if [[ -n "$6" ]]; then
     visual_identifier="$(print_icon $6)"
     if [[ -n "$visual_identifier" ]]; then
+      # Add an whitespace if we print more than just the visual identifier
+      [[ -n "$5" ]] && visual_identifier=" $visual_identifier"
```
dritter: Could you here add a comment about the whitespace as well?

powerlevel9k.zsh-theme
ytang: Done.

powerlevel9k.zsh-theme
ytang: Done.

ytang:
Thanks for the swift approval! It works for Konsole, too.

dritter:
Just tested this in `iTerm2`, `Terminal` and `Konsole`. Works like a charm in all of them.
Thx @ytang

bhilburn:
Thanks so much, @ytang! Also, kudos for searching through our previous bugs to see where else this issue has been mentioned, and what was discussed there. Thanks for the solid PR! `=)`

bhilburn:
With @dritter's test & approval, this PR will go into the next release.

IevgenSobko:
@dritter @ytang  has this fix found a way to master?
It seems doesn't work for right-hand segments. After update left-hand segments got fixed under Konsole but not right-hand ones

kirVershinin:
@ytang this fix was perfect for me for a month, but to day it show only cut off icons.
I already tried to add one more space for your decision but it didn't help. Any ideas what this could be?

braydie:
This was also working for me but appears to have stopped working for me as well...

braydie:
Yeah it was working for me to, but isn't now

PR #856, opened 2018-06-05
[WIP] Make full use of newline segment

dritter:
This PR is a proposal to get rid of the `ADD_NEWLINE` Variables. At least to substitute this behaviour with the more flexible `newline` segments.
This is still WIP, as there is #771 out, but I want to discuss the idea anyway.

This is still not finished. There are some minor bugs on the left prompt (e.g. rendering the segment separator in some cases) and some major on the right prompt (segments not indented correctly).

So, here are my questions:
### General
- Do we want such a functionality? Is it really superior to plain variables?
- Do we want unique prefixes for each newline? -> Like having a automatic variable for every newline?

### Right prompt
- As we cannot use real newlines to break the line, we need to move the cursor to make it look like a newline. The downside of this approach is that we have to indent the segments manually. This involves some calculation and some vodoo, as the segments are printed from left to right, and we have to check if one of the following segments is a newline (we would have to indent the previous segments by the width of the following ones). This is especially harder when taking conditional segments into account.
- Cursor movement is always a source of weird bugs (thinking of moved cursors and autocompletion).

But even without the right prompt, I think this might be worth it.

//cc @docwhat I'm pretty sure you have a opinion about this. ;)

### DIFF
--- a/functions/utilities.zsh
+++ b/functions/utilities.zsh
@@ -209,6 +209,9 @@ segment_in_use() {
     fi
 }
 
+# Count occurrences of items in a flat array.
+#   * $1: The item to search for
+#   * $2: The array to search in
 count_in_array() {
   local needle="${1}"
   # Explicitly split on whitespaces
--- a/functions/utilities.zsh
+++ b/functions/utilities.zsh
@@ -209,6 +209,21 @@ segment_in_use() {
     fi
 }
 
+count_in_array() {
+  local needle="${1}"
+  # Explicitly split on whitespaces
+  local haystack=(${=2})
+  local count=0
+
+  for key in ${haystack}; do
+    if [[ "${key}" == "${needle}" ]]; then
+      count=$((count + 1))
+    fi
+  done
+
+  echo "${count}"
+}
+
 # Print a deprecation warning if an old segment is in use.
 # Takes the name of an associative array that contains the
 # deprecated segments as keys, the values contain the new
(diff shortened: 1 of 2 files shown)


docwhat:
> I'm pretty sure you have a opinion about this. ;)

Always.  If you don't like my opinion, I have others.

docwhat:
You don't need multi-line `RPROMPT` support; Use a multi-line `PROMPT` instead along with some math to push things to the right.

```sh
# example
POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=( redfish newline bluefish newline greenfish )
POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=( dog newline cat )

# becomes (very fake code here...)
PROMPT="${redfish}\n${bluefish}${padding}${dog}\n${greenfish}"
RPROMPT="${cat}"
```

Alternatively, never let anyone add a newline in `RIGHT_PROMPT_ELEMENTS` and just add a `padding` segment:

```
# Example
POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=( redfish newline bluefish padding dog newline greenfish )
POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=( cat )
```

dritter:
Hah! Great Idea! Thx @docwhat . That will be interesting to implement with ZSH I guess. :D

docwhat:
Yeah, `padding` should be able to handle arbitrary characters to pad with (e.g. space or a line or whatever).

PR #859, opened 2018-06-07
[WIP] Staging 0.6.5

dritter:
This is a staging branch to be merged in `next`. 

It includes the following changes (in order of merge):
- `master`. For easier merging later on, `master` was merged in first. There were two conflicts, one being a conflict in the `README.md` (#781 vs #684) and another one in the `powerlevel9k.zsh-theme` (#795 vs #610). The latter one being the more impacting one, as I had to drop one completely (both implemented the same thing). I went with #795, even this one is more recent. The reason for that is that this commit is the one on `master`, and I don't want to deprecate a variable just for naming reasons (or the naming must improve vastly, which didn't here).
- #797. Conflicts: `functions/vcs.zsh`: Two commits avoided wrapper behaviour. #685
    in grep, #797 in git.
- #798
- #783 Conflicts: `functions/icons.zsh` and `powerlevel9k.zsh-theme`: On `master` the Java Segment was added, on `next` the Dropbox Segment.
- #641 Conflicts: `test/segments/rust_version.spec` and `powerlevel9k.zsh-theme`: On `master` the "Rust" was refactored as Visual Identifier; On `next` the color names changed.
- #826 Conflicts: `test/segments/rust_version.spec` and `powerlevel9k.zsh-theme`: On `master` the Rust version is now parsed differently; On `next` the color names changed.
- #816 
- #830 
- #838 
- #847 
- #844 
- #854 Conflicts: `test/segments/command_execution_time.spec` and `test/segments/go_version.spec`: All conflicts come from renaming color names on `next` (#703) and adding a whitespace to the Visual Identifier (#854) on `master`.
- #818 Conflicts: `functions/icons.zsh`: Java Segment (#783) was added on `master` with icons.
- #871 

Not yet merged (This is on purpose and will be done in #863 ):
- #771

PR #833, opened 2018-05-10
Rewrite - Test branch in progress

onaforeignshore:
After the various discussions about streamlining and speeding up powerlevel9k - especially with precompiling as much as possible and removing tests wherever possible, I have updated the entire code base (based on #771).

This is a WIP as it does not now have a rewritten zsh-async.p9k yet, but it seems to be a whole lot faster.

@bhilburn @dritter Please give it a try and let me know what you think.

PR #863, opened 2018-06-09
[WIP] Staging 0.7.0

dritter:
Here is the big merge of #771 and #859 . Please have a look if everything is all right.
I did a quick test (just open the shell, and see that everything is all right) and got an exception in the `status` segment. I fixed that with 8a6597def7ba2bda54fe6d695ded7c2c477a9b0e . That is the only additional change I've made on this branch (besides a bit of cleanup in d11aba28c0f3233ae388d25cb2e2534deddc4d3f and fixing the path to the `test_stats` segment).

Biggest pain points in this merge was that code that was moved changed in `next` and git not being able to track that, so I had to do that manually. I went through all segments and all icons to track the code by hand.

I hope I got everything right and did not forgot anything.

Btw. besides a vast change in the code base, my biggest concerns about this are:
- `vi_mode` segment. The changes in #771 originate from #344 and were built for an async environment. I am not sure how this will behave once we have different generators. Currently, @onaforeignshore disabled it for the default generator.
- ~~`vcs` segment. @onaforeignshore enabled queues in mercurial. This needs to be tested as well.~~ Mercurial Queues were removed. See [chat](https://gitter.im/bhilburn/powerlevel9k?at=5b1bf27a659aad4a3262af02). Commit: 3064838b47c4aeda468562b079a3102b4f22aac4
- Loading segments. In `powerlevel9k.zsh-theme` we go through all possible segments and load only the segments in use. This has two problems: `segmentsInUse` does not take joined segments into account. And secondly we could loop over the configured segments and try to load them (This is probably the shorter list, so this is an performance issue).

~~I'll have a look at the tests tomorrow.~~ Tests now pass.

Quick links:
[Compare 0.6.5 and 0.7.0](https://github.com/dritter/powerlevel9k/compare/staging_065...dritter:staging_700)
[Compare #771 and 0.7.0](https://github.com/dritter/powerlevel9k/compare/onaforeignshore:code_separation...dritter:staging_700)


onaforeignshore:
I'm reviewing this PR over the next few days. Initial missing item I have noticed is custom segments. We still have to decide where this will be. The latest trend is to place them under `~/.config`.

onaforeignshore:
This branch still uses only `registerIcon`, I expanded that into `registerSegment` in #833, which IMHO  is a better structure as it registers everything to do with a segment in one call.

Since #833 is only a test bed, it shouldn't be pulled straight into this if we want to use it, but I am willing to do the recoding to incorporate that into here. Should only take a few days to do.

Do we want to just go all out, or do we want to do it in stages, bearing in mind that this PR will differ from #833 's `registerSegment`, meaning that people will have to update their custom segments again.

Do we want to go for the speed that predetermining all the segment formatting gives us or stick to this in-between version?

@bhilburn @dritter What are your thoughts on this?

dritter:
Haven't had time to look into #833 [Diff-Quicklink](https://github.com/dritter/powerlevel9k/compare/staging_700...onaforeignshore:rewrite).

About the question how to proceed, and what should be in the next Release:
In my opinion we shouldn't release a intermediate version. I'd like to keep the count of breaking changes as little as possible. Therefore, if we already know that some changes will break backwards compatibility, we should merge them in. The changes of this branch are a massive change anyway, so it is a good idea to include the performance optimizations (like `registerSegment`).

But, to avoid having such massive changes all the way, it would be good to create small PRs with one feature at a time. Otherwise it will be reviewing / testing Hell.

Tl;dr:
I would suggest to chop #833 in small separate PRs with this one as base. And to incorporate the performance improvements before releasing a new version.

README.md
```diff
@@ -210,7 +210,7 @@ without affecting the icon color.
 
 You can also change the battery icon automatically depending on the battery
 level. This will override the default battery icon. In order to do this, you
-need to define the `POWERLEVEL9k_BATTERY_STAGES` variable.
+need to define the `P9k_BATTERY_STAGES` variable.
```
onaforeignshore: Shouldn't the "k" be "K"?

dritter: Good catch!

dritter:
Closing in favor of #865 and following PRs.

Issue #864, opened 2018-06-09
Make RPROMPT_PREFIX and RPROMPT_SUFFIX configurable


roganartu:
I've been playing around with a multi-line prompt idea, but I don't like the two options available for where the right prompt ends up. It's either on the bottommost line, or the line above it.

I want to be able to have it on the topmost line.

Current:
<img width="1437" alt="screen shot 2018-06-09 at 5 00 50 pm" src="https://user-images.githubusercontent.com/1133705/41196077-ba94e06e-6c06-11e8-8d50-037b740dca8b.png">

Desired:
<img width="1440" alt="screen shot 2018-06-09 at 5 01 38 pm" src="https://user-images.githubusercontent.com/1133705/41196083-cfbae614-6c06-11e8-9510-c80d416f1e52.png">

Achieved with config:
```zsh
POWERLEVEL9K_RPROMPT_ON_NEWLINE=false
POWERLEVEL9K_RPROMPT_PREFIX='%{'$'\e[3A''%}' # three lines up
POWERLEVEL9K_RPROMPT_SUFFIX='%{'$'\e[3B''%}' # three lines down

POWERLEVEL9K_MULTILINE_NEWLINE_PROMPT_PREFIX=$'\u2502'
PROMPT_BLANK="${POWERLEVEL9K_MULTILINE_NEWLINE_PROMPT_PREFIX}\n"
POWERLEVEL9K_MULTILINE_LAST_PROMPT_PREFIX="${PROMPT_BLANK}${PROMPT_BLANK}${PROMPT_BLANK}\u2570\uf460%f "
```

I'll submit a PR that implements this shortly, but my zsh skills are limited to trial and error + googling so I can make no promises of quality. Seems simple enough though.

softmoth:
I think this should probably wait for #856 to be resolved.

bhilburn:
Oops, I had missed #865. Let's take the conversation there, instead.

PR #865, opened 2018-06-09
make RPROMPT_PREFIX and RPROMPT_SUFFIX configurable

roganartu:
Implements #864


bhilburn:
This seems like a pretty straight-forward change, and I see @dritter has already added it to the v0.7.0 list.

@dritter @onaforeignshore - This one looks okay, to me. I'm going to re-target it to the `next` branch, which means it will fail tests because the code has moved, but the change itself otherwise seems sane. Thoughts?

dritter:
Thx @roganartu and sorry for the delay.

PR #871, opened 2018-06-15
Hide one-space indent at right margin

softmoth:
ZSH displays the right prompt indented one space from the right
margin. This can be tweaked with `ZLE_RPROMPT_INDENT=0`, but that
makes display problems very likely and I don't recommend it. But
all is not lost.

The %E prompt format does continue the current background color
through that last indent space. So it can *look* like our prompt
goes right up to the margin, if:
1) We remove the trailing space from the last segment, and
2) We add %E before clearing the background color

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -222,6 +222,12 @@ right_prompt_segment() {
   [[ -n "$4" ]] && fg="%F{$4}" || fg="%f"
 
   # If CURRENT_RIGHT_BG is "NONE", we are the first right segment.
+
+  if [[ "$CURRENT_RIGHT_BG" != "NONE" ]]; then
+    # This is the closing whitespace for the previous segment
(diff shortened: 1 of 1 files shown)


dritter:
Awesome! I didn't know that `%E` exists. Thanks @softmoth . With this we could remove the right whitespace in the last segment as well.

But the current implementation won't work if the last Segment hast an unmet condition.
There are three ways to fix this:
- always set `%E` after a segment. Not optimal, but would work.
- store the last Segment in a global variable.
- search for the last printed segment

softmoth:
Please refresh the patch, I updated it thus:

1) print the between-segment whitespace at the beginning of right_prompt_segment, if it's not the first segment
2) print %E in build_right_prompt after all segments are done

This is simple and should behave consistently w/o extra state variables.

softmoth:
Screenshot

bhilburn:
@softmoth - This is a really cool PR, and as @dritter said, we didn't even know this was possible. Thanks so much for submitting this! We are discussing how and where to pull this in, now.

PR #872, opened 2018-06-15
Support the SEARCH vi mode from zsh-vim-mode

softmoth:


softmoth:
I updated this branch to base it off of #873.

softmoth:
Not similar to that, although two plugins may have a small bit of overlap on a couple key bindings. But this change is just to represent when in incremental search mode (like when typing ^R in default emacs keymap). So vi_mode users (who have the plugin installed) would see "INSERT" change to "SEARCH" when they hit ^R, and then back to "INSERT" when they hit <Esc> or  a movement key, for example.

dritter:
Thx @softmoth ! Merged into `next`, but needs `zsh-async` generator.

PR #873, opened 2018-06-15
Hide vi_mode segment if string is empty

softmoth:


dritter:
Sorry for the delay @softmoth . I merged this now into `next`. The changes will need to have the async generator available (which is still WIP).

PR #875, opened 2018-06-17
[WIP] Test Bed - Testing possible future improvements

onaforeignshore:
Here are the updates to #863 to incorporate the new `registerSegment`. This supercedes #833 as all the ideas have been incorporated here.

It contains all the code needed to make use of the new format, as well as a new function to rename variables to match the new structure - `printDeprecationVarWarning`

TODO: Decide where custom segments will be located and load them at initialization.


onaforeignshore:
After working all day on this, the next step is to update the tests

bhilburn:
Just as a note: right now, this branch / PR looks really complicated, but that's because it's built on top of #863, which is not yet in `next` - so it's duplicating that commit tree in the PR. Once we have merged #863, this branch / PR should be rebased onto `next`, which should make it much easier to manage.

onaforeignshore:
zsh-async still has two major issues with segments not working that need to be fixed before it can be released:

- command_execution_time
- status

onaforeignshore:
Once the code separation is completed, I will create multiple PRs to implement all the changes here.

PR #878, opened 2018-06-17
Show if GIT_DIR is in use, warn if clobbering $PWD

softmoth:
If GIT_DIR is set, it overrides any repository that the current
directory may be part of. This is confusing when not expected,
and the prompt can be very helpful in recognizing this condition.

For one example of why setting GIT_DIR can be useful in the first
place, see
https://developer.atlassian.com/blog/2016/02/best-way-to-store-dotfiles-git-bare-repo/


softmoth:
shot-20180617 1

dritter:
Thx @softmoth

bhilburn:
Thanks so much @softmoth! =)

Issue #877, opened 2018-06-17
[Discussion] Minimum System Requirements


dritter:
We never specified any minimum system requirements, but there are implicit requirements. Besides ZSH being the biggest part of this, I didn't want to constrain it to ZSH only.

- `functions/colors.zsh` need at least [ZSH 5.1](https://github.com/bhilburn/powerlevel9k/pull/859#issuecomment-397842819)
- Conditionals like `[[ -v "VARIABLE" ]]` (found [here](https://github.com/bhilburn/powerlevel9k/pull/830/files#diff-ea5a405f0331722fc10a832209ffe20bR491)) to test if a variable is declared need at least ZSH 5.3. Workaround is to use our own `defined` from [`functions/utilities.zsh`](https://github.com/bhilburn/powerlevel9k/blob/7234d52ebd76b6b89e172ac5b19339f0a37ab596/functions/utilities.zsh#L11-L15). I we decide to go with ZSH 5.3, we could deprecate our `defined` function.

I am sure there are others. I just wanted to collect my findings so we could set up a wiki page. Also, it is important to know what ZSH versions we should test.

onaforeignshore:
@bhilburn Any thoughts on this? Would be nice to know what the minimum is as it affects not only coding, but also testing

bhilburn:
Yup, I think this is a great idea. We should for sure do this for v0.7.0.

@onaforeignshore - You have a great point re: testing. Let's settle on our minimum reqs, here, and then make sure we are testing against that (as well as the latest).

So, generally, I like using Ubuntu 16.04 LTS as a good benchmark for minimum requirements, which ships ZSH 5.1.1 by default ([see here](https://packages.ubuntu.com/xenial/zsh)) - so I think we have to keep our own `defined`, for now. Once 18.04 LTS matures a bit and starts to displace 16.04, we can chat about bumping it up.

What else should we cover? Also, it would be good to test for minimum requirements at init and issue an error if they aren't present.

dritter:
Too late to keep 5.1. At least for the battery segment.. To keep that, we should rework that condition to use our `defined` function.

Other minimum requirements we should document are `git`, `mercurial`, `svn`.. In other words: we should look at every segment and document the minimum version of the used programs, especially segments that make heavy use of other programs like `VCS`- , `battery`- and other segments.

robobenklein:
[...]

PR #884, opened 2018-06-27
test-in-docker: fix with newer ZSH versions

docwhat:
The way I was filtering out entries in the frameworks array stopped working in newer versions of ZSH; it would convert the array into a string (you could see it with `typeset -p frameworks`)

So I rewrote it.

I don't see anything in the release notes for ZSH that would explain this and I didn't find any option that would restore this behavior.

I also added a bunch of debugging and warning stuff to catch problems in the future.

And there is a `--dry-run` flag now.

Related: #882 (cc: @dritter)

### DIFF
--- a/docker/base-5.2/Dockerfile
+++ b/docker/base-5.2/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:17.04
+FROM ubuntu:17.10
 
 RUN \
   apt-get update && \
--- a/docker/base-5.2/Dockerfile
+++ b/docker/base-5.2/Dockerfile
@@ -6,7 +6,7 @@ RUN \
   DEBIAN_FRONTEND=noninteractive apt-get install -y \
   curl \
   git \
-  zsh \
+  zsh=5.2-5ubuntu1.2 \
   mercurial \
   subversion \
   golang \
--- a/test-in-docker
+++ b/test-in-docker
@@ -6,6 +6,7 @@ set -eu
 default_version='4.3.11'
 
 setopt extended_glob glob_subst numeric_glob_sort
+setopt warn_create_global warn_nested_var 2> /dev/null
 cd "${${(%):-%x}:A:h}"
 
 # TODO: Crazy Logic to munge TERM to something supported in Ubuntu 14.04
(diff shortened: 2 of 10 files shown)


docwhat:
@dritter I wouldn't mind a review on this, if you have time.

test-in-docker
```diff
@@ -15,13 +16,18 @@ term=screen-256color
...
+typeset -r versions
 
 # List of frameworks
-typeset -a frameworks
+typeset -aU frameworks
 frameworks=( docker/*/Dockerfile(N.on:h:t) )
-frameworks=${(@)frameworks:#base-*}
+for i in {$#frameworks..1}; do
+  # Remove all base entries
+  [[ "${frameworks[$i]}" = base-* ]] && frameworks[$i]=()
```
dritter: Just a stylistic note: I would go here with double equals. That is more readable IMHO.

test-in-docker
```diff
@@ -6,6 +6,7 @@ set -eu
 default_version='4.3.11'
```
dritter: Maybe we should set the minimum version to something like 5.3 in regard of #877

docwhat: I started to do that in a94df3ddcb027d88717f854cd5f610a771c1c71c but then remembered why I have it the way I do.

The default should be the _oldest_ version of zsh we support, not newest.

Either way, it is easier to change now and it has the same lookup properties the command line does.

test-in-docker
docwhat: Done in 3c27414e7e985536a8b286d09b2248e7dee8263e

dritter:
Btw. Did you see my questions in the chat?

test-in-docker
dritter: Well. The oldest supported version of ZSH _is_ 5.3 in `next` (at least in the battery segment). Otherwise it is [5.1](https://github.com/bhilburn/powerlevel9k/pull/859#issuecomment-397842819).

docker/base-5.2/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:17.04
+FROM ubuntu:17.10
```
docwhat: Correct.  17.04 (and 17.10) are not LTS versions; they will go away.

bhilburn:

PR #887, opened 2018-06-29
Windows battery support

ratijas:
Initial support for battery widget on Windows OS.

It uses [WMIC](https://docs.microsoft.com/ru-ru/windows/desktop/WmiSdk/wmi-start-page) output which looks pretty reliable to parse with stuff like grep and sed.

I admit there's still some work to do, like supporting systems without batteries and respecting `$POWERLEVEL9K_BATTERY_LOW_THRESHOLD` variable.

But it's a good starting point, and I would be glad to hear feedback from MINGW / MSYS2 users.

### DIFF
--- a/functions/utilities.zsh
+++ b/functions/utilities.zsh
@@ -86,7 +86,7 @@ case $(uname) in
       OS='OSX'
       OS_ICON=$(print_icon 'APPLE_ICON')
       ;;
-    CYGWIN_NT-*)
+    CYGWIN_NT-* | MINGW32_NT-* | MINGW64_NT*)
       OS='Windows'
       OS_ICON=$(print_icon 'WINDOWS_ICON')
       ;;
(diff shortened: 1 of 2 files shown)


dritter:
Hi @ratijas !

Thank you very much for your PR, and sorry for the delay. We've done quite some work on the `next` branch in the meantime. Could you update your PR?
And could you provide the output of `wmic Path Win32_Battery Get /format:list` and add a test for the battery segment on windows. That would be awesome.

dritter:
Btw. I just merged another PR that touches the battery segment. Could you update your PR?

For the records, [here](https://forums.lenovo.com/t5/ThinkPad-T400-T500-and-newer-T/Lenovo-T470-Core-i5-7200u-Slow-performance-on-battery/m-p/3804422#M118897) is a sample output of `wmic Path Win32_Battery Get /format:list`.

dritter:
Closing in favour of #1120

PR #888, opened 2018-06-30
Fix os detection for MSYS

blackb1rd:


ratijas:
Wow, cool. How did you achieve that?

I have two MSYS2 shells: 32 and 64 bit versions, both report something like `MINGW(32|64)_NT-*`, but i never saw `MSYS_NT-*`.

Maybe it would be reasonable to add it into this commit? https://github.com/bhilburn/powerlevel9k/pull/887/commits/0fc517ce788ee35bba57e53494859e89777c3fe1

blackb1rd:
@ratijas how did you run MSYS? there are 3 shell types(you could see more detail in `msys2_shell.cmd` file) and when call `uname` command, the result will show as below:
mingw32 -> MINGW32_NT-*
mingw64 -> MINGW64_NT-*
msys2 -> MSYS_NT-*

Normally, MSYS2 is using msys2 shell type then result of `uname` will be `MSYS_NT-*`.

bhilburn:
Thanks so much for putting together this fix and contributing it, @blackb1rd! Much appreciated, and welcome to P90x :)

Issue #885, opened 2018-06-27
Is there a way to truncate directory name from left?


munjalpatel:
I see there is a `truncate_from_right` strategy. However, I want to truncate names from left instead of right. Is there a similar `truncate_from_left` option?

dritter:
Hi @munjalpatel !

@onaforeignshore was so kind to add this missing truncation option. Could you try that PR (#889) and see if that works for you? That would be awesome.

bhilburn:
Thanks so much for opening this, @munjalpatel! Closing this in favor of tracking the discussion in #998, where the work is happening `=)`

PR #889, opened 2018-07-01
Dir - New Strategy - truncate_from_left

onaforeignshore:
New strategy for the dir segment - truncate_from_left

This adds the ability to shorten a directory from the left. It leaves just the end of a folder name untouched. E.g. your folders will be truncated like so: "/..9k/..st/segments". How many characters will be untouched is controlled by P9K_SHORTEN_DIR_LENGTH.

NOTE: Created in response to #885


onaforeignshore:
Replaced by #998

PR #891, opened 2018-07-02
kubecontext: fix context, namespace comparison

krmcbride:
Maybe the original author happened to name their context `k8s_namespace` 🤔

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1429,7 +1429,7 @@ prompt_kubecontext() {
 
     local k8s_final_text=""
 
-    if [[ "$k8s_context" == "k8s_namespace" ]]; then
+    if [[ "$k8s_context" == "$k8s_namespace" ]]; then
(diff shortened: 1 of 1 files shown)


dritter:
Good catch @krmcbride !

Summoning @iScrE4m , @zikes and @captjt for better Review :)

iScrE4m:
Thanks for doing this, I was actually embarassed to fix it, hoping noone else was naming them the same. Now my shame every time I look at the prompt can go away.

zikes:
LGTM :+1: 

Slightly related, but I recently discovered https://github.com/jonmosco/kube-ps1 and it might be worthwhile integrating that somehow rather than "rolling our own" :thinking:

bhilburn:
Thanks so much for the fix, @krmcbride, and thanks for the reviews @iScrE4m @zikes @captjt! Great work all around. Welcome to P9k @krmcbride :)

Interesting idea @zikes! I'll check it out.

PR #893, opened 2018-07-06
Improve upsearch

dritter:
This PR improves upsearch to be more performant and more maintainable.

In terms of performance, it changes multiple recursive calls to (well, builtins, but anyway) `pushd` and `popd` to just one external call to `find`. This is possible, because `find` can take multiple paths to search in. E.g. `find /Users/dr/test/1/12/123 /Users/dr/test/1/12 -maxdepth 1 -name .bla`

To build up the paths to search in, we loop through all directories, until the root is reached. This is a simple string operation, as we just cut off the last slash.
bildschirmfoto 2018-06-26 um 23 43 34
(And so on, until the root was reached).

After that we just pass these paths to `find`.

Previously, that was a recursive call:
bildschirmfoto 2018-06-26 um 23 43 07
As you can see there is much more going on.

The `upsearch` function now returns the results in opposite order. Previously it returned the shortest match first, now it's the longest.
bildschirmfoto 2018-07-06 um 18 09 39

Another difference is that with `truncate_with_folder_marker` strategy, the first character is not slash any more.. This could be added easily again, but I wanted to keep the truncation strategies more consistent.

Unfortunately, I wasn't able to `time` the old upsearch, so there is no real comparison..


dritter:
@onaforeignshore did some research and found a recursive glob as described [here](https://unix.stackexchange.com/a/64164) which is even more efficient. Btw. using find never was, as Christo found out:
```zsh
# Using current upsearch...
( repeat 10000; do; upsearch_old $1 > /dev/null; done; )  9.62s user 13.44s system 99% cpu 23.092 total

# Using upsearch with find...
( repeat 10000; do; upsearch_new $1 > /dev/null; done; )  18.05s user 22.16s system 90% cpu 44.275 total

# Using upsearch with zsh globbing...
( repeat 10000; do; results=(${(f)"$( print -rl (../)#$1(:a:h) )"}) ; done; )  7.38s user 22.29s system 99% cpu 29.960 total
```

bhilburn:
This one is really exciting. I'll be super happy to have this merged into `next`! Nice work, @dritter @onaforeignshore

dritter:
Heads up:
Upsearch with globbing does not work with symlinked paths. I already asked for help on the [mailing list](http://www.zsh.org/mla/users/2018/msg00698.html). If anyone here can help out, that would be great.

dritter:
Test-Folder: /tmp/xxx/1/12/123/1234/12345/123456/1234567/gap dir/ülü/hello/this/it/is
Command: `time (repeat 10000; do; __p9k_upsearch .shorten_folder_marker >/dev/null; done)`

#### Upsearch with loop
Upsearch Code:
```zsh
local -a results
local currentDir="${PWD}"

while [[ "${currentDir}" != / ]]; do
  currentDir="${currentDir:h}"
  [[ -f "${currentDir}/${1}" || -d "${currentDir}/${1}" ]] && results+=("${currentDir}")
done

# Print in lexical order
print -l ${(@O)results}
```
**2,29s user 1,95s system 99% cpu 4,265 total**

#### Upsearch with find
Upsearch-Code
```zsh
  local currentPath="${PWD}"
  local -a paths
  paths=()
   while [[ -n "${currentPath}" ]]; do
    # Remove the last dir part (delete after last slash)
    currentPath="${currentPath%/*}"
    # Add to paths array
    paths+=("${currentPath}")
  done
   # Search in all Paths for file. The paths are escaped
  # automatically by ZSH. We just need to pass every value
  # from the array ((@) modifier).
  # command find "${(@)paths}" -maxdepth 1 -name "${1}"
  command find "${(@)paths}" -maxdepth 1 -name "${1}" 2>/dev/null
```
**18,18s user 27,20s system 91% cpu 49,649 total**

#### Upsearch recursive
Upsearch Code (requires `setopt extended_glob`):
```zsh
  local -a results
  # Search upwards (See https://unix.stackexchange.com/a/64164)
  results=((../)#$1(:a)) 2>/dev/null

  # Print in lexical order
  print -l ${(@o)results}
```
**2,32s user 7,10s system 99% cpu 9,465 total**

#### Upsearch from current master
Upsearch Code:
```zsh
  if [[ "$PWD" == "$HOME" || "$PWD" == "/" ]]; then
    echo "$PWD"
  elif test -e "$1"; then
    pushd .. > /dev/null
    upsearch "$1"
    popd > /dev/null
    echo "$PWD"
  else
    pushd .. > /dev/null
    upsearch "$1"
    popd > /dev/null
  fi
```
**229,82s user 93,12s system 99% cpu 5:24,70 total**

dritter:
I changed the behaviour now to use globs with the cost of a bit of performance. The same test in the same directory now finishes in:
**4,66s user 2,11s system 99% cpu 6,824 total**

For me this is acceptable, considering this test was done in 10000 iterations.

The benefits of my change are:
- You can find multiple files at once. This is especially helpful, because I want to refactor the `truncate_with_package_name` shorten strategy (needs to search for `composer.json` or `package.json`). Before, you had to do the upsearch twice.
- You can specify which type the folder marker is of. The second parameter to `__p9k_upsearch` is the glob qualifier. To search for files only, use `.`, for directories use `/`, for symlinks `@`, for a combination concat them with `,` (and [many more](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Glob-Qualifiers)). See [testUpsearchWithFileGlobs](https://github.com/bhilburn/powerlevel9k/blob/6202c68b3ed5cf3323694d83252f32d55bc5d7b1/test/functions/utilities.spec#L207).
- You can do a fuzzy matching, thanks to globs.

PR #894, opened 2018-07-06
Add vagrant segment

dritter:
This adds a segment for `vagrant` and shows if a VM is running or not. The feature is based on #893, therefore that PR should be merged first.
bildschirmfoto 2018-07-07 um 01 06 32

Unfortunately, I haven't found an icon for vagrant, so the visual identifier is just `V`..


onaforeignshore:
The closest you will get at the moment is the vim V icon in Nerd `\uE62B ´ or `\uF237 ` (V with lines through it)

bhilburn:
Nice work, @dritter =)

Issue #879, opened 2018-06-19
pyenv plugin ignore .python-version file


j9ac9k:
the pyenv plugin does not appear to pick up the current python version.  Looking at the history, it appears that #680 removed the query to `pyenv version` and instead looked directly at the `PYENV_VERSION` environmental variable.

If we look at the pyenv documentation [here](https://github.com/pyenv/pyenv#choosing-the-python-version) we can see that having the environmental variable of `PYENV_VERSION` is only one of the methods to select the current python version.  If no variable is set, one can place a `python-version` file in a directory.  Furthermore, you can set a global python interpreter as well.

This issue was driven by #679, but I think #680 was not the correct fix, as @dinhnv wanted to remove the global environment variant; but that still leaves those of us that are dependent on local `.python-version` files as not being utilized.

Here is a screenshot that hopefully illustrates the issue.

![screen shot 2018-06-19 at 8 55 33 am 1](https://user-images.githubusercontent.com/646398/41608943-89404ca2-739e-11e8-9f02-a9f3c49a4a23.png)

Ideally the pyenv plugin would show that I'm using the `pytest-book` python version.

I would be happy to attempt a pull request to address this.

bhilburn:
This issue was fixed by the merge of #900! Let us know if you are still seeing issues!

PR #900, opened 2018-07-11
Fix `pyenv` segment

### LINKED ISSUE #839: Pyenv version not read from application-specific .python-version file
Hey there,

If `>pyenv version` would result in a non-system wide version set through an application specific .pyenv-version file (output: `my-version-name (set by /Folder/to/current/folder/.python-version)`), the pyenv prompt element does not show up.

This is because the script does not check for the outcome of `>pyenv version` but only for a `$PYTHON_VERSION` environment variable being set. Pyenv virtual environments that are not system wide can also be set through the .python-version file present in the root folder of a project as described in the following section: [pyenv#choosing-the-python-version](https://github.com/pyenv/pyenv#choosing-the-python-version)

```
# pyenv: current active python version (with restrictions)
# https://github.com/pyenv/pyenv#choosing-the-python-version
prompt_pyenv() {
  if [[ -n "$PYENV_VERSION" ]]; then
    "$1_prompt_segment" "$0" "$2" "blue" "$DEFAULT_COLOR" "$PYENV_VERSION" 'PYTHON_ICON'
  fi
}
```

AlexaraWu:
Fix #839 #879.

* Get the version being used by taking the output of the `pyenv version-name` command.
* Get the global Python version by reading global  `$(pyenv root)/version` file.
* Add variable `POWERLEVEL9K_PYENV_PROMPT_ALWAYS_SHOW` similar to `rbenv`.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1264,7 +1264,7 @@ prompt_ram() {
 # Segment to display rbenv information
 set_default POWERLEVEL9K_RBENV_PROMPT_ALWAYS_SHOW false
 prompt_rbenv() {
-  if ! [ -x "$(command -v rbenv)" ]; then
+  if [ $commands[rbenv] ]; then
     local rbenv_version_name="$(rbenv version-name)"
     local rbenv_global="$(rbenv global)"
 
@@ -1632,7 +1632,7 @@ prompt_virtualenv() {
 # https://github.com/pyenv/pyenv#choosing-the-python-version
 set_default POWERLEVEL9K_PYENV_PROMPT_ALWAYS_SHOW false
 prompt_pyenv() {
-  if ! [ -x "$(command -v pyenv)" ]; then
+  if [ $commands[pyenv] ]; then
     if [[ -n "$PYENV_VERSION" ]]; then
       "$1_prompt_segment" "$0" "$2" "blue" "$DEFAULT_COLOR" "$PYENV_VERSION" 'PYTHON_ICON'
     else
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1270,7 +1270,7 @@ prompt_rbenv() {
   elif [ $commands[rbenv] ]; then
     local rbenv_version_name="$(rbenv version-name)"
     local rbenv_global="$(rbenv global)"
-    if ! [[ $rbenv_version_name == $rbenv_global && "$POWERLEVEL9K_RBENV_PROMPT_ALWAYS_SHOW" = false ]]; then
+    if [[ "${rbenv_version_name}" != "${rbenv_global}" || "${POWERLEVEL9K_RBENV_PROMPT_ALWAYS_SHOW}" == "true" ]]; then
       "$1_prompt_segment" "$0" "$2" "red" "$DEFAULT_COLOR" "$rbenv_version_name" 'RUBY_ICON'
     fi
   fi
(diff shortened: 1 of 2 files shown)


powerlevel9k.zsh-theme
```diff
@@ -1628,11 +1628,21 @@ prompt_virtualenv() {
...
+# Segment to display pyenv information
 # https://github.com/pyenv/pyenv#choosing-the-python-version
+set_default POWERLEVEL9K_PYENV_PROMPT_ALWAYS_SHOW false
 prompt_pyenv() {
-  if [[ -n "$PYENV_VERSION" ]]; then
-    "$1_prompt_segment" "$0" "$2" "blue" "$DEFAULT_COLOR" "$PYENV_VERSION" 'PYTHON_ICON'
+  if ! [ -x "$(command -v pyenv)" ]; then
+    if [[ -n "$PYENV_VERSION" ]]; then
+      "$1_prompt_segment" "$0" "$2" "blue" "$DEFAULT_COLOR" "$PYENV_VERSION" 'PYTHON_ICON'
+    else
```
dritter: That could be an `elif ! [ -x "$(command -v pyenv)" ]; then`. That way we could save the outer if.. I see that you've done it because the `$PYENV_VERSION` should only be there, if `pyenv` is there too, but IMHO if `$PYENV_VERSION` is set, we should output that information as quickly as possible.

powerlevel9k.zsh-theme
```diff
@@ -1628,11 +1628,21 @@ prompt_virtualenv() {
...
+      local pyenv_version_name="$(pyenv version-name)"
+      local pyenv_global="$(pyenv version-file-read $(pyenv root)/version)"
+      if [[ $pyenv_version_name == $pyenv_global && "$POWERLEVEL9K_PYENV_PROMPT_ALWAYS_SHOW" = false ]]; then
```
dritter: If we flip that condition, we could save that return. That would be probably more readable, no?

powerlevel9k.zsh-theme
AlexaraWu: What if I incorrectly used a `.zshrc` file including something like `export PYENV_VERSION=3.5.5` without `pyenv` installed?

dritter: What if someone installed `pyenv`, but never calls it and set `PYENV_VERSION=1.2.3`? What if someone wrote a script and called it `pyenv`?

I think there are always scenarios we cannot handle. That is why I would rely on the variable in the first place..

AlexaraWu: Fine, maybe I thought too much. :smile:

AlexaraWu:
@dritter PTAL

powerlevel9k.zsh-theme
```diff
@@ -1628,11 +1626,18 @@ prompt_virtualenv() {
...
+# Segment to display pyenv information
 # https://github.com/pyenv/pyenv#choosing-the-python-version
+set_default POWERLEVEL9K_PYENV_PROMPT_ALWAYS_SHOW false
 prompt_pyenv() {
   if [[ -n "$PYENV_VERSION" ]]; then
     "$1_prompt_segment" "$0" "$2" "blue" "$DEFAULT_COLOR" "$PYENV_VERSION" 'PYTHON_ICON'
+  elif [ $commands[pyenv] ]; then
+    local pyenv_version_name="$(pyenv version-name)"
+    local pyenv_global="$(pyenv version-file-read $(pyenv root)/version)"
+    if ! [[ $pyenv_version_name == $pyenv_global && "$POWERLEVEL9K_PYENV_PROMPT_ALWAYS_SHOW" = false ]]; then
```
dritter: I thought more of `if [[ "${pyenv_version_name}" != "${pyenv_global}" || "${POWERLEVEL9K_PYENV_PROMPT_ALWAYS_SHOW}" == "true" ]]`. But flipping the entire condition should work as well..

AlexaraWu: `if [[ "${pyenv_version_name}" != "${pyenv_global}" || "${POWERLEVEL9K_PYENV_PROMPT_ALWAYS_SHOW}" == "true" ]]` would be more readable.

bhilburn:
@AlexaraWu - Thanks so much for putting this PR together, and then working with us through the review process to get it ready for merge! I'm really happy this will be in our upcoming v0.6.6 release. Welcome to P9k as a first-time contributor! `=)`

@dritter - Thanks for the great review work and merge recommendation! Much appreciated =)

PR #901, opened 2018-07-11
[bugfix] Zsh < 5.3: Replace [[-v with ((${+

robobenklein:
For zsh 5.1 compatibility.

From https://github.com/bhilburn/powerlevel9k/pull/830/files#diff-ea5a405f0331722fc10a832209ffe20bL490 it only worked in 5.3+.

I'm not sure if this is better or if the `defined` util function should be preferred?

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -498,7 +498,7 @@ prompt_battery() {
     fi
   fi
   # return if POWERLEVEL9K_BATTERY_HIDE_ABOVE_THRESHOLD is set and the battery percentage is greater or equal
-  if [[ -v "POWERLEVEL9K_BATTERY_HIDE_ABOVE_THRESHOLD" && "${bat_percent}" -ge $POWERLEVEL9K_BATTERY_HIDE_ABOVE_THRESHOLD ]]; then
+  if defined POWERLEVEL9K_BATTERY_HIDE_ABOVE_THRESHOLD && [[ "${bat_percent}" -ge $POWERLEVEL9K_BATTERY_HIDE_ABOVE_THRESHOLD ]]; then
     return
   fi
 


dritter:
IMHO we should use the `defined` function as this is the way we do it everywhere. `[[ -v ${VAR} ]]` may be the future, but until we raise our minimum ZSH requirement, we should not use another way.

robobenklein:
Alright, I'll change it again if that's still not right.

robobenklein:
https://travis-ci.org/bhilburn/powerlevel9k/builds/402849456#L641

"Parse Error"? Why did it get an exit 0 then?

Also, test seems to run on zsh 5.0.2?

robobenklein:
@dritter Here's why my first thought was to use the `((${+` check instead of the `[[ -v` or `typeset -p` options, I made a short script which hopefully shows the reasoning:

```zsh
#!/bin/zsh

N=2000000
local SOME_VARIABLE
SOME_VARIABLE="something"

TIMEFMT=$'\nreal\t%E\nuser\t%U\nsys\t%S'

echo
echo "typeset -p:"
time (repeat $N { typeset -p "SOME_VARIABLE" > /dev/null 2>&1 })

echo
echo '${+var}:'
time (repeat $N { (( ${+SOME_VARIABLE} )) })

echo
echo '[[ -v var ]]'
time (repeat $N { [[ -v SOME_VARIABLE ]] })

echo
echo "Variable NOT set: "
unset SOME_VARIABLE

echo
echo "typeset -p:"
time (repeat $N { typeset -p "SOME_VARIABLE" > /dev/null 2>&1 })

echo
echo '${+var}:'
time (repeat $N { (( ${+SOME_VARIABLE} )) })

echo
echo '[[ -v var ]]'
time (repeat $N { [[ -v SOME_VARIABLE ]] })

```

And the output:

```

typeset -p:

real	19.24s
user	9.36s
sys	9.66s

${+var}:

real	0.95s
user	0.95s
sys	0.00s

[[ -v var ]]

real	0.31s
user	0.31s
sys	0.00s

Variable NOT set: 

typeset -p:

real	23.28s
user	13.19s
sys	9.88s

${+var}:

real	0.94s
user	0.94s
sys	0.00s

[[ -v var ]]

real	0.28s
user	0.28s
sys	0.00s

```

onaforeignshore:
These are my results on zsh 5.5.1 (x86_64-apple-darwin17.5.0)

typeset -p:

real	84.15s
user	36.93s
sys	47.00s

${+var}:

real	1.46s
user	1.44s
sys	0.01s

[[ -v var ]]

real	0.62s
user	0.61s
sys	0.01s

Variable NOT set:

typeset -p:

real	91.43s
user	43.60s
sys	47.67s

${+var}:

real	1.39s
user	1.38s
sys	0.00s

[[ -v var ]]

real	0.58s
user	0.58s
sys	0.00s

robobenklein:
Hey @onaforeignshore could you verify that with the proposed changes in #905 that the `defined` function really performs 30x better on OSX? It's kind of hard to believe but I don't have other MacOS computers to test on.

If so, I'll propose that the change to `(( ${+` can wait until a full release, and we can use #905 as a quick patch for right now?

bhilburn:
Merged as part of #905!

Issue #898, opened 2018-07-10
Context Segment Colors envvars not being used


robobenklein:
#### Describe Your Issue
What is happening?

![image](https://user-images.githubusercontent.com/5423266/42534105-d386c414-8459-11e8-8195-2d1be00cdf70.png)
![image](https://user-images.githubusercontent.com/5423266/42534368-8f4d3598-845a-11e8-9c3e-1cdff4858474.png)
(Ubuntu 16.04)
![image](https://user-images.githubusercontent.com/5423266/42534570-ff46d444-845a-11e8-86f4-5ba08fc833d8.png)
(RHEL 7.5 - I realize zsh <5.1 isn't supported? https://github.com/bhilburn/powerlevel9k/issues/877 )

Note - does *not* occur on my local machine:

![image](https://user-images.githubusercontent.com/5423266/42534239-3725a224-845a-11e8-86c8-7c73cb8b9ec2.png)
(Ubuntu 18.04)

#### Have you tried to debug or fix it?

This occurred just today as of the 0.6.5 release. It was working this morning, before I ran the update. I couldn't find any reason this broke locally.

I haven't seen any change notes in the Github release that specified a greater ZSH requirement, but to me it looks like I would just prefer the previous / older release since I didn't have any problems there.

#### Environment Information

P9K is being managed by Zplugin, I've tried fresh reinstalls and on multiple systems.

Systems shown in screenshots use the same ZSHRC and variables to start.

bhilburn:
Uh oh, looks like we might be tagging a v0.6.6? @dritter 

@robobenklein - Can you clarify what version of ZSH you are running? Is this < 5.1, based on your comment about RHEL 7.5?

@dritter - Might be good to roll back to our own function for legacy OS support.

dritter:
Hi @robobenklein !

Thanks for that lightning-fast report. I looked a bit into this, and it looks like this is not a bug, but a feature ;)
The `context` segment is now aware of `sudo` powers. So if you recently did something with sudo, this state is in charge. You can test this behaviour with `sudo -n true && echo "true" || echo "false"`, so you are probably changing the color for the wrong state.

Old states:
`ROOT` - If you are root
`DEFAULT`

New states:
`ROOT` - If you are root
`SUDO` - If you got elevated rights via sudo
`DEFAULT`
`REMOTE` - If you are via SSH in the machine
`REMOTE_SUDO` - If you got elevated rights and SSHed in the machine

[...]

PR #902, opened 2018-07-11
Add new states to context segment in README

dritter:
Explains the states of the `context` segment in the README. Related to #898


bhilburn:
Thanks so much for fixing the docs, @dritter, after our discussion in #898! Much appreciated.

PR #904, opened 2018-07-12
FIx status segment

### LINKED ISSUE #903: Error icon does not work
When I execute an incorrect order in the terminal, it does not show me the error icon, instead it shows me the correct icon.

Screenshot:
Sreenshot error
---
 I do not know how they are handling the errors, but I wanted it to work similar to this:

Like this

This is in my `.zshrc` file:
```
##----------------- Powerline9k Customization
POWERLEVEL9K_MODE='awesome-fontconfig'
POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(context status dir vcs)
POWERLEVEL9K_DISABLE_RPROMPT=true
POWERLEVEL9K_CONTEXT_TEMPLATE="\uF17C" # Linux icon  (%n@%m) default

POWERLEVEL9K_STATUS_CROSS=true
POWERLEVEL9K_STATUS_OK=true
##----------------

ZSH_THEME="powerlevel9k/powerlevel9k"
```

I'musing `Oh-my-zsh`
in = `Debian 9`
PL9K version = `v0.6.5`
font: `Roboto Medium for Powerline`, 
terminal: `Terminator`, 
zsh v. =   `v5.3.1`

dritter:
This fixes #903 . The problem here was that I made some variables local, but `RETVAL` must be global. In `powerlevel9k_prepare_prompts` (precmd hook) we save the return value of the last executed command. This must be done at the very top of that function, otherwise we get the return value of another command. E.g.:
```
local xxx
RETVAL=$?
```
will always have a RETVAL of 0, because it gets the return value of `local xxx`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,3 +30,4 @@ script:
   - test/segments/vcs.spec
   - test/segments/kubecontext.spec
   - test/segments/laravel_version.spec
+  - test/segments/status.spec
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1779,10 +1779,16 @@ powerlevel9k_preexec() {
 
 set_default POWERLEVEL9K_PROMPT_ADD_NEWLINE false
 powerlevel9k_prepare_prompts() {
-  local RETVAL RPROMPT_PREFIX RPROMPT_SUFFIX
+  # Return values. These need to be global, because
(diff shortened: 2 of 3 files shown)


bhilburn:
Thanks so much for putting together a fix for #903 so quickly, @dritter! Plus, we get even further improvement with segment tests. Excellent work, as usual `=)`

PR #912, opened 2018-07-14
[Bugfix] Fix error in laravel_version when artisan is not present

brandon14:
If PHP could not find the artisan file when building the prompt, it would output `Could not open input file: artisan` to stdout. Where only stderr was being redirected to /dev/null, this output would get placed into `laravel_version` variable and displayed on the prompt.

Also the output of `php artisan --version` follows the format `Laravel Framework <version number>`, so I adjusted pulling just the version number out accordingly.

The reference PR and conversation about this error can be found here: https://github.com/bhilburn/powerlevel9k/pull/818

On a side note, this is hands down my favorite zsh theme!

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1118,7 +1118,7 @@ prompt_vpn_ip() {
 ################################################################
 # Segment to display laravel version
 prompt_laravel_version() {
-  local laravel_version="$(php artisan --version 2> /dev/null"
+  local laravel_version="$(php artisan --version 2> /dev/null)"
   if [[ -n "${laravel_version}" && "${laravel_version}" =~ "Laravel Framework" ]]; then
     # Strip out everything but the version
     laravel_version="${laravel_version//Laravel Framework /}"
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1118,8 +1118,8 @@ prompt_vpn_ip() {
 ################################################################
 # Segment to display laravel version
 prompt_laravel_version() {
-  local laravel_version="$(php artisan --version 2>&1 | grep -oe '^Laravel Framework [0-9.]*')"
-  if [[ -n "${laravel_version}" ]]; then
+  local laravel_version="$(php artisan --version 2> /dev/null"
+  if [[ -n "${laravel_version}" && "${laravel_version}" =~ "Laravel Framework" ]]; then
     # Strip out everything but the version
     laravel_version="${laravel_version//Laravel Framework /}"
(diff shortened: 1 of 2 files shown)


onaforeignshore:
Does `php artisan --version` return multiple lines? On my clean install it returns one line. Just wondering because you use `grep` and we try to steer clear of externals as much as possible

brandon14:
@onaforeignshore it does return only one line. Ah yeah that makes sense. I was just following after the php version segment, but i suppose a more portable way could be done

bhilburn:
@brandon14 - Thanks so much for putting together this PR! I also appreciate you modeling the changes on other segments, which should generally be a good idea, hah. Unfortunately it seems you found what should probably be a bug in the `php` segment, per @onaforeignshore's note about calling out to external programs `=)`

@dritter has identified this PR as a priority and a good fit for the v0.6.6 release, so I'd like to get this cleaned up and merged as soon as we can!

Since the command only returns one line, I think we ought to be able to do the parsing with some ZSH regex. @onaforeignshore - agree / disagree?

powerlevel9k.zsh-theme
```diff
@@ -1118,11 +1118,10 @@ prompt_vpn_ip() {
 ################################################################
 # Segment to display laravel version
 prompt_laravel_version() {
-  local laravel_version="$(php artisan --version 2>/dev/null)"
+  local laravel_version="$(php artisan --version 2>&1 | grep -oe '^Laravel Framework [0-9.]*')"
   if [[ -n "${laravel_version}" ]]; then
```
onaforeignshore: We can drop the `grep` by changing the if statement as follows:
```zsh
  local laravel_version="$(php artisan --version 2>&1 /dev/null)"
  if [[ -n "${laravel_version}" && "${laravel_version}" =~ "Laravel Framework" ]]; then
```

brandon14:
Yeah that seems to work fine @onaforeignshore . Ill get those changes up ASAP

bhilburn:
Thanks so much for pulling in that change, @brandon14, and for suggesting it, @onaforeignshore! I think the code looks great.

We do have one last problem, where a ton of the unit tests are failing with this error:
```
command not found: build_left_prompt
```

I can't see how the changeset in this PR could possibly cause those errors, but `master` is currently passing. I'm not super sure what to make of it. @onaforeignshore @dritter - anything pop out at you?

brandon14:
@dritter @bhilburn sorry about that I didn't see the missed parenthesis there. It's been a busy weekend for me or I would have fixed it sooner. My apologies.

bhilburn:
Thanks so much, @dritter! I totally missed that line.

And thanks, @brandon14, for fixing the issue! I'm really happy we are getting this merged, and it will be in the v0.6.6 release. Welcome to P9k as a first-time contributor! `=)`

Issue #907, opened 2018-07-13
Strange behaviour with truncate_to_unique


G4Zz0L1:
I've a strange behaviour with the option `POWERLEVEL9K_SHORTEN_STRATEGY="truncate_to_unique"`

If I set it that way the prompt returns
`prompt_dir:66: no matches found: //.*/`
if the folder is a hidden one, or 
`prompt_dir:66: no matches found: //D*/`
if the folder starts with D

But it's pretty random, in some folder it returns that and in other works normally

If i remove that row from the .zshrc it works flawlessly in every folder

gnome-terminal 3.28.1-1ubuntu1mint2
zsh 5.4.2 (x86_64-ubuntu-linux-gnu)
oh-my-zsh and p9k cloned from git, latest

nvonbulow:
I am having this same issue

bhilburn:
@G4Zz0L1 - Thanks so much for the bug report, and thanks for confirming, @nvonbulow!

@onaforeignshore has already put together a fix. Can you check out #915, give that PR a shot, and let us know if it resolves your issue?

diraol:
It kind of works for me too, but with a strange behavior regarding the beginning of the path printed on the screen:
![2018-07-18-115453_313x69_scrot](https://user-images.githubusercontent.com/192702/42889725-7cb8c334-8a81-11e8-8739-498033ae331c.png)

That `~h/d/` seems wrong to me, it should be `~/` or `/h/d/`

onaforeignshore:
What is the content of your `$HOME` variable?

nvonbulow:
I'm having the same weird path beginning as @diraol with the unwanted `~h`. My `$HOME` is set to `/home/nick`

onaforeignshore:
I have updated the fix. Can you please pull it again and see if the issue persists?

nvonbulow:
The issue is still there. It doesn't look like anything changed

diraol:
It is better now for me.

nvonbulow:
With the addition of the code simplification commit, everything works for me now.

onaforeignshore:
@nvonbulow That's good news.

bhilburn:
Thanks so much, @nvonbulow and @diraol, for testing the fix in #915! And thanks for filing the initial bug report, @G4Zz0L1! Great work, everyone.

The fix #915 has been merged into master `=)`

PR #915, opened 2018-07-17
[Bugfix] Fix for "truncate_to_unique" strategy

onaforeignshore:
Bugfix for issue #907 

The issue is that this strategy requires absolute paths to work, which is usually only available when `POWERLEVEL9K_DIR_PATH_ABSOLUTE` is set to true. This fix works around this limitation by using absolute paths and then changing the path at the end to `~/...` format if required.

This requires the use of a helper function `getUniqueHomeFolder` to determine the unique name of the home folder in order to replace that with `~`.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -756,9 +756,8 @@ getUniqueHomeFolder() {
     local -a matching
     local -a paths
     local cur_path='/'
-    # the first time we run the script, the working directory *should* be the home folder
-    [[ -z $HOME ]] && paths=${PWD} || paths=$HOME
-    paths=(${(s:/:)paths})
+    # all users have the $HOME variable set automatically... see http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html
+    paths=(${(s:/:)HOME})
     for directory in ${paths[@]}; do
       test_dir=''
       for (( i=0; i < ${#directory}; i++ )); do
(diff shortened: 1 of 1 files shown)


powerlevel9k.zsh-theme
```diff
@@ -744,6 +744,38 @@ prompt_command_execution_time() {
...
+# set the true.
+#
+local home_path=""
+getUniqueHomeFolder() {
+  if [[ "$home_path" == "" ]]; then
+    local trunc_path directory test_dir test_dir_length
+    local -a matching
+    local -a paths
+    local cur_path='/'
+    # the first time we run the script, the working directory *should* be the home folder
```
bhilburn: This comment worries me a little bit - can you expand on what you mean? If the segment doesn't get drawn for the first time from `/home/<user>`, will this code fail?

G4Zz0L1: I've tested it, even if the start folder isn't my home, it seems to work

image

onaforeignshore: Usually, the terminal starts in the home folder, which is why I stated *should*, but if you check the next line, we test for the existence $HOME and use that, otherwise use ${PWD}.

onaforeignshore:
I rewrote it to simplify this a lot. Since we use the code to find a truncated path twice, I simply made it a function and passed `$HOME` and then `$PWD` to it. Think it makes more sense now.

diraol:
It is working nice to me!

bhilburn:
@onaforeignshore - I like the new changeset you pushed a lot! Great work simplifying the code.

At this point, it's clear that we've had a lot of people test this PR, and everyone is reporting success. Great work, @onaforeignshore! Super happy we got this resolved quickly `=)`

PR #916, opened 2018-07-17
Additional fix to `pyenv` segment

AlexaraWu:
File `$(pyenv root)/version` may not be created if the `pyenv global` command never be used.


bhilburn:
Thanks for the quick fix-up, @AlexaraWu! Good call =)

PR #919, opened 2018-07-19
[Enhancement] Add option to always show remote branch

JulienPivard:
Add `POWERLEVEL9K_VCS_GIT_ALWAYS_SHOW_REMOTE_BRANCH='1'` to your zshrc if you would to see remote branch in any cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -663,6 +663,7 @@ customization is provided via:
 |`P9K_VCS_HG_HOOKS`|`(vcs-detect-changes)`|Layout of the segment for Mercurial repositories.|
 |`P9K_VCS_SVN_HOOKS`|`(vcs-detect-changes svn-detect-changes)`|Layout of the segment for SVN repositories.|
 |`P9K_VCS_ACTIONFORMAT_FOREGROUND`|`red`|The color of the foreground font during actions (e.g., `REBASE`).|
+|`P9K_VCS_GIT_ALWAYS_SHOW_REMOTE_BRANCH`|`false`|Set to true If you would to always see the remote branch.|
 
 
 ##### vcs symbols
--- a/segments/vcs.p9k
+++ b/segments/vcs.p9k
@@ -160,6 +160,7 @@ function +vi-git-aheadbehind() {
   hook_com[misc]+=${(j::)gitstatus}
 }
 
+p9k::set_default P9K_VCS_GIT_ALWAYS_SHOW_REMOTE_BRANCH false
 function +vi-git-remotebranch() {
   local remote branch_name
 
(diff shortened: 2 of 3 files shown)


zikes:
That's awesome! Would you mind adding a test for it in https://github.com/bhilburn/powerlevel9k/blob/master/test/segments/vcs.spec? You should be able to run the tests via Docker, instructions are at https://github.com/bhilburn/powerlevel9k/blob/master/TESTS.md

JulienPivard:
Tests add. Tell me if it's ok for you.

JulienPivard:
I will go to fix this quickly :)

bhilburn:
@JulienPivard - Fantastic! Let us know if you need any help or guidance as you port to the new codebase! I see Travis is currently failing with the git_vcs tests:

```
ASSERT:expected:<%K{002} %F{000} test test %k%F{002}%f > but was:<%K{002} %F{000} test→origin/test test %k%F{002}%f >
testAlwaysShowRemoteBranch
testAlwaysShowRemoteBranch:10: command not found: build_left_prompt
ASSERT:expected:<%K{002} %F{000} master →origin/master %k%F{002}%f > but was:<>
shunit2:ERROR testAlwaysShowRemoteBranch() returned non-zero return code.
```

JulienPivard:
Thanks :) I think I would need more explanation about the script `test-in-docker` because I find the explanation given at https://github.com/bhilburn/powerlevel9k/blob/master/TESTS.md is a little fuzzy

dritter:
Hi @JulienPivard !

There are four different ways to test P9K:
1. Most obvious: Directly in your shell with your environment; Like you use it.
2. Old way: With vagrant and virtualbox. This starts a whole operating system in a virtual machine. This is still my way to go, if I want to test on BSD. Downside: Just one ZSH version available..
3. New way: With `test-in-docker`. This way starts a docker container with the ZSH version and ZSH Framework you want. This is my way to go, if I want to test something on a specific ZSH version. Because this is more flexible, I use this method if I want a different (more limited) environment than my own machine.
4. Automatic way: We have a lot of test scripts that test specific parts of P9K everytime somebody pushes a commit. The tests are executed by Travis, an external Service that we connected with this Repo. You could execute these tests on your machine locally as well, we added `shUnit` as git submodule. After you installed the submodules, it is possible to simply execute the test scripts (see `test/` folder).

So, option 1-3 are for manual testing, but it is possible to execute the test scripts in that environments as well.

As @bhilburn mentioned, the tests in `test/segments/vcs_git.sh` fail, most probably on every environment. It would be the easiest thing to execute that test locally. If they don't fail on your machine, you could still check them in a more limited environment.

Hope that clarifies how to test a bit.

JulienPivard:
Ok my bad, I forgot that zsh reported the line number in the function and not in the file.
No problem it's work

dritter:
Thx @JulienPivard

PR #923, opened 2018-07-20
Fix font issue debugging script

dritter:
The new GitHub Issue Templates show that a lot of folks have trouble with fonts and problems with the `debug/font-issues.zsh` script. The script is a port of the font detection from [dylanaraps/neofetch](https://github.com/dylanaraps/neofetch). In the our previous version I missed some functions (`get_ppid` and `get_process_name`) that now cause trouble.
Last but not least, I updated the font detections functions.

It would be super helpful if somebody could test this under WSL and Linux.

### DIFF
--- a/debug/font-issues.zsh
+++ b/debug/font-issues.zsh
@@ -1,6 +1,21 @@
 #!/usr/bin/env zsh
 #vim:ft=zsh ts=2 sw=2 sts=2 et fenc=utf-8
 
+source functions/colors.zsh
+source functions/icons.zsh
+source functions/utilities.zsh
+# Map our $OS to neofetch $os
+os="$OS"
+
+
+trim() {
+    set -f
+    # shellcheck disable=2048,2086
+    set -- $*
+    printf '%s\n' "${*//[[:space:]]/ }"
+    set +f
+}
+
 get_ppid() {
     # Get parent process ID of PID.
     case "$os" in
--- a/debug/font-issues.zsh
+++ b/debug/font-issues.zsh
@@ -12,7 +12,7 @@ trim() {
     set -f
     # shellcheck disable=2048,2086
     set -- $*
-    printf '%s\n' "${*//[[:space:]]/ }"
+    printf '%s\n' "${*//[[:space:]]/}"
     set +f
 }
 
--- a/debug/font-issues.zsh
+++ b/debug/font-issues.zsh
@@ -1,111 +1,221 @@
 #!/usr/bin/env zsh
 #vim:ft=zsh ts=2 sw=2 sts=2 et fenc=utf-8
 
+get_ppid() {
+    # Get parent process ID of PID.
+    case "$os" in
+        "Windows")
+            ppid="$(ps -p "${1:-$PPID}" | awk '{printf $2}')"
+            ppid="${ppid/PPID}"
+        ;;
+
+        "Linux")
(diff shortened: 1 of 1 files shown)


JulienPivard:
All seems to be ok with Macos.

But with my Linux (debian) I have an error with ps.
```
error: improper list

Usage:
 ps [options]
...
```
So I take a look in your code and it seems the `$os` is not set and if I add `os=$(uname)` at the begining of the scripts the OS detection work. But I have other problem `trim` is not found, but is the use of this command necessary ? The ideal would be to avoid having to install soft and more to run your script.

JulienPivard:
If I add that function at the begining of the script 
``` shell
trim()
{
    local var="${*}"
    # remove leading whitespace characters
    var="${var#"${var%%[![:space:]]*}"}"
    # remove trailing whitespace characters
    var="${var%"${var##*[![:space:]]}"}"
    echo -n "${var}"
}
```
I have 
```
===== Font debugging =====
You are using gnome-terminal with Font 

WARNING It does not seem like you use an Powerline-enabled or Awesome Terminal Font!
Please make sure that your font settings are correct!
``̀`
I use a powerline font so I tool another look in your script and in the function `get_term_font` it have no case for gnome-terminal

**Comment by dritter** [comment]
Hah! Good point @JulienPivard !

As this is a port from neofetch, I'll have to lookup how they resolve their `$OS`. I overlooked this bug, because with a running P9K, there is a `$OS` present as well (but with other values probably).

I'll have a look into that tomorrow.

**Comment by JulienPivard** [comment]
I did not know there was a $OS variable. But in the script there is `$os` and not `$OS` I think the problem is there. I take a look and `$os` return nothing but `$OS` return OSX for my mac and Linux for my debian. So i think it's juste a problem with the case :)

**Comment by dritter** [comment]
You are right. [Here](https://github.com/dylanaraps/neofetch/blob/master/neofetch#L780-L811) is the code that fills `$os`. Our variable is names `$OS`.
So, either we want to change `$os` to `$OS` and make the debugging script compatible with our values, or we want to add the `get_os` function as well. But that are thoughts for tomorrow. ;)

**Comment by dritter** [comment]
@JulienPivard I (hopefully) fixed the font debugging script. Could you give it another try?

**Comment by JulienPivard** [comment]
```
get_process_name:10: aucun fichier ou dossier de ce type: /proc/ 1960/comm
grep: /proc/ 1960/status: No such file or directory
===== Font debugging =====
You are using  with Font 

WARNING It does not seem like you use an Powerline-enabled or Awesome Terminal Font!
Please make sure that your font settings are correct!
```

I think trim don't remove all the space.

**Comment by JulienPivard** [comment]
It's work... for trim

```
===== Font debugging =====
You are using gnome-terminal with Font 

WARNING It does not seem like you use an Powerline-enabled or Awesome Terminal Font!
Please make sure that your font settings are correct!
```

JulienPivard:
if I have the time I would try with other terminals.

bhilburn:
Great idea, @dritter! This should help quite a bit. I would love to get this in for v0.6.6 to help with debugging going forward. Would you like to do further testing or is this ready to go?

@julienfalque - Thanks so much for spending the time to help us test this! That is extremely valuable for the project =)

bhilburn:
This is now fixed in master, and the v0.6.6 release will include the fix. Thanks so much!

PR #924, opened 2018-07-21
[New Segment] TaskWarrior

nzachow:
# Description

This is a new segment created to display data from TaskWarrior.
The segment was named **tw**.

img-2018-07-21-023435

Data is obtained with:

`today=$(task status:pending +DUETODAY count)`
`over=$(task status:pending +OVERDUE count)`
`pending=$(task status:pending count)`

The icon used is: TODO_ICON

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1517,6 +1517,22 @@ prompt_todo() {
   fi
 }
 
+# taskwarrior: show data from taskwarrior
+prompt_tw() {
+  if $(hash task 2>&-); then
+    today=$(task status:pending +DUETODAY count)
+    over=$(task status:pending +OVERDUE count)
+    pending=$(task status:pending count)
+    if [[ "$pending" = <-> ]]; then
+      "$1_prompt_segment" "$0" "$2" "244" "$DEFAULT_COLOR" "Overdue:$over Today:$today Pending:
+$pending" 'TODO_ICON'
+    else
+      "$1_prompt_segment" "$0" "$2" "244" "$DEFAULT_COLOR" "No pending tasks!" 'TODO_ICON'
+    fi
+  fi
+}
+
+
 ################################################################
 # VCS segment: shows the state of your repository, if you are in a folder under
 # version control
--- a/README.md
+++ b/README.md
@@ -152,6 +152,7 @@ The segments that are currently available are:
   output of an arbitrary command.
 * [`command_execution_time`](#command_execution_time) - Display the time the current command took to execute.
 * [`todo`](http://todotxt.com/) - Shows the number of tasks in your todo.txt tasks file.
+* [`tw`](#tw) - Display count of pending tasks using data from TaskWarrior.
 * `detect_virt` - Virtualization detection with systemd
 * `newline` - Continues the prompt on a new line.
 * `openfoam` - Shows the currently sourced [OpenFOAM](https://openfoam.org/) environment.
@@ -627,6 +628,14 @@ segment, as well:
 # Output time, date, and a symbol from the "Awesome Powerline Font" set
 POWERLEVEL9K_TIME_FORMAT="%D{%H:%M:%S \uE868  %d.%m.%y}"
 ```
+
+##### tw
+
+This segment add information about pending tasks from TaskWarrior. Data is separated in three
+different classes: overdue, due today and pending.
+
+More information about TaskWarrior can be found [`here`](https://taskwarrior.org/).
+
 ##### user
 
 The `user` segment will print the username.


powerlevel9k.zsh-theme
```diff
@@ -1505,6 +1517,22 @@ prompt_todo() {
   fi
 }
 
+# taskwarrior: show data from taskwarrior
+prompt_tw() {
+  if $(hash task 2>&-); then
+    today=$(task status:pending +DUETODAY count)
```
dritter: This could be a local variable. No need to be global.

powerlevel9k.zsh-theme
```diff
@@ -1505,6 +1517,22 @@ prompt_todo() {
   fi
 }
 
+# taskwarrior: show data from taskwarrior
+prompt_tw() {
+  if $(hash task 2>&-); then
```
dritter: You are calling `task` four times in your segment. Calling external programs is usually quite slow. Is there a way to fire one command, gathering all necessary data and searching in that data afterwards?

powerlevel9k.zsh-theme
```diff
@@ -1505,6 +1517,22 @@ prompt_todo() {
...
+    today=$(task status:pending +DUETODAY count)
+    over=$(task status:pending +OVERDUE count)
+    pending=$(task status:pending count)
+    if [[ "$pending" = <-> ]]; then
```
dritter: Does that mean that overdue tasks are pending as well? Otherwise you'll miss the overdue ones, if there are no pending ones.
This may be expressed by states as well. Probably if you want to change the color if there are overdue tasks..

powerlevel9k.zsh-theme
nzachow: I understand that this can be a problem but reading through
the TaskWarrior docs I could not find a way to do this. They
have an export function to return data in a JSON. One approach
would be to use `task +PENDING export` and count tasks based
on the "due" value but I don't think this would be faster.
What do you think ?

powerlevel9k.zsh-theme
nzachow: Yes, overdue tasks are pending as well. This means we could
simplify the code to `task +OVERDUE count`. Also there is
a better syntax for pending tasks: +PENDING.
The idea to use states to represent this information seems
really good, however I'm not sure how to go about implementing it.

dritter:
The `test-in-docker` script is for manual testing only (and they are defunct ATM; see #884). To write automated tests you need `shunit`, which is a git module of p9k. Here is a short introduction (which I should copy over to `TESTS.md`..):

### Installation
You need to install the submodule first to be able to run the unit tests. You can do that with `git submodule init && git submodule update`.

### Test location
The tests live under `test/`. Segment specific tests under `test/segments/`.

### Executing tests
The tests are shell scripts on their own. So you can execute them right away.

### General Information
The tests usually have a `setUp()` function which is executed before every test function. Speaking of, test functions must be prefixed with `test`. In the tests, you can do [different Assertions](https://github.com/kward/shunit2#-asserts). It is always a good idea to mock the program you want to test (just have a look at other tests).

### Travis
We use [Travis](https://travis-ci.org/) for Continuous Integration. This service executes our tests after every push. For now, we need to tell travis where to find the tests, which is what happens in the `.travis.yml` file.


Hope that helps.

nzachow:
Thanks @dritter !

That was really helpful.
I managed to create some tests and everything seems to be working now.

powerlevel9k.zsh-theme
onaforeignshore: You can get all the data from a single call, but that requires that you process the JSON data. You can do that as follows:
```zsh
    local -a tasks; tasks=( "${(@Qf)$( task export 2> /dev/null )}" ) # put the JSON file from task into an array, removing quotes
    local date=$( date +"%Y%m%d" 2> /dev/null )
    local today=0 over=0 pending=0
    local -a match
    for line in "${(s:,:)tasks[@]}"; do # split the array into multiple lines on delimiter comma (,)
      match=( ${(s.:.)line} ) # split line into keys and values
      [[ ${match[1]} == "status" && ${match[2]} == "pending" ]] && (( ++pending ))
      if [[ ${match[1]} == "due" ]]; then
        [[ ${match[2]:0:8} == $date ]] && (( ++today )) # tasks due today
        (( ${match[2]:0:8} < $date )) && (( ++over )) # overdue tasks
      fi
    done
```
This has the added advantage that you can get other information, like "priority" = "H" for how many high priority, etc.

onaforeignshore: I don't know if it is faster though, because the JSON data exports all the tasks, including done and deleted, so the array can get massive.

bhilburn:
Hey @nzachow - Thanks so much for filing this PR! It looks like it's kind of gone into a holding pattern since July, but I would really like to get this one finished up and merged into our `next` branch! It's a great contribution - sorry it's been a bit of a pain to tie up.

@nzachow - The `next` branch is a new codebase, and is organized a bit differently (much more cleanly). The only major changes for you will be moving your segment code to its own file in the `segments/` directory, and shortening any env names from `POWERLEVEL9K_` to `P9K_`.

@dritter @onaforeignshore - Was there anything else you wanted to see done, here, before it gets merged? Obviously the failing tests & conflicts need to be resolved, but I suspect once @nzachow updates the code that will be easy / resolve on its own.

nzachow:
Hi @bhilburn !

From what I recall the biggest hold up was something
related to the tests. I have the impression that they were
passing locally but Travis wasn't able to run them correctly,
and I could not find why.

After that I kinda forgot about this PR ...

I'll take a look into the `next` branch to see how easy would
be to merge this code.

onaforeignshore:
@bhilburn @nzachow The new codebase doesn't make use of segment states in the same way as the original code. You now register a complete state with its own icon and colours. This will require some rewriting to use the new way.

There is still a call to `gawk` which is definitely not available on all systems that needs to be resolved.

bhilburn:
Agreed, the external call to `gawk` needs to be replaced with pure ZSH. @nzachow - let us know if you need help with that.

@nzachow - Check out some of the other segment code in the segments/ directory in the `next` branch, it should serve as a good example of how to make things work, now, and please let us know if you have any questions. We need to update the developer docs for the new codebase, so you going through this process would actually be really helpful for us!

thche:
@bhilburn would be very interested in seeing this segment in the wild.
if you need help HMU!

PR #925, opened 2018-07-21
[Bugfix] Resolve #918 Transparent background

### LINKED ISSUE #918: Transparent background broke
#### Describe Your Issue
I was using the `none` color as described here https://github.com/bhilburn/powerlevel9k/issues/240#issuecomment-200072402 to get transparent background color but when I upgraded from 0e3c1924fe40835dfe189227b8a165de147e65a1 to afcb1b7f84466898c632f47a75d7f1cdf5334f13 it stopped working and my background is now dark grey

Is there a way to continue to have a transparent background?

I think it comes from the change `functions/colors.zsh` between these two commits.
Before (0e3c192): https://github.com/bhilburn/powerlevel9k/blob/0e3c1924fe40835dfe189227b8a165de147e65a1/functions/colors.zsh
After (afcb1b7): https://github.com/bhilburn/powerlevel9k/blob/afcb1b7f84466898c632f47a75d7f1cdf5334f13/functions/colors.zsh
https://github.com/bhilburn/powerlevel9k/compare/0e3c1924fe40835dfe189227b8a165de147e65a1...afcb1b7f84466898c632f47a75d7f1cdf5334f13

#### Environment Information
ZSH version: 5.5.1
Oh-My-ZSH: Yes
P9k install: cloned in `~/.oh-my-zsh/custom/themes/powerlevel9k`
Terminal emulator: Terminator

JulienPivard:
Before function getColorCode consider value 'none' like a bad value and search for another value that could match.
Now function getColorCode consider value 'none' like a good value and no longer seeks to replace it.
When it's use to set background the segment become transparent.

### DIFF
--- a/functions/colors.zsh
+++ b/functions/colors.zsh
@@ -79,6 +79,10 @@ function getColorCode() {
     else
       echo -n "$1"
     fi
+  # Check if value is none with any case.
+  elif [[ $1 = [nN][oO][nN][eE] ]]
+  then
+    echo -n 'none'
   else
     typeset -A codes
     # https://jonasjacek.github.io/colors/
--- a/functions/colors.zsh
+++ b/functions/colors.zsh
@@ -80,7 +80,7 @@ function getColorCode() {
       echo -n "$1"
     fi
   # Check if value is none with any case.
-  elif [[ $1 = [nN][oO][nN][eE] ]]
+  elif [[ "${(L)1}" == "none" ]]
   then
     echo -n 'none'
   else


functions/colors.zsh
```diff
@@ -79,6 +79,10 @@ function getColorCode() {
     else
       echo -n "$1"
     fi
+  # Check if value is none with any case.
+  elif [[ $1 = [nN][oO][nN][eE] ]]
```
dritter: It would be more readable to write that as `elif [[ "${(L)1}" == "none" ]]`

JulienPivard: Yes I add to the next commit

bhilburn:
@julienfalque - Thanks so much for putting together this PR! We'll be able to get this in our next release, v0.6.6, which, I'm really happy about.

@onaforeignshore - Do you mean outside of the if / elif / else structure? I'm a fan of keeping that conditional flow to prevent unneeded evaluations, although the lines couldn't be concatenated to:

```
  # Check if value is none with any case.
  elif [[ "${(L)1}" == "none" ]]; then
    echo -n 'none'
```

JulienPivard:
You're welcome
@bhilburn you missed your tag ^^

onaforeignshore:
@bhilburn Just FYI.

```
Testing: [[ "${(L)s}" == "none" ]]
( repeat 1000000; do; [[ "${(L)s}" == "none" ]]; done; )  0.84s user 0.00s system 99% cpu 0.845 total

Testing: [[ ${s} == [Nn][Oo][Nn][Ee] ]]
( repeat 1000000; do; [[ ${s} == [Nn][Oo][Nn][Ee] ]]; done; )  1.37s user 0.00s system 99% cpu 1.370 total
```

JulienPivard:
it's already changed in 5ce384f

onaforeignshore:
@JulienPivard  The (L) in `${(L)1}` means convert to lowercase, which means your test will always fail as `NONE` will never be equal to `none`. To test for `NONE` you need `${(U)1}` instead.

JulienPivard:
It's worse with : 
```shell
if [[ "${(U)1}" == "NONE" || "${(U)2}" == "NONE" ]]; then
```
result : 
test1

dritter:
I fixed a problem in combination with newlines in #960 . Could you try out that PR, and and see if the problem exists there?
And if it does, could you share your configuration with us?

JulienPivard:
@dritter It's work.
test_conf_zsh

bhilburn:

Issue #926, opened 2018-07-23
Missing `etc` state for the `dir` segment in docs


xufab:
I was wondering why my colors were wrong for the `dir` segment in `/etc/**` and I discovered there was an undocumented `etc` state.

This allow the use of `POWERLEVEL9K_DIR_ETC_BACKGROUND` and `POWERLEVEL9K_DIR_ETC_FOREGROUND` variables.

I've added the state in the docs here https://github.com/bhilburn/powerlevel9k/wiki/Stylizing-Your-Prompt but I don't know if there is other places where it's missing, that's why I'm creating this issue.

dritter:
Thanks @xufab for taking the time to fix that issue.
Indeed, the info should be added to the [README.md](https://github.com/bhilburn/powerlevel9k#dir) as well. Would you mind creating a PR for that?

xufab:
No problems

bhilburn:
Merged into master as part of #944! Will be in the v0.6.6 release.

PR #928, opened 2018-07-23
[Docs] Add etc state description in dir docs

xufab:
#### Description
Add etc icon variable in the dir segment docs

#926 

#### Questions
I didn't found a coherent image for the gear character. Whre does the three other images comes from?


bhilburn:

PR #929, opened 2018-07-23
Add missing colors to fix color comparison

### LINKED ISSUE #921: OS_ICON segment separator issue
Hi. I'm using the 'os_icon' element and have run into a problem when configuring the foreground and background colours. For example, when I set the colours (as below) I'm not getting the matching colour separator:

POWERLEVEL9K_OS_ICON_BACKGROUND='cyan'
POWERLEVEL9K_OS_ICON_FOREGROUND='black'

cyan

This is also occurring for magenta and black bg colours. All other colours are fine.

Environment:
zsh 5.4.2 (x86_64-ubuntu-linux-gnu)
oh-my-zsh
P9K from git repo
urxvt, gnome-term, tilix

UbuntuMono Nerd Font Mono
POWERLEVEL9K_MODE='nerdfont-complete'
get_icon_names output looks good

Any help would be appreciated.

P.S. I'm also having difficulty setting the fg/bg colours for the _user_ element. Any settings I make in .zshrc are being ignored. 

POWERLEVEL9K_USER_FOREGROUND='black'
POWERLEVEL9K_USER_BACKGROUND='green'

dritter:
Added magenta and cyan that were only present with their alternative
names (fuchsia and aqua). See [here](https://en.wikipedia.org/wiki/X11_color_names#Clashes_between_web_and_X11_colors_in_the_CSS_color_scheme).

This fixes #921 

@vaskark could you checkout this PR and see if that helps?
Disclaimer: 
By the help of the wikipedia article, I just found `cyan` and `magenta` to have problems. Please post here, if you found other colors that do not work as intended.

For everyone to test:
You need to set a background color for a segment that is unknown to our internal colors array. If that happens, the segment separator to the next segment shows up in black. This is how you know that this color is unknown.

### DIFF
--- a/functions/colors.zsh
+++ b/functions/colors.zsh
@@ -313,7 +313,7 @@ function foregroundColor() {
 # into ZSH-Style color codes.
 function getColorCode() {
   # Early exit: Check if given value is already numerical
-  if [[ "$1" = <-> ]]; then
+  if [[ "$1" == <-> ]]; then
     # Pad color with zeroes
     echo -n "${(l:3::0:)1}"
     return
--- a/functions/colors.zsh
+++ b/functions/colors.zsh
@@ -6,6 +6,270 @@
 # https://github.com/bhilburn/powerlevel9k
 ################################################################
 
+typeset -gAh __P9K_COLORS
+# https://jonasjacek.github.io/colors/
+# use color names by default to allow dark/light themes to adjust colors based on names
+__P9K_COLORS=(
+  black 000
+  maroon 001
+  green 002
+  olive 003
+  navy 004
+  purple 005
+  teal 006
+  silver 007
+  grey 008
+  red 009
+  lime 010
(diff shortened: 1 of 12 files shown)


dritter:
There still seems to be some problems with at least `purple3` as reported in #931 ..

dritter:
As assumed, there was another bug in the left prompt segments. If there was a left middle segment that had has a background color that is unknown to the Terminal Emulator, like `purple3`, that would lead to storing that color as `CURRENT_BG`. The next segment separator is then printed in the foreground color `purple3`, without getting the color code, so the separator gets printed in default white.

The idea behind this PR is that we now always print the segment colors in their numerical representation. The numerical values should work in every Terminal Emulator that supports 256 colors. That way we don't need to put much effort into finding out if the Terminal supports a certain color.

Also I removed the equalization of bright and normal colors.

PR #933, opened 2018-07-26
Fixes background reset for ending separator

Syphdias:
#### Title
Fixes a bug that leads to a black background for the seperator
 - [Bugfix]
 
#### Description
The default was to reset to black at the end of the seperator instead of clear
As far as I can tell commit 4d18fdc6e9252a5d93ab9e046b3d98421d1be715 introduced this bug.
Before:
image
After:
image


JulienPivard:
It's done in #925 :)

Issue #817, opened 2018-04-30
Discussion: Speed vs Flexibility


dritter:
As you all know, we are in the process of merging #771 into `next`. There we came up with the question if we should sacrifice some flexibility for speed. I'll try to sketch some cases here.
Btw. "Flexibility" is the wrong word, cause both Opportunities are flexible.. Bear with me.

To vote either choose 🎉 (Choice 1; Flexibility) or 😄 (Choice 2; Speed) as a [Reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/).

### Current State
The current state of P9K is that you could change almost every variable anytime. So, for instance, if you want to change the Background Color of your `dir` segment, just write out `POWERLEVEL9K_DIR_HOME_BACKGROUND="green"` and it will change to green as soon as you hit enter. Few other variables do not work this way. Most prominent example is here the `POWERLEVEL9K_MODE`. This cannot be changed without re-sourcing P9K (in last consequence: Shell restart).

### State now implementation in the code separation branch
A new aspect that is introduced via #771 is that the segments are separate files, so they need to be loaded separately. We do that by checking which segments are currently active, and loading only that files that are needed (this mechanism is where [the discussion](https://github.com/onaforeignshore/powerlevel9k/commit/bbf597e5658eb9431e33bb2bd4ecbffed76c7941) between Chris and Me started).

### Choice 1: Flexibility 🎉 
This is basically the current state (see above). But we should work on `POWERLEVEL9K_MODE` etc. to be able to change it anytime.
Downsides are: We need to check for everything every time. So whenever the user hits enter, we need to check for every icon override, every color override, etc.

### Choice 2: Speed 😄 
The main idea here is to do as much as possible upfront. My claim here is that we have at least 99% execution time, and only 1% of the time where the user changes his prompt. So we should optimize for the 99%.
This includes: every `set_default` should be outside of the functions, `prompt_icon` no longer checks for overrides (should be done upfront), `left/right_prompt_segment` no longer checks for color overrides, Segments are loaded only once at startup, etc.
Downsides are: Major change in behavior (I expect a lot of support tickets here). The user must restart his shell (or we implement a command `p9k_reinitialize_prompt`). Lastly the Variables must be declared _before_ the theme is sourced (the old `POWERLEVEL9K_MODE` problem).

Speaking in Code:
```zsh
function print_icon() {
  local icon_name=$1
[...]

PR #934, opened 2018-07-27
Add tests

dritter:
This PR covers three topics:

## Add tests
Obviously, this is the main purpose of this PR. The Tests are mainly extracted from #344.

## Improve Test Structure
I improved the Tests stylistically and prepared them for the upcoming performance optimisations. This means two things:
1. I made use of local variables as much as possible, instead of defining and unsetting variables all the time. This is more readable IMHO.
2. Define the variables first, then load the Theme (in preparation for #817 )

## Test on many ZSH versions
The Tests are executed by Travis on ZSH 5.5.1, 5.5, 5.4.2, 5.4.1, 5.3.1, 5.3, 5.2, 5.1.1 both on Linux and MacOS.

## Caveats
There are a few Changes in `powerlevel9k.zsh-theme`:
- I changed some grep RegExes to runnable versions on Linux and MacOS (mainly changed `grep -o "[0-9.]*` to `grep -o -E "[0-9.]+`).
- Custom Segments can now have a visual identifier
- Fixed foreground color of `vi_mode` segment (changed from `default` to `white`)
- Changed `detect_virt` segment to print nothing if nothing was detected.

This PR aims on `master`, but if there are too many changes in behaviour, we could change base to `next`..


bhilburn:

PR #935, opened 2018-07-28
[Enhancement] New background jobs

onaforeignshore:
#### New background jobs segment
This update gets rid of the external calls. In addition, it can now show Running jobs as well as Suspended jobs.
 
#### Description
If `POWERLEVEL9K_BACKGROUND_JOBS_VERBOSE` is set, the segment will display if the job count is more than 0.

If `POWERLEVEL9K_BACKGROUND_JOBS_VERBOSE_ALWAYS` is set, it will always show the segment, with `⚙ 0` - if there are none. (Note that this will _always_ override `POWERLEVEL9K_BACKGROUND_JOBS_VERBOSE` to be true).

If `POWERLEVEL9K_BACKGROUND_JOBS_EXPANDED` is set, the new format will display
`⚙ 0r 1s` - indicating (r)unning and (s)uspended jobs,
otherwise it will display just the total of the two,
`⚙ 1` - indicating the total.

#### To Do
I still have to update the tests if you like the update.

Any comments/suggestions?


bhilburn:
@onaforeignshore - Thanks so much for filing this! A few thoughts:

a) I *really* like the removal of calling external programs. Nice work!
b) Since this is a new feature and we're already drowning in stuff for v0.6.6, I'd like to instead put this in our first release of v0.7.0
c) I see you're triggering the update the job counts in `precmd`. This is fine when that segment is loaded, but it would be nice to not do that if the user isn't actually using the segment. Given the segment-level loading of the v0.7.0 codebase, can we conditionalize the `precmd`s based on loaded segments to avoid the overhead of those that are unused?

onaforeignshore:
@bhilburn c) Already done in `070-registerSegment` - you can register `precmd` in the code for the segment.

bhilburn:
@onaforeignshore - Perfect =)

onaforeignshore:
Replaced by #999

PR #937, opened 2018-07-31
Use SUDO_COMMAND to check for sudo

### LINKED ISSUE #852: About `sudo` detection
I just encountered an error message while using `next`

```
May 31 11:52:26 mymachine sudo[10180]:      tim : a password is required ; TTY=pts/2 ; PWD=/home/tim ; USER=root ; COMMAND=/usr/bin/true
```
## System Information:
OS: Fedora 28
ZSH:  5.5.1
Used segments: `os_icon`  `context`  `dir`  `status`  `time`

## Steps to reproduce:
1. Start a shell running `journalctl -f` as root to monitor the error message
2. Start a new non-root zsh with powerlevel9k on `next` branch
3. Just enter a `cd` or `ls` (any command will do)
4. The error message appears on `journalctl`

docwhat:
`sudo -n true` only checks that we _could_ use `sudo`, not if we are in
a sudo session.

closes #852

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -623,12 +623,12 @@ prompt_context() {
   if [[ $(print -P "%#") == '#' ]]; then
     current_state="ROOT"
   elif [[ -n "$SSH_CLIENT" || -n "$SSH_TTY" ]]; then
-    if sudo -n true 2>/dev/null; then
+    if [[ -n "$SUDO_COMMAND" ]]; then
       current_state="REMOTE_SUDO"
     else
       current_state="REMOTE"
     fi
-  elif sudo -n true 2>/dev/null; then
+  elif [[ -n "$SUDO_COMMAND" ]]; then
     current_state="SUDO"
   fi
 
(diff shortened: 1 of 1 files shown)


dritter:
Thanks @docwhat for that PR.

As @onaforeignshore [pointed out](https://github.com/bhilburn/powerlevel9k/pull/944#issuecomment-411690962), the `prompt_user` segment has a [`sudo` state](https://github.com/bhilburn/powerlevel9k/blob/571a859413866897cf962396f02f65a288f677ac/powerlevel9k.zsh-theme#L654) as well. Could you change that as well?

docwhat:
@dritter Done.

dritter:
Awesome! Will merge that later in #944 today.

bhilburn:
@docwhat - Thanks so much for putting this together. I really do appreciate it, and it'll be included in the v0.6.6 release that we will get out the door ASAP.

@dritter - nice job catching that comment from @onaforeignshore in the other PR! That would have been easy to miss, and thanks for staging this for v0.6.6 release `=)`

bhilburn:

tippl:
This might be kicking a dead thread, sorry about that, but it seems the SUDO state was misunderstood a bit.

Original idea of the state was to have a yellow prompt if the sudo has cached user creds, as such "danger, sudo in this terminal will not ask password" thus the "can i run sudo" implementation.

$SUDO_COMMAND is set by sudo for processes started by sudo, thus does not reflect the functionality.

For it to be set, zsh would have to run in a sudo session, however that means the process has root permissions, so the ROOT state is used instead here, making the SUDO state not used at all.

I do not know if it is possible to implement this feature without incurring the "password is required" line in log, so i am not against removing the feature entirely, just mentioning this here so we don't have dead code that does nothing.

PR #941, opened 2018-08-02
Bugfix: Don't show remote branch if ambiguous due to tag

towo:
If there is a tag with the same name as the remote branch, `git symbolic-ref --short HEAD` will display `heads/$branchname` instead of a simple `$branchname`.

This patch will drop the `heads/` from the tracking branch check.


dritter:
Hi @towo !

Sorry for the delay. I just merged this PR into our development branch `next` and added a short test.

Thank you very much & welcome to P9K ;)

bhilburn:
Thanks so much for the PR2 @towo! This is a really great first contribution! `=)`

Thanks for merging, @dritter !

Issue #936, opened 2018-07-30
Add fallback icon for missing linux distro icons


eloo:
#### Describe Your Issue
I've just updated to the latest version of the theme and now my OS_ICON is broken.
I'm using a Debian and now the icon is only "Deb" instead of the Tux icon.
I guess it would be great that at least for fonts which are containing the Tux icon this Icon will be used instead of a simple string.
What do you think?

#### Environment Information
This information will help us understand your configuration. 

  - Debian
  - Font: source code pro + powerline awesome
  - Font-Mode: awesome-patched

dritter:
Hi @eloo !

Sorry for the late response. I agree that, as we add more fine-grained control over the Linux icons in Nerd-Fonts, it feels like a bug for other fonts to fall back to a textual representation. That is probably why we get more and more bug reports like this..

Long story short: Would you mind creating a PR for that?

eloo:
yes i can.
i hope i will find time on the weekend for that

bhilburn:
Merged into master as part of #944! Will be in the v0.6.6 release.

Issue #927, opened 2018-07-23
Colors between 008-015 are not working as I expected


Slabity:
#### Describe Your Issue
Colors from the `008`-`015` range are not working properly in my terminal.

I expect this: https://i.imgur.com/6uUBOgX.png

But I get this: https://i.imgur.com/tjzI9IE.png

#### Have you tried to debug or fix it?

Yes. It appears the issue originates from 4d18fdc6. In `functions/colors.zsh`, there are two points where we subtract 8 from the actual color given. Removing these seems to fix my color issues.

In `function getColor()`:
```
  if [[ "$1" = <-> ]]; then
    if [[ "$1" = <8-15> ]]; then
      1=$(($1 - 8))
    fi
  else
  ...
```

In `function getColorCode()`:
```
  if [[ "$1" = <-> ]]; then
    # ANSI color codes distinguish between "foreground"
    # and "background" colors. We don't need to do that,
    # as ZSH uses a 256 color space anyway.
    if [[ "$1" = <8-15> ]]; then
      echo -n $(($1 - 8))
    else
      echo -n "$1"
    fi
  else
  ...
```

#### Environment Information
Not sure how much of this matters:

  - ZSH version 5.5.1
  - Pk9 master branch
  - Termite emulator
  - NixOS (x86_64-unknown-linux-gnu)

Slabity:
This seems quite intentional, but I do not really understand the reason behind mapping colors `008-015` to `000-007`

dritter:
Hi @Slabity !

Sorry for the late response. I wrote the code you mentioned. This code is quite old (about 3 years). Commit: 0e37d8ef865daddd6af02ba78b4a3bb86fd5917e. TBH, I am not sure why I did it like this. Probably because the bright colors did not work on my machine, or at least I saw no difference and I wanted to treat these colors as if they were the non bright version..

I think the code could be removed safely.

dritter:
Btw. I removed the code in #929 , which is merged in #944 . Could you give it a try and confirm that this works for you?

Slabity:
Looks like it worked. Thanks!

bhilburn:
Merged into master as part of #944! Will be in the v0.6.6 release.

PR #944, opened 2018-08-05
Prepare 066

### LINKED ISSUE #918: Transparent background broke

dritter:
This is a preparation for `v0.6.6`.

The following PRs are incorporated in here:
- #934 - Add Tests
- #884 - test-in-docker: fix with newer ZSH versions
- #928 - [Docs] Add etc state description in dir docs
- #937 - Use SUDO_COMMAND to check for sudo
- #925 - [Bugfix] Resolve #918 Transparent background
- #923 - Fix font issue debugging script
- #921 - Add missing colors to fix color comparison
- #951 - Add fallback icon for missing linux distro icons
- #956 - Fix broken link in readme

### Bugs fixed
- #927 
- #852 
- #926 
- #936 

### What's not included in this PR
- Still no "real" fix for #825


dritter:
It would be great, if somebody could test this PR and confirm that it works. :)

onaforeignshore:
#937 was applied to context but not to user segments

powerlevel9k.zsh-theme
```diff
@@ -189,7 +189,7 @@ left_prompt_segment() {
 # End the left prompt, closes the final segment.
 left_prompt_end() {
   if [[ -n $CURRENT_BG ]]; then
-    echo -n "%k%F{$CURRENT_BG}$(print_icon 'LEFT_SEGMENT_SEPARATOR')"
+    echo -n "%k$(foregroundColor ${CURRENT_BG})$(print_icon 'LEFT_SEGMENT_SEPARATOR')"
```
onaforeignshore: If all color codes are now numeric, why do we need to add this extra function call?

onaforeignshore: `$CURRENT_BG` will contain a numeric, since `getColorCode` was already called in `left_prompt_segment`. Do we really have to add another function call to slow the prompt down more?

dritter: Yes, `$CURRENT_BG` is already numeric. I made this change anyway, to match up with the other occurrences of foreground color. To go full nuts I removed the possibility to stop an fore/background color with a call to that method without params.

In an upcoming release we should either go all in with helper methods for getting a prepared string that contains "start a foreground color", or nothing. I am undecided between the two ways. In another programming language I would prefer a helper method that wraps a cryptic DSL to a nice interface, but here I would indeed prefer going for speed and not call a subshell for that. We should think about a concept for that.

Slabity:
Just tested this on my system. Colors seem to be working just fine.

However, I did notice something else. Despite setting `POWERLEVEL9K_HOME_ICON='\uf015'` to change my home folder icon, the tilde in the directory still exists. Same with `POWERLEVEL9K_FOLDER_ICON`. Is that expected?

dritter:
Thx for testing @Slabity .
Yep that is intentional. You can set `POWERLEVEL9K_DIR_OMIT_FIRST_CHARACTER=true`, to get rid of the first character.

PR #950, opened 2018-08-10
Clarify none background

dritter:
This PR is based on #944 and the purpose is to clarify the initial value of `CURRENT_BG` and `CURRENT_RIGHT_BG`. We now use `initial` as - well - initial value of these variables. This is why we can get rid of the `NONE` guards in `isSameColor` that bailed out if a color `NONE` was given. Unfortunately our numerical comparison that helps us comparing values like `009` and `9` fails in comparing a numerical value and a string (`none`). This is why I added a check if one of that given colors is a numerical one, and if so, fall back to a string comparison.

Not sure if this should be included in `v0.6.6`, or if we should include this in `next`.


bhilburn:
I'm glad you're trying to tackle this, @dritter. The recent issues about color handling remind me a little of some of our early days dealing with fonts, hah.

The `none` thing not being a numeric code is really annoying. I really don't like checking to see if both colors are numerics before doing the comparison, because most of the time, they will be numerics, so we are injecting two more conditionals __per segment__ into the rendering process. (Just to be clear, this PR doesn't add that - that's been around since we first added support for `none`).

Not withstanding finding a way to blow away this whole thing entirely, which I don't see an obvious way to do, this seems like a good way to replace the existing conditional pair with two different ones that solve the problem. I'm on-board. If you guys are, too, let's get this staged for v0.6.6.

dritter:
Hmm. Probably we need that string comparison anyway. Imagine the following cases:
1. `isSameColor 'xxx' 'black'`: This one should return false, as `xxx` is not a color in our colors array.
2. `isSameColor 'xxA' 'xxB'`: This one should return false as well, because two unknown different colors should not be recognised as equal. So, if we compute an "error" integer for them (in `getColorCode`), they will end up being the same.

Issue #955, opened 2018-08-16
right prompt color error


DuncanHuang:
Hi, I'm setting "POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS" using "ip", but color error

![2018-08-16_10-40-27](https://user-images.githubusercontent.com/1020091/44184581-4c9bea00-a142-11e8-85c6-175ec9a1e699.png)

I find is 264897a426c7286d0734895b35ea437a6accc1bc this commit after, i checkout to 87acc51acab3ed4fd33cda2386abed6f98c80720 this problem does not exist.

![2018-08-16_10-41-02](https://user-images.githubusercontent.com/1020091/44185119-085e1900-a145-11e8-99d4-981ce993c618.png)

I try modify file "functions/colors.zsh", change new version

```text
'black'   '000'
'red'     '001'
'green'   '002'
'yellow'  '003'
'blue'    '004'
'magenta' '005'
'cyan'    '006'
'white'   '007'
```
to old version

```text
'black'  '000'
'maroon' '001'
'green'  '002'
'olive'  '003'
'navy'   '004'
'purple' '005'
'teal'   '006'
'silver' '007'
```
this problem does not exist, i think this may be the color mapping error.

This problem in macOS does not exist, but in CentOS 6、7 is exist.

Provide my setting parameters

```text
ZSH_THEME="powerlevel9k/powerlevel9k"
POWERLEVEL9K_CONTEXT_TEMPLATE="%n"
POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(context dir vcs)
POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(ip root_indicator background_jobs command_execution_time time)
POWERLEVEL9K_SHORTEN_DIR_LENGTH=3
POWERLEVEL9K_SHORTEN_STRATEGY=truncate_from_right
POWERLEVEL9K_PROMPT_ON_NEWLINE=true
POWERLEVEL9K_PROMPT_ADD_NEWLINE=true
POWERLEVEL9K_MULTILINE_FIRST_PROMPT_PREFIX='╭'
POWERLEVEL9K_MULTILINE_LAST_PROMPT_PREFIX='╰'$'\uE0B0 '
POWERLEVEL9K_EXECUTION_TIME_ICON=''
POWERLEVEL9K_COMMAND_EXECUTION_TIME_THRESHOLD=0
[...]

PR #959, opened 2018-08-17
[Discussion] Change bright colors to old version

dritter:
This should fix the second issue in #955 that @DuncanHuang brought up.

The colors were changed in `v0.6.5`. This PR updates the colors to the previous version.

So the colors `red`, `green`, `yellow`, `blue`, `magenta`, `cyan` and `white` point the bright colors:
```
'black'   '000'
'red'     '001'
'green'   '002'
'yellow'  '003'
'blue'    '004'
'magenta' '005'
'cyan'    '006'
'white'   '007'
```
I did a separate PR for this, because this needs a bit of discussion. The [page](https://jonasjacek.github.io/colors/) that was the source for the color names lists the colors mentioned above in the range of 8-15, like they are in `v.0.6.5`. But that means that users who upgrade may have to change their color definitions. That is why we need to discuss this here. The commit to look at is 8e966e3636ce7c7162b60763088252b7111985e5

In short:
- Option A
Change colors like in this PR. This means that users who upgrade P9K from an old Version (pre `v0.6.5`) get the _same_ colors as before, but if they upgrade from `v0.6.5`, colors change. Second downside is that we turn away from other color definitions.

- Option B
Leave colors as they are. This means that users who upgrade P9K from an old Version (pre `v0.6.5`) get _different_ colors.

### DIFF
--- a/functions/colors.zsh
+++ b/functions/colors.zsh
@@ -11,23 +11,23 @@ typeset -gAh __P9K_COLORS
 # use color names by default to allow dark/light themes to adjust colors based on names
 __P9K_COLORS=(
   black 000
-  maroon 001
+  red 001
   green 002
-  olive 003
-  navy 004
-  purple 005
-  teal 006
-  silver 007
+  yellow 003
+  blue 004
+  magenta 005
+  cyan 006
+  white 007
   grey 008
-  red 009
+  maroon 009
   lime 010
-  yellow 011
-  blue 012
+  olive 011
+  navy 012
   fuchsia 013
-  magenta 013
+  purple 013
   aqua 014
-  cyan 014
-  white 015
+  teal 014
+  silver 015
   grey0 016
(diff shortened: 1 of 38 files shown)


AuroraWright:
The new color scheme looks terrible on Solarized Dark (on the macOS Terminal.app at least). Colors are muted and dark-ish, and I basically need to customize most blocks (including the "dir" one) for them to stand out.

bhilburn:
@dritter - Thanks so much for putting this together, explaining the situation, and laying out the options. I would like to revert back to the previous colors; breaking existing configs is never a good plan, IMO.

It looks like this PR was filed before I had actually merged the v066 staging branch into master, so there is a bit of git logistics to go through, here. I'll get that sorted and then get this branch merged. =)

@AuroraWright - Thanks for your input here, by the way!

PR #960, opened 2018-08-17
Fix color comparisons with newlines

dritter:
It seems like the color comparison breaks if `getColorCode` returns lowercased `none`.

Before:

The second line outputs a white segment separator, which shouldn't be there.
Relevant trace is:
```
λ1 left_prompt_segment:27 cmdand 	foreground='%F{none}'
λ1 left_prompt_segment:29 if 	[[ 012 != NONE ]]
λ1 left_prompt_segment:29 if 	isSameColor none 012
λ1 isSameColor:1 if 	[[ none == NONE || 012 == NONE ]]
λ1 isSameColor:5 cmdsubst 	getColorCode none
λ1 getColorCode:2 if 	[[ none == <-> ]]
λ1 getColorCode:8 	local colorName=none
λ1 getColorCode:10 if 	[[ none == none ]]
λ1 getColorCode:11 then 	echo -n none
λ1 isSameColor:5 	local color1=none
λ1 isSameColor:6 cmdsubst 	getColorCode 012
λ1 getColorCode:2 if 	[[ 012 == <-> ]]
λ1 getColorCode:4 then 	echo -n 012
λ1 getColorCode:5 then 	return
λ1 isSameColor:6 	local color2=012
λ1 isSameColor:8 	return 1
λ1 left_prompt_segment:30 then 	echo -n '%K{none}%F{012}'
```
Line 29 of `left_prompt_segment` is:
`if [[ $CURRENT_BG != 'NONE' ]] && ! isSameColor "${backgroundColor}" "$CURRENT_BG"; then`
So, the check for `isSameColor none 012` returns false:
```
λ1 left_prompt_segment:29 if 	[[ 012 != NONE ]]
λ1 left_prompt_segment:29 if 	isSameColor none 012
```

After:

(The last segment separator is not of matter here)


bhilburn:

PR #966, opened 2018-08-22
[Bugfix] Fix icons cut off in RPROMPT segments

ytang:
This PR is a follow-up to #854, thanks to the comment by @IevgenSobko. Specifically, visual identifiers in RPROMPT were still cut off because of the color reset. The fix is even simpler than the previous PR: just remove the `%f` after the visual identifier.

Here is a screenshot showing the cut-off before (the first line) and after (the second line) applying the patch:
powerlevel9k-fix-rprompt-icons-cut-off


LordTyrius:
Thanks for yet another fix! This is important to me, as adding a space after the icon does no longer seem to be a workaround in konsole as suggested by the readme. It seems like the space now has the same background color as the segment, and still overdraws the right part of the glyph.

Am I understanding correctly that this PR is scheduled to land in 0.7.0?

bhilburn:
@ytang - Thanks so much for this PR! Sorry it's taken us a few weeks to circle around to it.

There are now conflicts with the merge due to the change in the color naming that was merged in a different PR. They look pretty straight-forward to resolve, though.

@dritter - I see you tagged this one for v0.7.0, and we for sure need to port it over to the new codebase. I think this one is probably also worth merging into `master`, here, as soon as we can since it's a useful bugfix. Thoughts?

dritter:
I agree, this is also worth merging into `master`.
And as @LordTyrius pointed out, this fix is only for older Konsole versions. For newer Konsole versions this won't work, and this needs to be fixed on Konsole's side AFAIR.

ytang:
FYI, `test/segments/context.spec` also fails in master and should be unrelated to my patch.

dritter:
@ytang I created #1020 to fix the issues with `context.spec`.

PR #973, opened 2018-09-03
Update README.md for EXECUTION_TIME icon

bspeice:
Note how to change the "Dur" icon for execution time

#### Update README.md to note EXECUTION_TIME icon change
 - [Docs]
 
#### Description
Super easy note, but I think the default "Dur" looks strange, and so want to make sure people know how easy it is to change!

#### Questions
Anything I need to update on a Wiki? Should this be generalized to include notes on what the icon names are and how you can override any of them?


bspeice:
Found a reference on the docs about how to change icons, while I'd love to see that note made more prominent, I don't think this is the right change.

Issue #431, opened 2017-03-10
Truncate VCS Branch


rawkode:
Sometimes branch names can get a little out of control and take up the entire terminal (Especially if you're using BitBucket/GitLab/GitHub and the branch names can be inferred from a task)

Would it be possible to allow this to be truncated, like directories?

dritter:
Awesome idea! 👍  
I want to refactor the dir segment anyway. And this is a perfect addition. I'll do a PR for that in the next weeks..

bhilburn:
I agree, this is a really great idea, @rawkode 

@dritter - We could perhaps include some of the ideas from @ikelner's commit in [his fork on Gitlab](https://gitlab.com/lokiexinferis/powerlevel9k/commit/935a0942cae55bb2c092903739c9296754af2dde) into the `dir` segment refactor.

bhilburn:
@dritter - Are you still hacking on this one?

dritter:
Yep, I am still on it. But refactoring the `dir` segment is harder than I thought. I'll turn to truncating VCS branches after dir segment is refactored. You can track my progress [here](https://github.com/dritter/powerlevel9k/tree/rework_dir_truncation). ;)

bhilburn:
How are things going on this one?

I just don't want things to bitrot if you have useful work done =)

dritter:
This is bigger than I expected.. 😲  But I have a working draft. Still needs some polish. And maybe a rethinking of the truncation strategy. I changed it to accept an array of methods, but I think this is still not sufficient. One would want to execute some truncations more than once with different settings. E.g. first split a string by X and then by Y..

bhilburn:
Leaving this one open until @dritter says otherwise `=)`

ionsquare:
Any update on this? It would be really useful to have.

dritter:
This is already in `master` now (See #798 ). Thanks to @floor114 .

PR #979, opened 2018-09-06
[Enhancement] Adds truncate_to_unique_but_last as POWERLEVEL9K_SHORTEN_STRATEGY

Syphdias:
I thought it would be nice to see the current directory in its full length and have the rest of the path truncated. I think it's a good compromise between space efficiency and readability/orientation.
I'd say it's a mix between `truncate_to_unique` and `truncate_from_right`
image

A nice addition would be to use POWERLEVEL9K_SHORTEN_DIR_LENGTH to determine how many directories should not be truncated from the right. Maybe as $2 parameter to getUniqueFolder and poping that many items from the paths array.

If you like `truncate_to_unique_but_last` I could try to implement the DIR_LENGTH into `truncate_to_unique` instead of inventing a new "strategy".

Please let me know what you think!


powerlevel9k.zsh-theme
```diff
@@ -858,6 +858,16 @@ prompt_dir() {
           [[ $current_path == "~"* ]] && current_path="~${trunc_path//${home_path}/}" || current_path="/${trunc_path}"
         fi
       ;;
+      truncate_to_unique_but_last)
+        # see truncate_to_unique
+        if (( ${#current_path} > 1 )); then # root and home are exceptions and won't have paths
+          # cheating here to retain ~ as home folder
+          local home_path="$(getUniqueFolder $HOME)"
+          trunc_path="$(getUniqueFolder $(dirname $PWD))"
```
onaforeignshore: Use `${PWD:h}` instead of `$(dirname $PWD)` - it's internal and thus a lot faster 😉

powerlevel9k.zsh-theme
```diff
@@ -858,6 +858,16 @@ prompt_dir() {
...
+          local cwd="$(basename $PWD)"
```
onaforeignshore: Use `${PWD:t}` instead of `$(basename $PWD)`

powerlevel9k.zsh-theme
Syphdias: Ah, good to know, thanks.
Using `:h` I found a way to use substitution to get the left path so I can split it up more easily. (`:t` would not work in this case as I understand it)

This also does not break if there are more `:h`s than directory levels.
```
% unique_part="${PWD:h:h}"; echo $unique_part ${PWD#${unique_part}}
/home/chkugler/git/private/confrc /submodules/powerlevel9k
% unique_part="${PWD:h:h:h:h:h:h:h:h:h:h:h:h:h:h:h:h:h:h}"; echo $unique_part ${PWD#${unique_part}} 
/ home/chkugler/git/private/confrc/submodules/powerlevel9k
```

I am having trouble getting a variable to control how many directories are cut off though
```
% echo $PWD
/home/user/git/private/confrc/submodules/powerlevel9k
% LEN=3
% # 3 for PWD, 2 characters in ":h", $LEN for setable variable 
% local_pwd=${(r#$((3+2*${LEN}))##:h#)p}                       
% echo $local_pwd                  
PWD:h:h:h
% echo ${PWD:h:h:h}                
/home/user/git/private
% echo ${(P)local_pwd}
/home/user/git/private/confrc/submodules/powerlevel9k
```
I suspect `(P)` somehow ignores my `:h`s. Any idea on that?


On a side note: the version in this pull request breaks stuff if your in a root folder, e.g. `/etc`, `/home`, `/tmp`. I'll try to fix this and add it to the tests if I can.

onaforeignshore: From [ZSH Expansion](http://zsh.sourceforge.net/Doc/Release/Expansion.html)

`21. Re-evaluation
Any ‘(e)’ flag is applied to the value, forcing it to be re-examined for new parameter substitutions, but also for command and arithmetic substitutions.`

So try `echo ${(e)local_pwd}` instead

Syphdias:
Hi @onaforeignshore, looks like you've done some restructuring. I guess merging to next isn't that easy anymore. Should I just merge next my branch and migrate my changes to the right place?

bhilburn:
Hey @Syphdias - We are in the middle of a big change! Check out this wiki page I just wrote to explain it:
[Note to Contributors](https://github.com/bhilburn/powerlevel9k/wiki/*Note-to-Contributors)

@Syphdias - Anyway, yes, you'll want to base your changes on the new `next` branch. @dritter @onaforeignshore  - does the recent PR merge make the codebase ready to base new feature braches on, or should @Syphdias wait until another few come through?

onaforeignshore:
@bhilburn There are more changes coming through
Stage 1 - split code
Stage 2 - apply Coding Guideline
Stage 3 - implement autoload functions
Stage 4 - implement `P9K_` variable name changes
Stage 5 - implement `p9k::register_icon` and `p9k::register_segment`
Stage 6 - wrap up

Syphdias:
@onaforeignshore Sounds like the code I edited will mostly stay the same besides from the location and some variable name changes. I hope I will get around to changing the structure some time today. Maybe I just copy the changes and reset my branch to `next`. The commit history of this PR will be gone, but I can live with that.
Then you can take a look an decide if you want to wait with merging.

segments/dir.p9k
```diff
@@ -96,10 +115,14 @@ prompt_dir() {
...
         if (( ${#current_path} > 1 )); then # root and home are exceptions and won't have paths
-          # cheating here to retain ~ as home folder
-          local home_path="$(getUniqueFolder $HOME)"
-          trunc_path="$(getUniqueFolder $PWD)"
-          [[ $current_path == "~"* ]] && current_path="~${trunc_path//${home_path}/}" || current_path="/${trunc_path}"
+          local cwd="current_path"
+          # 12 for "current_path", 2 characters in ":h", $LEN for setable variable
+          local cwd="${(r#$((12+2*${POWERLEVEL9K_SHORTEN_DIR_LENGTH}))##:h#)cwd}"
+          # split paths and shorten left one (beware the /)
+          local left_path="${${(P)cwd}}"
```
Syphdias: I think [changes to `(P)`](https://github.com/zsh-users/zsh/blob/master/ChangeLog#L6715) since zsh 5.2 make this possible. I could implement a workaround with `if [[ "$(zsh --version)" == *"5.1.1"* ]] ...`, a for-loop and `dirame`...
This is what I started with. It would replace these lines in function:
```
local cwd="current_path"
local cwd="${(r#$((12+2*${POWERLEVEL9K_SHORTEN_DIR_LENGTH}))##:h#)cwd}"
local left_path="${${(P)cwd}}"
```

Syphdias:
Hi,

I currently have 3 topics I'd like to have some feedback/advice on:
1. I found a clever way to not use a loop to get what I wanted to achieve. Unfortunately it is quite complicated and does not work with version 5.1. That's why I used a workaround with a loop for that, which is also easier to read. I don't know about the performance side of the two solutions. Maybe the version check slows it down even if the workaround is not used. What do you think about abandoning the "clever way" in favor of the easier to read, more compatible version. Again, I don't know which one performs better. https://github.com/bhilburn/powerlevel9k/pull/979/files#diff-bfaa9e0694796fdef87d21d0f56de667R92
2. If `P9K_DIR_SHORTEN_LENGTH` is set to 0 the shortening works as before. If it is unset, it does not shorten the path. I tried to default the value to 0 by using `${P9K_DIR_SHORTEN_LENGTH:-0}`, `${P9K_DIR_SHORTEN_LENGTH:=0}` and `p9k::set_default P9K_DIR_SHORTEN_LENGTH 0` inside the case of the function. Sadly it only works if `p9k::set_default` is used outside the function, which breaks other tests. Any thoughts or ideas?
3. @onaforeignshore, you changed the way `truncate_to_unique` worked with https://github.com/bhilburn/powerlevel9k/blame/next/segments/dir.p9k#L90. Was this on purpose? You only changed this strategy to resolve symlinks but no other? I just realized this now, and this is also the reason why you had to change the tests to fit OS X (`/private/tmp`). That's why I currently changed it back, not realizing that this was the case. If it stays that way I could probably get rid of some testing complexity.

bhilburn:
1. Yikes, I really don't like conditionalizing on the ZSH version, and I would bet calling out to the ZSH process to check the version loses much of the processing gains you would have otherwise had, anyway. Let's go with the simpler version =)
2. Need @onaforeignshore to comment here
3. and here!

onaforeignshore:
3. Yes, it was on purpose, and yes it is going to stay this way. This was done to support users on macOS and other OSes that don't have a `/tmp` folder

Syphdias:
> 2. If `P9K_DIR_SHORTEN_LENGTH` is set to 0 the shortening works as before. If it is unset, it does not shorten the path. I tried to default the value to 0 by using `${P9K_DIR_SHORTEN_LENGTH:-0}`, `${P9K_DIR_SHORTEN_LENGTH:=0}` and `p9k::set_default P9K_DIR_SHORTEN_LENGTH 0` inside the case of the function. Sadly it only works if `p9k::set_default` is used outside the function, which breaks other tests. Any thoughts or ideas?

Turns out I was wrong. Same behavior on `master`. So my changes did not break anthing ;) Though I'd prefer it to be different:
```
 ~/.oh-my-zsh/custom/themes/powerlevel9k % git status -sb
## master...origin/master
 ~/.oh-my-zsh/custom/themes/powerlevel9k % echo $POWERLEVEL9K_SHORTEN_DIR_LENGTH

 ~/.oh-my-zsh/custom/themes/powerlevel9k % POWERLEVEL9K_SHORTEN_STRATEGY="truncate_to_unique"
 ~/.oh-my-zsh/custom/themes/powerlevel9k % # no change until i set POWERLEVEL9K_SHORTEN_DIR_LENGTH
 ~/.oh-my-zsh/custom/themes/powerlevel9k % POWERLEVEL9K_SHORTEN_DIR_LENGTH="does not matter only has to be set"
 ~/.o/cu/t/p
```

> 3. Yes, it was on purpose, and yes it is going to stay this way. This was done to support users on macOS and other OSes that don't have a `/tmp` folder

I think you misunderstood. I'll try to illustrate why I think it should be changed back.  
On master:
```
 ~/.oh-my-zsh/custom/themes/powerlevel9k % git status -sb
## master...origin/master
 ~/.oh-my-zsh/custom/themes/powerlevel9k % cd ~/test/test2
 ~/test/test2 % pwd # the path displayed by dir segment with ignores symlinks
/home/chkugler/test/test2
 ~/test/test2 % /bin/pwd # the path behind the symlinks (resloved)
/home/chkugler/test
 ~/test/test2 % # three strategies for reference
 ~/test/test2 % POWERLEVEL9K_SHORTEN_DIR_LENGTH=5
 ~/test/test2 % POWERLEVEL9K_SHORTEN_STRATEGY=truncate_absolute_chars
 …test2  POWERLEVEL9K_SHORTEN_STRATEGY=truncate_middle
 ~/test/test2 % POWERLEVEL9K_SHORTEN_STRATEGY=truncate_to_unique
 ~/te/t
```

But on `next` with the `${PWD:A}`:
```
 ~/.oh-my-zsh/custom/themes/powerlevel9k % git status -sb
## next...origin/next
 ~/.oh-my-zsh/custom/themes/powerlevel9k % cd ~/test/test2
 ~/test/test2 % pwd # the path displayed by dir segment with ignores symlinks
/home/chkugler/test/test2
 ~/test/test2 % /bin/pwd # the path behind the symlinks (resloved)
/home/chkugler/test
 ~/test/test2 % P9K_DIR_SHORTEN_LENGTH=5
 ~/test/test2 % P9K_DIR_SHORTEN_STRATEGY=truncate_absolute_chars
 …test2 % P9K_DIR_SHORTEN_STRATEGY=truncate_middle
 ~/test/test2 % # until here every strategy used (and also every other strategy) thinks we're in ~/test/test2
 ~/test/test2 % P9K_DIR_SHORTEN_STRATEGY=truncate_to_unique
 ~/te % # only with this strategy it resoves symlinks
```

My point is, why would only one Strategy have special behavior like this? And it has nothing to do with OS X, since all strategies would suffer the same problems with the `/tmp` dir in tests.

On the other hand it could be an idea to implement this behavior on purpose on a optional bases for all strategies but I'd put it [further up here](https://github.com/bhilburn/powerlevel9k/blob/next/segments/dir.p9k#L41) probably and have it default to `false` but this is a topic of another thread/PR

bhilburn:
@Syphdias - Great job highlighting the difference in behavior between `master` and `next`, here - you're right, that is different. I also think I agree with you regarding potential confusion around a single strategy - I think unified approaches are much easier to understand, and especially configure.

So, the fundamental change between `master` and `next` is that symlinks are resolved with that one strategy, but not the others. I do think we ought to treat these all the same, and if users want symlinks resolved, that can be a separate option.

@dritter @onaforeignshore - Thoughts?

functions/autoload/__p9k_get_unique_path
```diff
@@ -16,24 +16,46 @@ setopt localoptions # extendedglob
...
 ##
 # @args
-#   $1 string The directory path to be truncated.
+#   $1 string Absolute directory path to be truncated.
+##
+# @info
+#   Takes paths like /abs/dir/path or ~/abs/home/dir/path but
+#   not a relative path like ./dir/path or dir/path
+#   The path "." will be ignores.
+#   Does not work well with characters that have to be escaped, e.g. " ", "*"
```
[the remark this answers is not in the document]
Syphdias: The shortening works, but if you want to expand it by pressing `tab` it won't expand.
```zsh
 ❱ __p9k_get_unique_path '~/bla foo/ca cer/d*rwin'
~/b/ca /d*
 ❱ cd ~/b/ca /d* # expands falsely to ~/b/ca /dev
 ❱ cd ~/b/ca\ /d\* # expand correctly to ~/bla\ foo/ca\ cer/d\*rwin/
```

functions/autoload/__p9k_get_unique_path
```diff
@@ -16,24 +16,46 @@ setopt localoptions # extendedglob
...
+  trunc_path='/'
+fi
+
 for directory in ${paths[@]}; do
+  # finds out how many characters you need for the $directory to be unique
   test_dir=''
   for (( i=0; i < ${#directory}; i++ )); do
     test_dir+="${directory:$i:1}"
-    matching=("$cur_path"/"$test_dir"*/)
+    matching=("$cur_path/$test_dir"*/)
```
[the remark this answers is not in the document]
dritter: Yes. Currently, you try to find a unique name by building up a glob that is every time a character longer, as long you did not find exactly one match. This is a hit to the disk every time.
My hypothesis is now that you have all information necessary after the first glob to build up a unique directory.

Example:
In a folder structure like this:
```
.
├── asdf
├── asdf100
├── asdf101
├── asdf102
├── asdf2
├── asdf3
├── asdf4
├── asdf5
├── asdf6
├── asdf7
│   └── a
│       └── b
│           └── c
├── asdf8
├── asdf89
└── asdf99
```
we search the folder `asdf7` (Current code)
```zsh
declare -a matching
directory="asdf7"
test_dir=""
for (( i=0; i < ${#directory}; i++ )); do
  test_dir+="${directory:$i:1}"
  echo $test_dir;
  matching=("${PWD}/${test_dir}"*/);
  echo $matching
done
```

If we modify that loop a bit:
```zsh
declare -a matching
directory="asdf7"
test_dir=""
matching=("${PWD}/${directory[1]}"*/);
for (( i=0; i < ${#directory}; i++ )); do
  test_dir+="${directory:$i:1}"
  echo "Testing: $test_dir";
  pattern="${PWD}/${test_dir}*"
  matching=("${(M)matching:#${pattern}}")
  echo $matching
done
```
(not functional, but you get the idea).
This will need some measurings, but I think the second variant will be faster.

functions/autoload/__p9k_get_unique_path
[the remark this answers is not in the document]
dritter: Ah. Okay. Then the comment is misleading..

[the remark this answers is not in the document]
Syphdias: Yeah, I agree, I removed it. The function returns a unique path even if the path contains special characters like space or "*". Only the strategy aims to make it expandable on copy.

I just thought there probably is a magic letter for it. Turns out there is `(q)` but it's to ambitious and excapes `~` as well...
```zsh
 ❱ pwd='~/b/ca cer/d*rwin'; echo ${(q)pwd}
\~/b/ca\ cer/d\*rwin
```

Syphdias:
Hm, locally the vcs_git tests fail, even on next, but the travis tests are ok. Let's see what happens to my travis pipelines...

@dritter I'm not quite sure why you added the `$pathPrefix` in cbb6469c but I don't thing it's necessary in this case. And seems to break stuff even on next

Also you introduced `:A` for `PWD` at a higher level. I think `cd ..` behaves unexpected imho if the symlinks are resolved, as mentioned in this PR. Here an example:
```
 ❱ echo $PWD ${PWD:A}
/home/user /home/user
 ❱ ln -s /etc/exim4 exim_symlink
 ❱ cd /   
 ❱ echo $PWD ${PWD:A}
/ /
 ❱ cd ~/exim_symlink 
 ❱ echo $PWD ${PWD:A}
/home/user/exim_symlink /etc/exim4
 ❱ cd ..            
 ❱ echo $PWD ${PWD:A}
/home/user /home/user
```
Imagine you cd "into" a symlink and see the path the symlink linked to. In most cases `cd ..` will get you do a completely different place

dritter:
The `pathPrefix` is to prepend either `/...` or `~/...` to the truncated path, depending if the user is inside the home directory or not. I did this, because IMHO the truncation strategies should behave all the same in that regard. Another setting that should affect all strategies is `P9K_DIR_PATH_ABSOLUTE`, which is why I created #1063 .
`${current_path:A}` does not mutate the variable. So this is done for the comparison if the user is in his home directory or not..

Syphdias:
> The `pathPrefix` is to prepend either `/...` or `~/...` to the truncated path, depending if the user is inside the home directory or not. I did this, because IMHO the truncation strategies should behave all the same in that regard. Another setting that should affect all strategies is `P9K_DIR_PATH_ABSOLUTE`, which is why I created #1063 .

Well, currently not every strategy used the prefix thing, so I'm opting to not use it. ;-)

In a perfect world I think it should work like this:
1. honor `P9K_DIR_PATH_ABSOLUTE` to generate a valid absolute path (with our without `~/...` style)
2. strategies take whatever path they get, should be `/...` or `~/...`, they can use `P9K_DIR_SHORTEN_DELIMITER`, `P9K_DIR_SHORTEN_LENGTH` and `P9K_DIR_SHORTEN_FOLDER_MARKER` with own defaults (can be different per strategy)
3. any extra effects: `P9K_DIR_PATH_HIGHLIGHT_FOREGROUND`, `P9K_DIR_PATH_HIGHLIGHT_BOLD`, `P9K_DIR_OMIT_FIRST_CHARACTER`, `P9K_DIR_PATH_SEPARATOR`, `P9K_DIR_HOME_FOLDER_ABBREVIATION` and icons

> `${current_path:A}` does not mutate the variable. So this is done for the comparison if the user is in his home directory or not..

Interestingly you removed it in 8002cba89a67b92c5ed1554e95196836a7b8cd1c

Funny how a small PR ca get this big, just by changes around it

dritter:
> Well, currently not every strategy used the prefix thing, so I'm opting to not use it. ;-)

Not yet. But this is on my todo list in my head. I think it is not a good idea to let the various truncation strategies behave differently, even in small details. This leads to "why does feature X not work in strategy A, but does if I use strategy B" like errors quickly. I prefer consistency here.

My perfect world for the truncation strategies [failed](https://github.com/dritter/powerlevel9k/tree/rework_dir_truncation). In this branch I tried to make the strategies more flexible, so that users could choose to apply different truncation strategies and for different [things](https://github.com/bhilburn/powerlevel9k/issues/431). Even [now](https://github.com/bhilburn/powerlevel9k/issues/1057) there are users who want to mix different strategies.

> Interestingly you removed it in 8002cba

Yep. `$HOME` could be a symlink as well. So that comparison would have failed. That is why I removed it.

Syphdias:
Do you want me to to rebase and squash this PR into one commit based on current `next`?

Syphdias:
I rebased to current `next` HEAD and made some more adjustments and corrections.  
Just to be sure I kept the old branch at my repository under the branch `messy-merges`.

Syphdias:
I've resolved more conflicts with this because of code moving to other places than I care to count. I'm not doing it again unless it's going to get merged soon after.

Please let me know if you want to merge and think the changes need improving.

ccjensen:
this is functionality I would love to see. Is there a way to achieve this currently, or is this PR the only solution? If so, what is currently holding up this getting merged in?

dritter:
@ccjensen the only thing that holds us back from merging is that we consider this PR as a new feature, and we are currently working hard to push out a new release (already in feature freeze phase). But we will merge this, once 0.7.0 is out.

Syphdias:
Note to self: This is a great idea for a feature request: https://github.com/bhilburn/powerlevel9k/issues/1277#issuecomment-493754176

TL;DR: `POWERLEVEL9K_SHORTEN_DELIMITER="*"` -> `~/g*/pr*/co*/s*/powerlevel9k`

Issue #974, opened 2018-09-03
/etc/os-release does not exist


EmreAtes:
#### Describe Your Issue
At log in, the error message `grep: /etc/os-release: No such file or directory` is printed

#### Have you tried to debug or fix it?
In the file funcitons/utilities.zsh:105 the grep tries finding an `/etc/os-release` file, and doesn't work if that file doesn't exist.

#### Environment Information
I'm using the master branch of powerlevel9k, and my OS is Centos 6.4 (in `/etc/centos-release`, it says `CentOS release 6.4 (Final)`)

pgierz:
I also have this problem under redhat:

```
$ cat /etc/redhat-release
Red Hat Enterprise Linux Server release 6.10 (Santiago)
```

bhilburn:
Closed by merge of #988. Trying to fix it in `master` in #981, too.

HipHopCoderS:
> I also have this problem under redhat:
> 
> ```
> $ cat /etc/redhat-release
> Red Hat Enterprise Linux Server release 6.10 (Santiago)
> ```

I have same probleme!

dritter:
Good point @jandk008 , I added #981  to the [upcoming version](https://github.com/bhilburn/powerlevel9k/projects/5).

PR #981, opened 2018-09-07
[Bugfix] Fix for #974

pgierz:
Hi, I have a quick fix for #974; we simply check if a file `/etc/os-release` exists, and grep for the appropriate release id if it does. If it does not the case statement reverts back to the default linux icon. This isn't so nice if you *do* in face have one of the pre-made icons (e.g. centos), but if the files are named inconsistently between different hosts, I don't see a good way of getting around that.

### DIFF
--- a/functions/utilities.zsh
+++ b/functions/utilities.zsh
@@ -103,7 +103,7 @@ case $(uname) in
     Linux)
       OS='Linux'
       if [ -f /etc/os-release ]; then
-        os_release_id="$(grep -E '^ID=([a-zA-Z]*)' /etc/os-release | cut -d '=' -f 2)"
+        [[ ${(f)"$((</etc/os-release) 2>/dev/null)"} =~ "ID=([A-Za-z]+)" ]] && os_release_id="${match[1]}"
       fi
       case "$os_release_id" in
         *arch*)
--- a/functions/utilities.zsh
+++ b/functions/utilities.zsh
@@ -102,7 +102,9 @@ case $(uname) in
       ;;
     Linux)
       OS='Linux'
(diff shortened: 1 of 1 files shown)


bhilburn:
@pgierz - Great idea! The only thing I want to change about this PR is the use of `grep` - we are trying to remove as many calls to external programs as possible in the `next` codebase.

@onaforeignshore @rjorgenson - Any chance you can suggest a replacement pure-ZSH alternative to the use of `grep`, here? `=)`

bhilburn:
Okay, I just realized that #988 fixed this in the `next` branch, hah. So, the issue is fixed in `next`, but not in `master`. Also, the fix indeed still uses `grep`.

If we are able to replace the use of `grep` in this PR, I'd like to merge this into master and then replace the same code in `next` with this change, as well.

functions/utilities.zsh
```diff
@@ -102,7 +102,9 @@ case $(uname) in
       ;;
     Linux)
       OS='Linux'
-      os_release_id="$(grep -E '^ID=([a-zA-Z]*)' /etc/os-release | cut -d '=' -f 2)"
+      if [ -f /etc/os-release ]; then
+        os_release_id="$(grep -E '^ID=([a-zA-Z]*)' /etc/os-release | cut -d '=' -f 2)"
```
dritter:
```suggestion
    [[ ${(f)"$((</etc/os-release) 2>/dev/null)"} =~ "ID=([A-Za-z]+)" ]] && os_release_id="${match[1]}"
```
This already includes the file check, so the currently enclosing if could be omitted too.

dritter:
Hmm. Once again the travis tests on OSX hangs. Merging anyway. Thx @pgierz .

PR #989, opened 2018-09-15
0.7.0 Stage 3: Implement autoload functions

onaforeignshore:
This is the third stage of the 0.7.0 update

Changes in this PR

- Moved functions out to autoload.
  - From `segment/dir.p9k`:
    - `__p9k_get_unique_path`
  - From `fcuntions/utilities.zsh`:
    - `__p9k_segment_should_be_joined`
    - `__p9k_truncate_path`
    - `__p9k_upsearch`
  - Added autoload code to `functions/utilities`


functions/utilities.zsh
```diff
@@ -6,6 +6,19 @@
 # https://github.com/bhilburn/powerlevel9k
 ################################################################
 
+################################################################
+# Source autoload functions
+################################################################
+local autoload_path="$__P9K_DIRECTORY/functions/autoload"
+# test if we already autoloaded the functions
+if [[ ${fpath[(ie)$autoload_path]} -gt ${#fpath} ]]; then
```
dritter: What does this line technically do? Is there another way of checking if a function is already autoloaded?

onaforeignshore: Technically, it checks if the fpath already contains our autoload_path or not, and then adds it if required. This is code I got from the ZSH manual for autoload.

PR #997, opened 2018-09-28
[FIX] Fixed error with empty prompt element array

onaforeignshore:
This PR fixed an error message when you have an empty prompt element array


dritter:
Thx @onaforeignshore !

PR #999, opened 2018-09-28
[Enhancement] New Background Jobs segment (070 update)

onaforeignshore:
#### New background jobs segment
This update gets rid of the external calls. In addition, it can now show Running jobs as well as Suspended jobs.
 
#### Description
If `POWERLEVEL9K_BACKGROUND_JOBS_VERBOSE` is set, the segment will display if the job count is more than 0.

If `POWERLEVEL9K_BACKGROUND_JOBS_VERBOSE_ALWAYS` is set, it will always show the segment, with `⚙ 0` - if there are none. (Note that this will _always_ override `POWERLEVEL9K_BACKGROUND_JOBS_VERBOSE` to be true).

If `POWERLEVEL9K_BACKGROUND_JOBS_EXPANDED` is set, the new format will display
`⚙ 2r 1s` - indicating (r)unning and (s)uspended jobs,
otherwise it will display just the total of the two,
`⚙ 3` - indicating the total.

This PR replaces #935

### DIFF
--- a/segments/background_jobs.p9k
+++ b/segments/background_jobs.p9k
@@ -8,14 +8,20 @@
 # Register segment
 # Parameters:
 #   segment_name  context  foreground  background  Generic  Flat/Awesome-Patched  Awesome-FontConfig  Awesome-Mapped-FontConfig  NerdFont
-#                                                            ⚙                                                               
+#                                                                   ⚙                                                               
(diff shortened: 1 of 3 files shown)


dritter:
Thx @onaforeignshore

bhilburn:
This is great, @onaforeignshore, and nice work removing the external calls! Great job `=)`

PR #1011, opened 2018-10-09
[Bugfix] fix: command not found p9k::defined in debug/font-issues.zsh

qkdreyer:
p9k::register_icon in functions/icons.zsh makes calls to p9k::defined which is defined functions/utilities.zsh

Before :
image

After :
image


bhilburn:
Nice job, @qkdreyer!

PR #1015, opened 2018-10-10
Fix `register_segment` comment

lucatrv:
`foreground` and `background` words were swapped.


dritter:
Thx @lucatrv ! Good catch 👍

bhilburn:
Agreed! Thanks so much, @lucatrv! Very much appreciated =)

PR #1017, opened 2018-10-11
Fix dir_writable segment

dritter:
Segments without content (besides a visual identifier) need to overwrite the condition. This condition gets evaluated, hence we need to enclose the condition with `[[` brackets.


bhilburn:
Yay! Thanks for the quick debug & fix on this, @dritter! :)

PR #931, opened 2018-07-24
[New Segment] Haskell Stack Project Segment

DavSanchez:
Hello! I have made a new segment for your theme. It's designed to work with the Stack project manager for the Haskell programming language. Hope you like it!
 
#### Description
This segment appears provided you have a working Stack installation and are in a valid Stack project. The latter is figured out by searching in the directory and parents for a `stack.yaml` file (via the `upsearch` function).

It will print out the Haskell logo (with the NerdFonts option) and the strings `Stack` and the version of your Stack installation, for example `v1.7.1`.

#### Questions
I have some doubts with the PR though. I programmed it roughly in the same way your Go, Laravel or Java segments are programmed. The unit test work the same (I looked at your Go and Laravel ones as well) and they all pass in Travis CI. However, the segment shows the separator in white instead of purple, as you can see in the image. Am I doing something wrong?

screen shot 2018-07-24 at 22 27 17

Thanks!

PR #1018, opened 2018-10-11
[New Segment] Haskell Stack Project Segment (Adapted to your new codebase!)

DavSanchez:
Hello! I have adapted #931 to your new codebase. As I say there, it's designed to work with the Stack project manager for the Haskell programming language. Hope you like it!

**Description**

This segment appears provided you have a working Stack installation and are in a valid Stack project. The latter is figured out by searching in the directory and parents for a `stack.yaml` file (via the `__p9k_upsearch` function).

It will print out the Haskell logo (with the NerdFonts option, else it will print something similar: `λ=`) and the string `Stack`.

**Questions**

The tests work as expected, asserting correctly that the segment gets shown, but I'm trying to get the segment show up in my terminal to no avail. Maybe it's a problem of changing variable names to conform to your new naming convention...

screen shot 2018-10-11 at 16 43 49


segments/stack_project.p9k
```diff
@@ -0,0 +1,34 @@
...
+#   This segment displays if the current directory belongs to a Haskell Stack project.
+##
+# @args
+#   $1 string Alignment - left | right
+#   $2 integer Segment index
+#   $3 boolean Whether the segment should be joined
+##
+prompt_stack_project() {
+
+    local haskellstack_version=$(stack --version 2>/dev/null | grep -oE '[0-9.]+' | head -n1)
```
bhilburn: I realize there is no way to not call out to `stack`, here, but avoiding the external calls to `grep` and `head`, and instead doing that string parsing & slicing purely in ZSH, will be much faster.

We have plenty of people who can help you put together the appropriate ZSH regex, if you'd like some help!

dritter: Here is an example:
```zsh
[[ $(echo "Version 1.7.1 x86_64") =~ "([0-9.]+)" ]] && echo "${match[1]}"
```

segments/stack_project.p9k
```diff
@@ -0,0 +1,34 @@
...
+    local stackyaml_file="stack.yaml"
```
dritter: Is there a reason, why that is a separate variable? It is used only once in the code..

segments/stack_project.p9k
```diff
@@ -0,0 +1,34 @@
...
+    local stackyaml_file_search=$(__p9k_upsearch "${stackyaml_file}")
+    # Check for both a global Stack installation and a stack.yaml files (in current or parent directories)
+    if [[ -n "${haskellstack_version}" ]]; then
```
dritter: I would do the upsearch within this block. If there is no stack, we should not trigger upsearch.

test/segments/stack_project.spec
```diff
@@ -0,0 +1,103 @@
...
+
+# Required for shunit2 to run correctly
+setopt shwordsplit
+SHUNIT_PARENT=$0
+
+function setUp() {
+  export TERM="xterm-256color"
+  # Load Powerlevel9k
+  source powerlevel9k.zsh-theme
+
```
dritter: Is there a reason why `source segments/stack_project.p9k` is done in every test, instead of here?

DavSanchez: Yes. I tried putting that in `setUp()` as you suggested at Gitter, but then the mock icon assigned with `local P9K_STACK_PROJECT_ICON='x'` (at lines 51, 66, 80 and 93) is not loaded correctly, making the assertions fail. Is there another way to do it?

segments/stack_project.p9k
DavSanchez: You're absolutely right. Changing it right away!

segments/stack_project.p9k
DavSanchez: No reason whatsoever, I have changed it so `__p9k_upsearch` uses the string directly without storing it in any variable. Thanks!

segments/stack_project.p9k
DavSanchez: Wow, I didn't know you could do that with ZSH! Fine, I'm changing that lines a bit then.

test/segments/stack_project.spec
dritter: Yep. There are two ways to handle that:
1. Set the icon _before_ you source the segment. This would be then set for all tests.
2. Do not overwrite the icon. You could just copy/paste the glyph into the test. The test then should be resilient against different `P9K_MODE`s..

DavSanchez: Regarding option 2, with the glyph I guess you mean the generic one? Because it would vary if NerdFonts are installed wouldn't it?

dritter: Right. That is why the test should set a `P9K_MODE` itself (and backing up the old one before), so that the test always runs under the same conditions.

dritter:
Thx @DavSanchez

DavSanchez:
Thanks so much to you guys @dritter and @bhilburn. It's been exciting to contribute!

PR #1021, opened 2018-10-12
[Fix] Hide __p9k_detect_os grep output on Android

### LINKED ISSUE #1007: Termux
I want to hide this information . "grep ...."
screenshot_2018-10-08-09-35-45-380_com termux

onaforeignshore:
This PR fixes #1007

### DIFF
--- a/functions/utilities.zsh
+++ b/functions/utilities.zsh
@@ -52,7 +52,7 @@ function __p9k_detect_os() {
     FreeBSD | OpenBSD | DragonFly) __P9K_OS='BSD' ;;
     Linux)
       __P9K_OS='Linux'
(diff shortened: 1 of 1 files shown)


dritter:
I am not sure if that really makes the desired effect. If I do `[[ ${(f)"$(</etc/os-release 2> /dev/null)"} =~ "ID=([A-Za-z]+)" ]] && echo "${match[1]}" || echo "nope"`, I get:
```zsh
zsh: no such file or directory: /etc/os-release
nope
```

dritter:
Alright. I fixed it by wrapping the redirect in Parentheses.. Thx @onaforeignshore

PR #1022, opened 2018-10-12
[Fix] Replace calls to basename

onaforeignshore:
This PR replaces two instances of `$(basename $x)` with `${x:t}`


dritter:
Thx @onaforeignshore

PR #1024, opened 2018-10-13
[Enhancement] Adds support for multiple batteries in battery segment (on linux)

Syphdias:
I added support for two batteries.

Most of the logic stays the same. It does take into account if the capacities of the batteries are different and should work normally if only one battery is present.

The time remaining or until full charge is taken from the two `acpi` lines. I could only see one "active" at a time. Active being a battery being charged or discharged with a time count. That's why I don't aggregate the time valuesand just pick the one that is not empty.

### DIFF
--- a/segments/battery.p9k
+++ b/segments/battery.p9k
@@ -72,38 +72,27 @@ prompt_battery() {
 
   if [[ "$__P9K_OS" == 'Linux' ]] || [[ "$__P9K_OS" == 'Android' ]]; then
     local sysp="${ROOT_PREFIX}/sys/class/power_supply"
+    local bats=(${$(ls -d $sysp/(battery|BAT*))})
 
-    # Reported BAT0 or BAT1 depending on kernel version
-    [[ -a $sysp/BAT0 ]] && local bat0=$sysp/BAT0
-    [[ -a $sysp/BAT1 ]] && local bat1=$sysp/BAT1
+    # Return if no battery found
+    [[ -z $bats ]] && return
 
-    # Android-related
-    # Tested on: Moto G falcon (CM 13.0)
-    [[ -a $sysp/battery ]] && local bat0=$sysp/battery
+    local numerator="0"
+    local denominator="0"
+    local battery_status_full=true
+    local battery_status_charging=false
+    for bat in $bats; do
+      numerator+="+ $(cat $bat/capacity) * $(cat $bat/(energy|charge)_full)"
+      denominator+="+ $(cat $bat/(energy|charge)_full)"
+      [[ $(cat $bat/status) != Full ]] && battery_status_full=false
+      [[ $(cat $bat/status) == Charging ]] && battery_status_charging=true
+    done
 
-    local capacity battery_status
-    if (( ${+bat0} && ${+bat1} )); then
-      # two batteries
-      capacity=$(( ( $(cat $bat0/capacity) * $(cat $bat0/(energy|charge)_full) \
-        + $(cat $bat1/capacity) * $(cat $bat1/(energy|charge)_full) )          \
-        / ( $(cat $bat0/(energy|charge)_full) + $(cat $bat1/(energy|charge)_full) ) ))
-      [[ $(<${bat0}/status) =~ Charging || $(<${bat1}/status) =~ Charging ]] \
-        && battery_status=Charging
-      [[ $(<${bat0}/status) =~ Full && $(<${bat1}/status) =~ Full ]] \
-        && battery_status=Full
-    elif (( ${+bat0} ^^ ${+bat1} )); then
-      # only bat0 OR bat1 isset
-      capacity=$(<${bat0}${bat1}/capacity)
-      battery_status=$(<${bat0}${bat1}/status)
-    else
-      # Return if no battery found
-      return
-    fi
-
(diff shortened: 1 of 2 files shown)


dritter:
Hmm. Looking at the code. Couldn't it be changed to work with an arbitrary number of batteries? So that the code does not bother if it is ` BAT1` or `BAT9`..

segments/battery.p9k
```diff
@@ -72,29 +72,47 @@ prompt_battery() {
...
     else
       [[ $bat_percent =~ 100 ]] && current_state="charged"
       [[ $bat_percent -lt 100 ]] && current_state="charging"
     fi
     if [[ -f ${ROOT_PREFIX}/usr/bin/acpi ]]; then
-      local time_remaining=${${(s: :)$(${ROOT_PREFIX}/usr/bin/acpi)}[5]}
+      local time_remaining
+      declare -a acpi_lines
+      acpi_lines=( "${(@f)$(${ROOT_PREFIX}/usr/bin/acpi)}" )
+      for line in $acpi_lines; do
```
Syphdias: I put echo lines above and below line 110. The above one for `acpi_lines`, the below one for `line`. I get very different outputs depending on if I run the segement test or just use it. I only set a varibale in the test to be able to differentiate between log lines (`ut`):
```
acpi_lines (ut): Battery 0: Charging, 93%, 00:20:00 until charged Battery 1: Unknown, 78%
line (ut): Battery
line (ut): 0:
line (ut): Charging,
line (ut): 93%,
line (ut): 00:20:00
line (ut): until
line (ut): charged
line (ut): Battery
line (ut): 1:
line (ut): Unknown,
line (ut): 78%
acpi_lines (): Battery 0: Charging, 93%, 00:18:17 until charged Battery 1: Unknown, 78%
line (): Battery 0: Charging, 93%, 00:18:17 until charged
line (): Battery 1: Unknown, 78%
```
I tried to be as close to the original out put as possible, but the splitting still works differently on the test. I've run out of ideas how to debug this behavior. Do I use some feature that is not enabled in the test? I'd be happy to get some advice.

dritter: In the test we mock the entire output of the batteries. This is the reason why there is the `${ROOT_PREFIX}` in the segment code. If you look at the Test, you'll find a [`makeBatterySay`](https://github.com/bhilburn/powerlevel9k/blob/f7b99d2077742cd1671d9b58365abe59d1a565a0/test/segments/battery.spec#L44-L66) helper function. Acpi is mocked like [this](https://github.com/bhilburn/powerlevel9k/blob/f7b99d2077742cd1671d9b58365abe59d1a565a0/test/segments/battery.spec#L154-L155).
The test-output was stripped down as much as possible, to keep the test lean and clean. But with moving away from `grep`, `sed` etc. that test is more risky (might be that test works with single lines, but not with multi-lines)..

Syphdias: I'm aware of the mocking of acpi, but I don't get why it behaves differently. For the output shown above. I even changed it to `echo echo -e "Battery 0: ...\nBattery 1:..."`. Evidence of that can be seen in: `acpi_lines (ut): Battery 0: Charging, 93%, 00:20:00 until charged Battery 1: Unknown, 78%`  
Is there a better way to dump arrays?

dritter: Not sure if I understand you correctly. The best way, IMHO, to see what's happening inside the code as data flows through it, is to trace the code (`setopt xtrace`; just typed in the shell). This produces a lot of output, and as a prerequisite I usually configure P9K to only show the segment I want to debug, or enclose the desired code with `setopt xtrace` / `unsetopt xtrace` combo.

Hope that helps. If not, let's move this discussion into the chat.

Syphdias: @dritter thanks for the assist. I finally found the culprit. It's `shwordsplit`. If I unset it inside the battery prompt function, the tests work as expected.

Now I just need the right place to put it in...
1. If I keep the unset in the function it looks like it's only applied locally (maybe just in the function?), since if I set the option manually it is still set after a new prompt is build (pressing enter). So it should be fine in here(?)
2. I could also save the current option state if it is enabled, unset it, do my thing and set it back if it was set. But as mentioned I think it is set locally for the function, though I am not entirely sure.
3. If I unset it in the test, shunit seems to have some problems

I think it would be fine to set the option inside the function at the critical point where I need it (version 1). But, as should be clear by now, I am not certain if my "local" hypothesis holds true. What do you think? What's the right way/place to do it?

Syphdias: This could be an implementation for version 3
```
@@ -107,10 +107,14 @@ prompt_battery() {
       local time_remaining
       declare -a acpi_lines
       acpi_lines=( "${(@f)$(${ROOT_PREFIX}/usr/bin/acpi)}" )
+      [[ -o shwordsplit ]] \
+        && unsetopt shwordsplit \
+        && reenable_shwordsplit=true
       for line in $acpi_lines; do
         [[ -n ${${(s: :)line}[5]} ]] \
           && local time_remaining=${${(s: :)line}[5]}
       done
+      [[ $reenable_shwordsplit ]] && unsetopt shwordsplit
       unset acpi_lines
 
       if [[ $time_remaining =~ "rate" ]]; then
```

PR #1028, opened 2018-10-13
Remove VCS externals

onaforeignshore:
This PR removes VCS externals grap & tail


dritter:
Thx @onaforeignshore .

PR #1029, opened 2018-10-14
Fixes zsh version for Dockerfile for zsh 5.6.2

Syphdias:
```
root@34366f8998aa:/# apt-cache policy zsh
zsh:
  Installed: (none)
  Candidate: 5.6.2-3
  Version table:
     5.6.2-3 500
        500 http://deb.debian.org/debian buster/main amd64 Packages
```

How about something like this to determine version first?
```
$ apt-cache policy zsh |grep 5.5.1 |head -1 |awk '{print $2}'
5.5.1-1ubuntu2
```

### DIFF
--- a/docker/base-5.6.2/Dockerfile
+++ b/docker/base-5.6.2/Dockerfile
@@ -8,7 +8,7 @@ RUN \
   DEBIAN_FRONTEND=noninteractive apt-get install -y \
   curl \
   git \
-  zsh=5.6.2-1 \
+  zsh=5.6.2-3 \
   mercurial \
   subversion \
   golang \


dritter:
Thx @Syphdias .

Hmm. I did it this way, because I wanted to specify the ZSH version and avoid that distributions could upgrade ZSH without us noticing this. That would probably lead to testing the wrong ZSH version. But maybe it is better to just install the distributions latest ZSH version (`apt-get install zsh`).. Hadn't anticipated the frequency the zsh packages change..

Syphdias:
I tried to only use the version number without the distribution/release specifics: `apt install zsh=5.5.1` which, unfortunately, does not work.  
But if you grep only for the line with `"Candidate: $version"` and extract the version from there to put it back into the install command you would still have roughly the version you want with a smaller chance of change. Example:
```
RUN apt-get install zsh=$(apt-cache policy zsh |grep "Candidate: 5.5.1" |head -1 |awk '{print $2}')
```
Ugly, but should work until they change the version number on the patch level

PR #1035, opened 2018-10-24
[Bugfix] truncate_with_package_name outputs "null" if name is missing

xel-x:
Given a package.json without a name for the project the truncate_with_package_name strategy outputs "null" instead of a package name and keeps truncating.

This happens for example within laravel projects, as laravel defaults to using a [package.json](https://github.com/laravel/laravel/blob/master/package.json) without a name in it.

Of course it is possible to work around this with better configuration (trying composer.json before package.json) -- but I presume outputting "null" as the package name can still be considered a bug and should be avoided.

As the name of the projects root folder most likely resembles something like the project name I personally would prefer using that instead of not truncating at all.

This is what I did in this PR.

before:
outputs_null

after:
outputs_base_folder

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -930,6 +930,10 @@ prompt_dir() {
           || node -e 'console.log(require(process.argv[1]).name);' ${pkgFile} 2>/dev/null \
           || cat "${pkgFile}" 2> /dev/null | grep -m 1 "\"name\"" | awk -F ':' '{print $2}' | awk -F '"' '{print $2}' 2>/dev/null \
           )
+        if [ "$packageName" = null ]; then
+          # fix for missing name in pkgFile
+          packageName=$(basename `git rev-parse --show-toplevel`)
+        fi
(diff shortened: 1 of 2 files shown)


bhilburn:
The only concern I have about this approach is that it assumes that there is a git repository present. If that turns out not to be the case, we're back in the same boat.

Perhaps the thing to do is test for git repo and grab the name if possible, and if not, actually just use "unknown". Thoughts, @dritter?

xel-x:
As far as I understand the code it does only work for projects using git at the moment, which is why I did not even check for it.

I guess it would be good to have some "upsearch" to find the package.json or whatever file is set to be looked for and than use that directory.

This would make it possible to use the truncate strategy independently from git and use the correct directory name if no name is given.

bhilburn:
@xel-x - Agreed, that sounds like a safer approach. The `next` branch has an improved "upsearch" capability, by the way, that you could leverage here. Would be good to hear from @dritter on his thoughts on this first, though.

Issue #732, opened 2018-01-29
Git prompt performance


quicknir:
Decided to try out this theme, was very impressed by a number of things. However, a deal breaker is the performance of the git segment. Even in a pretty small git repo, there is noticeable lag after each time I hit enter until it is repainted (not present outside of git repos). On the other hand, sorin theme from prezto (which I use a very small fork of) is 100% instant. I'd suggest adopting some of the implementation from here:  https://github.com/sorin-ionescu/prezto/blob/master/modules/prompt/functions/prompt_sorin_setup.

quicknir:
As an update, I was basically able to get this to work by: 1) copying and pasting all the code from sorin theme, except deleting the lines of code after `# Define prompts`. 2) I took the hello world segment example, and I simply changed it so to output `_prompt_sorin_git` instead of `content`. That's it. And with this modification, the git info in powerlevel9k is super fast.

Obviously, it would be nicer to integrate this properly into powerlevel9k so you get the nice icons and such.

bhilburn:
@quicknir - Thanks so much for the feedback! Yes, the `vcs` segment has been a weak point of P9k for a while, now. It sounds like you have done some really great work, though!

Would you mind sharing what you've done? Feel free to either just post a `git diff` here, or even file a PR so I can see a changeset that you have tested and verified is working.

I would actually like to get this integrated as soon as possible - if we have a clear path to improving the `vcs` segment, I'd like to get it done.

quicknir:
@bhilburn Okay, I'll give this a shot, it may take a few days before I can look at it. You're giving me too much credit, I didn't really do anything but notice that another person (sorin) had done good work :-) I also want to think about customizability for git; traditionally with these things, they only show e.g. `*` to indicate that there are modified files, but I much prefer e.g. `*2` to indicate there are two modified files. I can send a screenshot of my setup if you like, but probably best would be to make it easy to customize (also, being able to customize color especially is rather crucial).

quicknir:
Okay, so I sat down to look at the code today. There are two non-trivial complications. The first is that the approach used by the sorin theme in the prezto package, uses a fair bit of code from another prezto module (prezto has a dedicated module for retrieving git info, it's pretty nice. However, that code might be very similar to the code you already have for extracting various bits of information, so I don't know to what extend you'll want to duplicate that in this repo (or even how that works in terms of license, etc). The thing that really drives the performance in sorin is being async. So in principle that can be fused with some of the existing functions.

[...]

PR #1045, opened 2018-10-30
Speedup vcs

dritter:
This cuts down the use of externals to a minimum. Probably the main issue with speed is the amount of calls to git and huge repositories. I tried to reduce the calls here, but do not expect it to solve issues like #732 ..


bhilburn:
@dritter - I think this is a really good idea. I also think we should have this in the `master` branch. Do you mind submitting another PR that ports this over to the older codebase so we can get it into `master` for anyone using that branch directly?

dritter:
:+1: Will do in the next days.

PR #1049, opened 2018-11-03
Speedup vcs 3

dritter:
These are two small improvements:
1. Use the unaliased git in all cases
2. Avoid using subshells if possible


bhilburn:
Great ideas, @dritter! Glad you thought of this `=)`

Issue #857, opened 2018-06-05
Option to output hyperlink in directory segment


stuaxo:
Now that VTE and other terminals support hyperlinks [1] it would be great if the directory in the prompt was a hyperlink to that directory.

That way you could right click on the prompt and open the file manager there.

[1]

https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

dritter:
Hi @stuaxo !
Wow. I didn't know this is possible at all. Thanks for that suggestion. Pretty awesome, as we could use the shortened name as link text and still point to the right directory. 

I am a bit concerned about:
- Styling. I am not sure how all Terminal Emulators would render the links. iTerm2 renders it like this:
![bildschirmfoto 2018-06-06 um 22 28 01](https://user-images.githubusercontent.com/1544760/41063442-f0f2950a-69d8-11e8-8062-f0bd07fd8559.png)
- [Detectability](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#detecting-availability-of-the-feature). It seems that there is no way to detect this properly. And if some Terminal Emulators won't ignore the escape sequences, it might be necessary to hide this feature in a feature flag (which would be complicated, as the `dir` segment is complicated anyway).
- [File Links](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#file-uris-and-the-hostname) on Remote Hosts. They wouldn't work, right? No big deal, if the File Manager just shows an error.
- [Encodings](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#encodings). Some people use UTF-8 characters in their Directory names.. The Specification says the behaviour is undefined then.

But an interesting feature anyway. 👍

stuaxo:
It's funny, as it's something I've been wanting for years :)

Anyway -

I think it will be ignored if not supported.

It only works on local hosts for now.

The dashed underline isn't my first choice, but it's useful enough that I'd still use it :)

I guess for the home icon could be a file file:/// link to home as well.


I really hope that features like this are taken further, but in the meantime, there are possibilities as-is.... I'd find it handy being able to scroll up in my terminal history and quickly navigate to directories by right-clicking part of the prompt :)

onaforeignshore:
I wrote a quick PR to test this "feature" and, while it works, it causes issues with the right prompt

bhilburn:
[...]

PR #1052, opened 2018-11-05
Dir add hyperlink

onaforeignshore:
This PR adds support for hyperlinking of folders to terminals that support it.

This will require extensive testing and feedback, so please feel free to test it :)

Implements #890 on the new code base (as requested in #857)

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,6 +464,12 @@ If you want to customize the last directory of the path, you can now set `P9K_DI
 
 You can also color the separator separately by setting the color using `P9K_DIR_PATH_SEPARATOR_FOREGROUND`.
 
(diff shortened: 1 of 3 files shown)


dritter (commented):
Code looks good to me. My only concerns are:
- does this get correctly counted when using the zero pattern (`local zero='%([BSUbfksu]|([FB]|){*})'`)? This is used in `truncate_with_package_name` strategy, and probably in the newline segment improvement..
- does this work in all terminals?


Second thought: maybe this is interesting for other agents as well. I am thinking of a segment that reads the to url out of git or a a package.json..

stuaxo:
@dritter 

In non-supported terminals this should just not show up.

General url support would be great if it'll work.

PR #1065, opened 2018-11-14
Protect locale

### LINKED ISSUE #587: locale being overridden
Hi,

I've just spent a loooong time trying to find the reason my system locale was being overridden to en_US and finally tracked it down to this line: https://github.com/bhilburn/powerlevel9k/blob/master/functions/icons.zsh#L21

This declaration of `local` is not in fact localised see: http://zsh.sourceforge.net/Doc/Release/Functions.html#Anonymous-Functions

dritter:
This PR fixes #587 .

### DIFF
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -13,6 +13,7 @@
 
 # Initialize the icon list according to the user's `POWERLEVEL9K_MODE`.
 typeset -gAH icons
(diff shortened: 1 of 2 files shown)


bhilburn:
Hahaha sorry I failed to get this done. I swear I've had this open on my desktop ever since I left that comment in #587! Just couldn't get to it =(

Thanks, @dritter 🥇

PR #1069, opened 2018-11-15
[New Segment] Added Reminder segment

bsidhipong:
#### Reminder segment
A reminder segment named "reminder" that evaluates the current time against an array of reminders and display the associated icon(s).
 
#### Description
Given a P9K_REMINDER_RULES array of strings (day, time, and icons separated by space), evaluate the current time of day against the array and output the appropriate icon(s).

P9K_REMINDER_RULES=(
	"mon-sat 08:30-10:00 $P9K_REMINDER_HOT_BEVERAGE"
	"sun 10:00-18:00 $P9K_REMINDER_TEACUP_WITHOUT_HANDLE"
	"* 20:30-23:59 $P9K_REMINDER_SLEEPING_FACE"
	"* 21:30-23:59 $P9K_REMINDER_PERSON_IN_BED"
)

reminder-screenshot

#### Questions
It'd be great if you can have a look and suggest what unit tests might be necessary.

### DIFF
--- a/segments/reminder.p9k
+++ b/segments/reminder.p9k
@@ -1,0 +1,71 @@
+# vim:ft=zsh ts=2 sw=2 sts=2 et fenc=utf-8
+################################################################
+# @title powerlevel9k Segment - Reminder
+# @source [powerlevel9k](https://github.com/bhilburn/powerlevel9k)
+##
+
+################################################################
+# Register segment
+# Parameters:
+#   segment_name  context  background  foreground  Generic  Flat/Awesome-Patched  Awesome-FontConfig  Awesome-Mapped-FontConfig  NerdFont
+#                                                                                                                             
+p9k::register_segment 'REMINDER' '' green "$DEFAULT_COLOR" '!!' ''   $'\UF0A4 '   '\u'"${CODEPOINT_OF_AWESOME_HAND_O_RIGHT}"   $'\UF0A4'
+
+################################################################
+# @description
+#   Display the reminder icon/text for the current time interval.
+##
+# @args
+#   $1 string Alignment - left | right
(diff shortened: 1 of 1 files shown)


segments/reminder.p9k
```diff
@@ -0,0 +1,71 @@
...
+    "fri" 5
+    "sat" 6
+  )
+	local dow=$( date +"%w" )
+	local now=$( date +"%H%M" )
+  local segment_content=""
+	local separator=""
+	for rule in $P9K_REMINDER_RULES; do
+		echo $rule | while read date_exp time_exp icon; do
+			time_exp=$( echo "$time_exp" | sed -e "s/[:]//g" )
```
dritter: This could be done with ZSH itself:
```suggestion
			time_exp=${time_exp//:/}
```

segments/reminder.p9k
```diff
@@ -0,0 +1,71 @@
...
+    "thu" 4
+    "fri" 5
+    "sat" 6
+  )
+	local dow=$( date +"%w" )
+	local now=$( date +"%H%M" )
+  local segment_content=""
+	local separator=""
+	for rule in $P9K_REMINDER_RULES; do
+		echo $rule | while read date_exp time_exp icon; do
```
dritter: Interesting way of reading in variables. What would happen if the icon is empty?

dritter:
Thanks for that nice segment @bsidhipong ! Much appreciated.

I have a few requests:
- Could you add some documentation on how to use this segment?
- Could you add some tests regarding the various configuration options? What happens if the icon is empty? What if it contains a whitespace? What if the date contains whitespace? What happens if someone uses seconds on his configuration? etc.

PR #1070, opened 2018-11-15
[Docs] Uniformly apply inline code formatting in README

Wieschie:
All POWERLEVEL variables and directory paths should be formatted with backticks for inline code.  If not marked as code, the ~ home directory can unintentionally apply strikethrough.


dritter:
Thx @Wieschie for keeping an eye on the docs!

bhilburn:
Thanks for the clean-up, @Wieschie! PRs like this, while perhaps simple to fix, are just as important to the project as code. Thanks so much!

PR #1073, opened 2018-11-16
[Bugfix] Fix +vi-git-untracked when using Git submodules

cwis-public:
When using Git submodules, the `+vi-git-untracked` was unable to check for new files because it was using `git --ls-files` on the `${vcs_comm[gitdir]}` directory, which is the `.git` directory inside the `.gitmodules` directory, and not the checked-out directory itself.

This fix detects the error message given by `git ls-files`, and, when this happens, retries the command using `git rev-parse --show-toplevel`.


dritter:
Hi @cwis-public ! Thanks for the fix, but I am closing this anyway in favor of #1071  which landed in `master` with #1080 .

bhilburn:
As @dritter said, thanks for debugging the problem and submitting the fix, @cwis-public! Sorry the timing overlapped on the parallel effort. Looking forward to your next PR!

PR #1075, opened 2018-11-17
Protect locale

dritter:
This protects the users locale. Before, we always set `LC_CTYPE` to `en_us.utf8`. Now we do that only internally. I had to wrap all occurrences of icons with a anonymous wrapper function to create scope so that `local` works correctly. I also moved all calls to `p9k::set_default` into that function, because they may contain icons as well (see dir segment).


dritter:
Merging this quickly, because:
a) there are not too much PRs open that contain new segments (need to be updated)
b) this fix should not hurt anything
c) tests are green

PR #1076, opened 2018-11-17
Enable colors on Travis

dritter:
shUnit is capable of outputting ANSI-Colors. By default shUnit detects the coloring capabilities of the terminal automatically. This seems to fail for travis. This PR enforces colored output.


bhilburn:
Hahah 👍

PR #1079, opened 2018-11-17
[Bugfix] Update VIRTUAL_ENV_DISABLE_PROMPT value

josselinauguste:
Following a prezto update, the flag value of the VIRTUAL_ENV_DISABLE_PROMPT variable is now 12 (https://github.com/sorin-ionescu/prezto/blob/e9387a177e04b05dcf4f82e622615ddd32c558db/modules/python/init.zsh#L103).

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1648,7 +1648,7 @@ prompt_vi_mode() {
 # https://virtualenv.pypa.io/en/latest/
 prompt_virtualenv() {
   local virtualenv_path="$VIRTUAL_ENV"
-  if [[ -n "$virtualenv_path" && "$VIRTUAL_ENV_DISABLE_PROMPT" != true ]]; then
+  if [[ -n "$virtualenv_path" && -z "$VIRTUAL_ENV_DISABLE_PROMPT" ]]; then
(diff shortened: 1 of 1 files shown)


dritter:
Hi @josselinauguste , thanks for this PR. TBH, I don't know if this is right or not. The relevant commit in prezto is https://github.com/sorin-ionescu/prezto/commit/ee885d42decf58ba90ddff250bb9c05f66009397 and they reference this change in pure https://github.com/sindresorhus/pure/commit/afa625b3bcae6c45126f5b731f9dc230bc52a5b4 .

Why don't we just check if `VIRTUAL_ENV_DISABLE_PROMPT` is not empty? That would work with any value. On the downside this would disable the segment, even if the value would be set to `false`.. But to me it seems better than checking magic values..

josselinauguste:
TBH, I don't know neither :-) The only thing I know is that the previous version didn't work, and this one does work.

I'm gonna try by checking if it's empty.

josselinauguste:
It seems to work, the PR is updated!

bhilburn:
Great recommendation, @dritter! 

@josselinauguste - Thanks for putting together this PR and iterating on it with @dritter! This is a great first contribution to P9k - welcome to the project! `=)`

PR #1086, opened 2018-11-18
[Bugfix/Enhancment] battery segment: Hides `(...)` if battery status is unknown on Linux

Syphdias:
This aims to only only show `(...)`instead of the remaining charge/discharge time, if the state of the batteries is unclear, i.e. while the status is `Charging` or `Discharging` but the charge rate, e.g. `power_now`.

If the `power_now` is 0 and the battery status is `Unknown` however. This will be ignored now. This can be the case one or multiple batteries are nearly full (\~98%) but stopped charging. Which seems to be the case quite frequently (examples from @dritter, @JulienPivard and myself)

I also added some test to ensure i behaves correctly.

### DIFF
--- a/segments/battery.p9k
+++ b/segments/battery.p9k
@@ -77,7 +77,7 @@ prompt_battery() {
   if [[ "$__P9K_OS" == 'Linux' ]] || [[ "$__P9K_OS" == 'Android' ]]; then
     # needed if files or directory do not match globbing pattern
     local sysp="${ROOT_PREFIX}/sys/class/power_supply"
-    declare -a bats
+    local -a bats
     bats=( $sysp/(BAT*|battery)(N) )
     [[ ${#bats} == 0 ]] && return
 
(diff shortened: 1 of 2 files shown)


segments/battery.p9k
```diff
@@ -77,38 +77,47 @@ prompt_battery() {
...
       energy_full+="+ 0$(< ${energy_full_tmp[1]:-/dev/null})"
-      local -a power_now_tmp
       power_now_tmp=( $bat/(power|current)_now(N) )
       power_now+="+ 0$(< ${power_now_tmp[1]:-/dev/null})"
       # get cumulative battery status (ignore if no status file exists)
-      local -a battery_status
-      battery_status=( $bat/status(N) )
-      [[ $(<${battery_status[1]:-/dev/null}) != Full && -n $(<${battery_status[1]:-/dev/null}) ]] \
+      battery_status_tmp=( $bat/status(N) )
+      battery_status=$(<${battery_status_tmp[1]:-/dev/null})
```
dritter: Is `$battery_status` global?

segments/battery.p9k
```diff
@@ -77,38 +77,47 @@ prompt_battery() {
...
+    local battery_status_calculating=false
     local battery_status_charging=false
     for bat in $bats; do
       # skip this loop if no files in $bat
       local -a bat_tmp
       bat_tmp=( $bat/*(N) )
       [[ ${#bat_tmp} == 0 ]] && continue
       # add "+ 0[number in file]" if the file exists (0 padding if glob fails)
-      local -a energy_now_tmp
+      declare -a energy_now_tmp
```
dritter: What is the advantage to use `declare -a` vs `local -a`?

Syphdias: Didn't you tell me about it being compatible with zsh version 5.0 and it would be nice if most of the code would be compatible even if not officially supported? Did I misunderstand that?
I'll happily get rid of the extra lines.

Also, apparently I missed L92.

segments/battery.p9k
Syphdias: Apparently, it is but it should not be global.

segments/battery.p9k
dritter: Yes I did. But is `local -a` invalid in ZSH 5.0? AFAIR only declaring and setting a value in one line is illegal.

dritter:
screenshot_20181120_020827

Besides the one odd value in between it looks good.

Syphdias:
@dritter For complaints about battery charge time please contact your laptop vendor ;)
I had similar behavior once but besides that, I couldn't reproduce it. It would be interesting to see the file contents of files in `/sys/class/power_suppy/(battery|BAT?)/` at that exact moment to understand if it is a quirk in the numbers being low/high at that point or a calculation going wrong.

PR #1102, opened 2018-11-27
[Enhancement] Allow to user to specify number of last directories to show with truncate_to_last

MichaelAquilina:
#### Description
Opening this PR up a bit early as a proof of concept and hoping to get some guidance about whether you think this the right approach to re-use an existing strategy or just an entirely new one. I'll update the README + tests once I have a clearer idea of how to approach this :)

This PR adds the ability for the user to specify how *many* directories to show before truncating all characters before it with the `truncate_to_last` strategy.

#### Questions
I guess the main question is should we re-use the `truncate_to_last` strategy or should we add an entirely new one?

There is also some discussion about the name of the variable used to specify the number of parent folders to show (which I've set to -1 to be backwards compatible).

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,6 +407,7 @@ Customizations available are:
 |`P9K_DIR_SHORTEN_LENGTH`|`2`|If your shorten strategy, below, is entire directories, this field determines how many directories to leave at the end. If your shorten strategy is by character count, this field determines how many characters to allow per directory string.|
 |`P9K_DIR_SHORTEN_STRATEGY`|None|How the directory strings should be truncated. See the table below for more informations.|
 |`P9K_DIR_SHORTEN_DELIMITER`|`..`|Delimiter to use in truncated strings. This can be any string you choose, including an empty string if you wish to have no delimiter.|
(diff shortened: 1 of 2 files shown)


dritter:
Hi @MichaelAquilina !

Thanks for you PR. Thinking about this, this is the same as the default truncation strategy.. Even `truncate_to_last` is redundant..

This should give you the same output:
```zsh
# Number of directories that you want to see
P9K_DIR_SHORTEN_LENGTH=3
# Trigger the default truncation strategy. Any name will do.
P9K_DIR_SHORTEN_STRATEGY=dir
```

MichaelAquilina:
Will try that instead then. Thanks @dritter !

MichaelAquilina:
Is this available on `master`? Trying out your recommendation does not seem to work for me there but seems to work on `next`

dritter:
Yep. That should work with:
```zsh
POWERLEVEL9K_SHORTEN_DIR_LENGTH=3
POWERLEVEL9K_SHORTEN_STRATEGY=dir
```

PR #1089, opened 2018-11-19
[WIP] Reorganize segments

dritter:
This PR reorganizes segments and draws together test, documentation and code. I did this as a proof of concept for the `anaconda` segment. The goal would be to reorganize all segments accordingly.
It boils down to create a directory per segment and move code and tests there, and create a new README in this directory for this segment.

### Motivation
There are multiple motivations for this:
1. More focus in the main README. As documentation grows in one file, it looses focus and gets overwhelming. Important information gets drown in the information flood. On the other hand, we always struggle with keeping the information short and precise, probably by leaving out details or going not so much in depth with every segment. By splitting the README into separate files, the main README gets more room for important things and we have more space to go into detail for every segment.
2. It is hard for users to follow the documentation. Some information is in the main README, some is in the wiki (e.g. https://github.com/bhilburn/powerlevel9k/wiki/Stylizing-Your-Prompt#special-segment-colors). This confusion could be solved by documenting the segment states in the according README in the segment.
3. Drawing together what belongs together. Keeping Code and Tests in separate directories creates a lot of distance.. By splitting the code feature-wise, we get a better architecture.

PR #1112, opened 2018-12-08
[Enhancement] A setting to hide stashed changes count for git

Thecentury:
#### Description
Added a setting `P9K_VCS_HIDE_STASHES` that allows to hide git stashed changes count.
Its default value is `false`.

Here is how status without stashed changes count looks like:
image


dritter:
Hi @Thecentury ,

thanks for the PR. TBH, I am a bit ambivalent here. Not sure if we want to add a switch for this. Disabling stashes could also be done by removing `git-stash` from `P9K_VCS_GIT_HOOKS`. That way VCS_INFO wouldn't even bother to call that hook in the first place.
Maybe this is more a documentation issue that I hope to solve with #1089 ..

Thecentury:
Hi @dritter, thank you for your reply. Didn't knew this can be done in this way.

dritter:
Alright. I documented the VCS hooks here f73e260c5bc6521f6d09353ccd542fad1063ad6e . That commit will land in `next` soon. Thanks @Thecentury

PR #1118, opened 2018-12-11
Get rid of absolute versions for Dockerfiles

dritter:
The Dockerfiles break regularly because of distros releasing patches for ZSH and it seems to be impossible to install a fuzzy version of ZSH (like 5.6.2-*) with `apt-get`. This PR switches to `apt` that has that ability.

//cc @Syphdias

### DIFF
--- a/docker/base-4.3.11/Dockerfile
+++ b/docker/base-4.3.11/Dockerfile
@@ -4,7 +4,7 @@ RUN \
   yum install -y \
   curl \
   git \
-  zsh \
+  zsh-4.3.11 \
   mercurial \
   subversion \
   golang \
--- a/docker/base-5.0.3/Dockerfile
+++ b/docker/base-5.0.3/Dockerfile
@@ -1,12 +1,12 @@
 FROM ubuntu:14.04
 
 RUN \
-  apt-get update && \
+  apt update && \
   echo 'golang-go golang-go/dashboard boolean false' | debconf-set-selections && \
-  DEBIAN_FRONTEND=noninteractive apt-get install -y \
+  DEBIAN_FRONTEND=noninteractive apt install -y \
   curl \
   git \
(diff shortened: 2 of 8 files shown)


Syphdias:
Looks good to me. Though I cannot speak for 4.3.11 since the centos image thows a segvault on running yum, bash or sh (dash works). I think that's a local problem with my machine.

Syphdias:
Currently the `centos:6` image seems to be broken (also for a colleague). The `centos:7` image works.

dritter:
Hmm. What exactly do you mean by saying the image is "broken"? Is it not possible to start the docker container? Or is just P9K broken inside the container? Thinking about it, that is only there because of historical reasons. Our new minimum system requirement is ZSH 5.1..

Syphdias:
As I said, I (and my colleague) get segfaults for starting various programs inside the container. We both use Debian (different version I think).

And since I can run the version with the label 7 from dockerhub. I assume the latest image from dockerub is faulty. Could be my system though.

```zsh
 ❱ docker run --rm -it centos:7 /bin/bash                                            
[root@3c19cabe14ad /]# ^C
[root@3c19cabe14ad /]# exit
^[[A%                                                                                                                                                                                                              chkugler@biscuittin:~% docker run --rm -it centos:6 /bin/bash
chkugler@biscuittin:~% echo $?     
139
 ❱ docker run --rm -it centos:6 /bin/dash
# yum
Segmentation fault (core dumped)
# echo $?
139
# /bin/bash
Segmentation fault (core dumped)
# echo $? 
139
# % 
```

Syphdias:
Ok apparently this is a [know bug](https://github.com/CentOS/sig-cloud-instance-images/issues/103)

dritter:
Alright. I merged it. Thanks for the review @Syphdias .

PR #1120, opened 2018-12-18
Windows battery support

dritter:
This ports over the changes by @ratijas from #887 onto a current `next`. Additionally I removed calling externals as much as possible and wrote tests for it.

@ratijas can you confirm this works for you as expected?


ratijas:
LGTM!

However my battery is dead now, so sadly I have no way to verify this on my own.

dritter:
Okay. Thanks for letting me know @ratijas .

I merge this even without confirmation, because:
- we had no battery support for Windows before
- tests are green with data from 3rd party
- this is too much code, that will diverge quickly, if not merged

PR #1121, opened 2018-12-27
[New Segment] Add gcloud profile segment to prompt

primeroz:
#### Description
Add a segment to show the currently enabled profile for gcloud ( google cloud cli tool ) , similar to the kubecontext one

The name of the active configuration is taken from the content of file `$HOME/.config/gcloud/active_config` 

For safety reason the default color scheme is set to "RED background with WHITE text" , this is changed to "GREEN background with BLACK text" if the name of the profile match "dev|development" 

In the following example the profile name is "Default"
segmentgcloud

#### Questions
* Is there a better way to manage color schemes based on the segment contents rather than hardcode it as i did ?
* The code was written against the `master` branch, i can see an error on trying to merge to `next`. What do you need me to do ?


primeroz:
Messed up branches

PR #1131, opened 2019-01-07
[New Segment] - add prompt segment for nodenv

greg-cook:
#### Description
- Adds a prompt segment for [`nodenv`](https://github.com/nodenv/nodenv)
- Copied from existing [`rbenv`](https://github.com/bhilburn/powerlevel9k/blob/next/segments/rbenv.p9k) segment

#### Questions
Please let me know whether specs are sufficient, copied from [nvm segment spec](https://github.com/bhilburn/powerlevel9k/blob/next/test/segments/nvm.spec) with some removed due to differences in the way they work.

#### Screenshot
image

### DIFF
--- a/segments/nodenv.p9k
+++ b/segments/nodenv.p9k
@@ -1,0 +1,41 @@
+# vim:ft=zsh ts=2 sw=2 sts=2 et fenc=utf-8
+################################################################
+# @title powerlevel9k Segment - Node Environment
+# @source [powerlevel9k](https://github.com/bhilburn/powerlevel9k)
+##
+
+(){
+  # Set the right locale to protect special characters
+  local LC_ALL="" LC_CTYPE="en_US.UTF-8"
+  ################################################################
+  # Register segment
+  # Parameters:
+  #   segment_name  context  background  foreground  Generic  Flat/Awesome-Patched  Awesome-FontConfig  Awesome-Mapped-FontConfig  NerdFont
+  #                                                                                                                      
+  p9k::register_segment "NODENV" "" "green" "${DEFAULT_COLOR}"  ''  $'\uE847 '  $'\uF219 '  '\u'$CODEPOINT_OF_OCTICONS_NODE' '  $'\uF219 '
+
+  ################################################################
+  # Register segment default values
+  p9k::set_default P9K_NODENV_PROMPT_ALWAYS_SHOW false
+}
+
+################################################################
+# @description
+#   Display the current nodenv information.
+##
+# @args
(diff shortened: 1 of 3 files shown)


dritter:
Hi @greg-cook !
Thanks for your contribution. One question: Is `nodenv` related to `nodeenv` (which we already have a segment for).. The current segment refers to [this project](https://github.com/ekalinin/nodeenv).

segments/nodenv.p9k
```diff
@@ -0,0 +1,41 @@
...
+# @description
+#   Display the current nodenv information.
+##
+# @args
+#   $1 string Alignment - left | right
+#   $2 integer Segment index
+#   $3 boolean Whether the segment should be joined
+##
+prompt_nodenv() {
+  if [[ -n "${NODENV_VERSION}" ]]; then
```
dritter: Despite looking very convenient, that shortcut ignores `$P9K_NODENV_PROMPT_ALWAYS_SHOW`. So, if somebody uses that setting, it won't work once `NODENV_VERSION` is set (regardless, if this is the same version as `nodenv global`).

segments/nodenv.p9k
```diff
@@ -0,0 +1,41 @@
...
+    p9k::prepare_segment "$0" "" $1 "$2" $3 "${NODENV_VERSION}"
+  elif [ ${commands[nodenv]} ]; then
+    local nodenv_version_name="$(nodenv version-name)"
```
dritter: Not sure how fast `nodenv` is, but we call it three times here. Is there a way to get the infos at once? Or a way to get the infos in a different way (thinking of our `__p9k_upsearch ` function)?

romkatv:
@greg-cook Would you mind sharing why you closed this PR?

greg-cook:
Honestly, I haven't been able to spend any time on this (and at this point I'm no longer using `nodenv`)

It is of course still a feature worth having if somebody else wanted to pick it up!

Larusso:
Funny I just wanted to open a patch for this as well.
I think it would make sense to support `nodenv` along with `rbenv` and `pyenv`.
I would even add `jenv` for that matter. So could we integrate it?

romkatv:
It's a remarkable coincidence that everyone who requests integration with `nodenv` has cookie monster for avatar. Or is it? :thinking:

kyledakid:
I need the nodenv segment also :( I dont want to use nodeenv (double e), it relies on virtualenv.
@greg-cook can you pls share me a way for this ?

PR #1132, opened 2019-01-09
[Bugfix] allow set -u to be used

docwhat:
The surface functions and `if` cases may be called by a script, which means they should work even with `set -u`set.

### DIFF
--- a/generator/default.p9k
+++ b/generator/default.p9k
@@ -314,7 +314,7 @@ function __p9k_build_left_prompt() {
 # @noargs
 ##
 function __p9k_build_right_prompt() {
-  [[ ${P9K_RIGHT_PROMPT_ELEMENTS[1]} == "" ]] && return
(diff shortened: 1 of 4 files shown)


dritter:
Hi @docwhat ,
Thanks for the PR. In the meantime I merged some big changes, could you update your PR?

Is there a way to test this automated? I'm asking, because I feel that this is something I'll eventually forget in future changes.. ;)

Thx.

docwhat:
I'm not sure it is worth it without a test.

Basically, the test would be something like

``` zsh
function someTest() {
  setopt local_options
  set -u
  # Run the two scripts I modified.
}
```

docwhat:
Okay, I pushed up changes with a test.

Something still isn't working.  If anyone wants to help?

dritter:
I just restarted the Tests. That was a hiccup in the Travis OSX VMs again..
Now everything is green. Merging. Thx @docwhat

PR #1137, opened 2019-01-15
[New Segment] - Add environment segment

alexperegrina:
#### Description
The `environment` segment will print the environment from hostname.

This segment has four configurable work environments that are obtained by looking at the hostname and taking the value after the first point. In case of obtaining not contemplated a special environment is assigned

Each of the states is identified with a key. The key is compared to the environment obtained from the hostname and in this way the state is obtained. 
`P9K_<name-of-segment>_KEY_<state> == $ENVIRONMENT`

Each of the states shows a text that is obtained from the variable `P9K_ <name-of-segment> _VALUE_ <state>` of each segment.

##### Local
image

##### Dev
image

##### Test
image

##### Prod
image

##### Other
image

### DIFF
--- a/.test.zshrc
+++ b/.test.zshrc
@@ -1,3 +1,0 @@
-#source /Users/alex/workspace/powerlevel9k@new-environment/powerlevel9k.zsh-theme
-source /Users/alex/workspace/powerlevel9k/powerlevel9k.zsh-theme
-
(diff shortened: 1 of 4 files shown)


dritter:
Hi @alexperegrina ,

Thanks for the PR. At a quick glance the code looks good to me (👍 for tests, images, structure; awesome!), but we want to get out a couple of versions first. One based off of `master`, one off of `next`. In other words: It'll take some time till I can have a closer look (sorry). Just wanted to let you know.

PR #1139, opened 2019-01-18
[Enhancement] Adding own gitconfig for git tests

Syphdias:
This PR does some modifications to the git unit tests:
- sets `HOME` to test folder so the global `.gitconfig` is created in this folder. (I just realized that the setting of `OLD_GIT_AUTHOR_EMAIL` and `OLD_GIT_AUTHOR_NAME` is also made unnecessary by this since the real global gitconfig is never touched)
- also: removes redundant setup and tear-down code
- also: modifies two tests to be more explicit and consistent with other commands (not output)

### DIFF
--- a/segments/vcs/vcs_git.spec
+++ b/segments/vcs/vcs_git.spec
@@ -259,7 +259,7 @@ function testActionHintWorks() {
   cd ../vcs-test2
   echo "yy" >> i-am-modified.txt
   git commit -a -m "Provoke conflict" &>/dev/null
-  git pull &>/dev/null
+  git pull --no-ff  &>/dev/null
 
   assertEquals "%K{003} %F{000} master %F{001}| merge 1/1 ↑1 ↓1%f %k%F{003}%f " "$(__p9k_build_left_prompt)"
 }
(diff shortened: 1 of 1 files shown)


segments/vcs/vcs_git.spec
```diff
@@ -16,6 +16,12 @@ function setUp() {
   cd $FOLDER
   P9K_MODE=default
 
+  # Prevent the use of system or user specific gitconfig
+  OLD_HOME="$HOME"
+  HOME="$(dirname ${FOLDER})"
```
dritter: This should be equivalent to `dirname`
```suggestion
  HOME="${FOLDER:h}"
```

Syphdias:
> Removing the `git init`s from the test makes them dependant on each other.. Was this done because of performance reasons?

As I understand it, the `setUp` and `tearDown` run separately for each test function. So the the `cd $FOLDER` and `git init` is already done in `setUp`. So as I understand it I don't introduce extra dependencies.

dritter:
Ha! You are right. I didn't see that because it was already there and therefore didn't show up in the diff.

Issue #1146, opened 2019-01-31
Zsh 5.7 interferes with newline 


quintrino:
#### Describe Your Issue
Since updating from Zsh 5.6.2 to 5.7 (The latest available with homebrew on Mac) the newline element appears to push the > end onto the following line. 

![image](https://user-images.githubusercontent.com/7043587/52029263-9e0c2a00-2566-11e9-81c5-151b68dec26c.png)
Here is my console when I revert back to zsh 5.6.2
![image](https://user-images.githubusercontent.com/7043587/52029310-c431ca00-2566-11e9-8307-b14d52e75328.png)
And here it is again after switching brew back to 5.7. 

The white tip of the custom element showing ~/.dotfiles is appearing on the next line. 

I also tested if it was just custom elements by setting the last element on the first line of my prompt as the os_icon

![image](https://user-images.githubusercontent.com/7043587/52029420-37d3d700-2567-11e9-86b0-b2740d66726f.png)

#### Have you tried to debug or fix it?

Only by reverting to 5.6.2

#### Environment Information

  - Works on 5.6.2, breaks on 5.7
  - My framework is zplugin which I used to install P9k
  - My P9k plugin version is 2019-01-23 6 ccdfdf Merge pull request #981 from pgierz/os_release_fix
  - My terminal emulator is iTerm version 3.27

#### Issues with Fonts & Icons

![image](https://user-images.githubusercontent.com/7043587/52029883-5e930d00-2569-11e9-8aec-237f985b42e5.png)

Seems to happen multiple times if I include multiple newline elements.

You may delete this section if your issue is not font / icon related.

  - Which font do you use?

SauceCodePowerline-Regular 11

  - Which [font configuration mode](https://github.com/bhilburn/powerlevel9k/wiki/About-Fonts) are you using? You can check this with (`echo $POWERLEVEL9K_MODE`).

echo $POWERLEVEL9K_MODE returns nothing but a plank line. 

  - Please share the contents of `$P9k/debug/font-issues.zsh`.

```
===== Font debugging =====
You are using iTerm2 with Font SauceCodePowerline-Regular 11

Your font settings seem to be all right. If you still have issues,
it is more likely to be a font issue than a Powerlevel9k related one.
```

  - If this is an icon problem, does the output of `$ get_icon_names` look correct?

POWERLEVEL9K_MULTILINE_FIRST_PROMPT_PREFIX: ╭─
[...]

PR #1147, opened 2019-02-01
Fix newlines in ZSH 5.7

dritter:
This fixes the newlines in ZSH 5.7 (#1146 ).

@quintrino @nicolasmeric @panosru Could you confirm this works for you?


quintrino:
I wasn't actually sure how to pull down a different version using zplugin, but when I made the changes manually the issue was resolved, so it looks like this is good to merge. 

Thanks for your super swift resolution and for helping make P9K ever better!

dritter:
Cool. Thanks for confirmation @quintrino . Merged.

Issue #1150, opened 2019-02-01
`dir` replaces `$HOME` anywhere in cwd, not just beginning


Wuestengecko:
#### Describe Your Issue

The `dir` segment replaces `$HOME` anywhere in the current working directory's path, not just at the beginning.

![screenshot](https://user-images.githubusercontent.com/1579756/52129426-c6b02480-2637-11e9-9c46-19e7c174bb66.png)

#### Have you tried to debug or fix it?

~~I just noticed this, and have not yet done anything in order to fix it.~~

The quick and easy fix for this is to change [line 800](https://github.com/bhilburn/powerlevel9k/blob/bd340d4a88797e760efb28c1180254e7779c990f/powerlevel9k.zsh-theme#L800) (or the segment's [line 54 on `next`](https://github.com/bhilburn/powerlevel9k/blob/5eb17c476c72505daa0119fb4bbb24a7ce75724a/segments/dir/dir.p9k#L54)) to

    ... && current_path=${current_path/#$HOME/"~"}

(notice the `#` before `$HOME`). This will make above screenshot keep `~` instead of `/root`, but leave `/mnt/root` as-is.

#### Environment Information
This information will help us understand your configuration. 

  - What version of ZSH are you using? `zsh 5.7 (x86_64-pc-linux-gnu)`
  - Do you use a ZSH framework (e.g., Oh-My-ZSH, Antigen)? No.
  - How did you install P9k (cloning the repo, by tarball, a package from your OS, etc.,)? Cloned repo.
  - What version of P9k are you using? `v0.6.6`, also reproduced on `master` and `next`
  - Which terminal emulator do you use? `kitty`

dritter:
Good finding @Wuestengecko . Would you create a PR for that? That would be awesome!

PR #1151, opened 2019-02-01
[Bugfix] Only abbreviate $HOME at the beginning of cwd

Wuestengecko:
#### Description

As requested in #1150.


dritter:
Thx @Wuestengecko and welcome to P9K!

PR #1155, opened 2019-02-04
Lint: prompt_background_jobs bad math expressions

docwhat:
Some simple lint.  It should also cause the occasional variable problem go away.

```
prompt_background_jobs:2: bad math expression: operand expected at end of string
```


dritter:
Thx @docwhat

PR #1156, opened 2019-02-05
[Enhancement] optionally fallback to repo name with `truncate_to_package_name`

wrboyce:
Adds three new settings: `P9K_DIR_SHORTEN_GIT_FALLBACK`, `P9K_DIR_SHORTEN_GIT_FALLBACK_USE_REMOTE`, and `P9K_DIR_SHORTEN_GIT_FALLBACK_REMOTE_NAME` for the `truncate_to_package_name` path shortening strategy.

If `GIT_FALLBACK` is non-zero and a package.json/composer.json does not exist then the repo directory name will be used as the "root" directory in the prompt path label. If `USE_REMOTE` is non-zero, then the remote of name `REMOTE_NAME` (defaulting to "origin") will be used in-place of the local dir name.


segments/dir/dir.p9k
```diff
@@ -147,22 +147,32 @@ prompt_dir() {
...
+            )
+          if [[ -n "${packageName}" ]]; then
+            # Instead of printing out the full path, print out the name of the package
+            # from the package.json and append the current subdirectory
+            current_path="`echo ${packageName} | tr -d '"'`$subdirectory_path"
+          fi
+        elif [ -n "${package_path}" ] && [ -n "${P9K_DIR_SHORTEN_GIT_FALLBACK}" ]; then
+          local repo_name
+          if [ -n "${P9K_DIR_SHORTEN_GIT_FALLBACK_USE_REMOTE}" ]; then
+            repo_name="${${$(git config --get remote."${P9K_DIR_SHORTEN_GIT_FALLBACK_REMOTE_NAME-origin}".url):t}%.git}"
```
Syphdias: I don't see how `P9K_DIR_SHORTEN_GIT_FALLBACK_REMOTE_NAME` would get much use since changing it would change it globally and there are probably very little people who just have repositories with a remote that is other than `origin`. Another idea could be to find out some remote with `${$(command git remote)[1]}`.

Also, I'd use `command` for external commands to avoid hitting an alias of function.

wrboyce: I was trying to keep external commands to a minimum, which is why I opted against automatic discovery of the remote. I think your point about `P9K_DIR_SHORTEN_GIT_FALLBACK_REMOTE_NAME` is valid, but is there any harm in leaving it there?

Agree'd re: `command`, I'll fix that.

wrboyce: > Also, I'd use `command` for external commands to avoid hitting an alias of function.

Fixed

Syphdias: A little thought experiment: You get a repo without `(package|composer).json` without an origin remote. Will you use this option? My opinion is, you won't because it will break all other repos and just rename your remote.

Issue #1134, opened 2019-01-10
Terminal printing lots of "Operation not permitted" errors regarding certain directories.


Nobody912:
#### Describe Your Issue
Terminal Prints:
> warning: could not open directory 'Library/[Insert Directories]': Operation not permitted

[See it in action:](https://gyazo.com/8e2d3018a5dfb7a89a2a694b2bcb56f3)

#### Have you tried to debug or fix it?
Reinstalled ZSH, Oh-My-ZSH, and PowerLevel9k multiple times. Not sure of what caused the error, I'm relatively new to all of this.

#### Environment Information
OS: **macOS 10.14.2 (18C54)**
System: **MacBook Pro (Retina, 13-inch, Early 2015)**
ZSH Version: **zsh 5.6.2 (x86_64-apple-darwin18.2.0)**
ZSH Framework: **Oh-My-ZSH**
Installation Method: **Repo**
P9K Version: **Release v0.6.6 (Not sure, probably latest)**
Terminal Emulator: **iTerm2**

Nobody912:
#### **echo $GIT_DIR**
`echo $GIT_DIR` only indents a line. 

#### **Dotfiles**
[https://imgur.com/a/r7HRxB2](https://imgur.com/a/r7HRxB2)

#### **~/.zshrc**
[https://pastebin.com/jAPJu37x](https://pastebin.com/jAPJu37x)

dritter:
Okay. `GIT_DIR` is not set. But it is still a Git Checkout. My best guess is that you have some symlinks lying around in your `Library` Folder. That symlinks seem to point somewhere you don't have access to.
Could you try to put the `Library` Folder in your `.gitignore`?

Nobody912:
Tried and put ```Library``` in the .gitignore. fixed the issue but now terminal load time is ungodly slow. I have some debugging work to do.

dritter:
@88maomao Well. Maybe. I would ask why there is a file/folder you don't have access to in your repo.

@Nobody912 You could have a look at our [measurings](https://github.com/bhilburn/powerlevel9k/pull/1046#issuecomment-435549640) [we](https://github.com/bhilburn/powerlevel9k/issues/1105#issuecomment-443549388) [did](https://github.com/bhilburn/powerlevel9k/issues/1054#issue-377635256). Basically, you could check the commands that are executed in the vcs segment, and see if they are slow itself.
[...]

PR #1157, opened 2019-02-06
Hide stderr from git ls-files

88maomao:
#### [Bugfix] Hide stderr from git ls-files
 
Hide `warning: could not open directory 'blabla/': Permission denied` if there are files/folders inside a git repo with no user access.

Fixes https://github.com/bhilburn/powerlevel9k/issues/1134

#### Questions
Should there be unittests for that?


dritter:
I don't think we need tests for this. Thanks @88maomao and welcome to P9K! :)

PR #1162, opened 2019-02-08
fix(p9k::defined): work when -u is set

docwhat:
Without this fix, `set -u` will cause commands to fail because of the
prompt.


docwhat:
Maybe the tests should all have `set -u` set in them?  I haven't looked into the tests in detail to see how to do that correctly.

dritter:
Having `set -u` in all test is probably a good idea.. But my gut feeling tells me that we have a lot of unused variables... IMHO we should address them in separate PRs..

PR #1176, opened 2019-03-03
Add Async Functionality per Segment

dritter:
This is my fourth iteration on async functionality and adds such (currently based solely on ZSH-Async) to every segment that is tagged as `async`.

#### Preconditions
This branch is based on the `segment_tags` branch (#1171 ).

#### Caveat
This is still in a pretty raw stage. So don't expect everything to be working. I drafted this at an early stage, to get feedback as early as possible.

#### Todo
These things are still open:
- [ ] Avoid warning, if no segment is tagged as async, and therefore ZSH-Async is not loaded. This happens because `async_flush_jobs` and `async_worker_eval` are always called.
- [ ] Support other async methods. See https://github.com/agkozak/agkozak-zsh-prompt#asynchronous-methods
- [ ] Avoid old prompts being eaten up. This happens on my machine when a segment is async, and I just press enter.
- [ ] Make path to ZSH-Async configurable.
- [ ] Fix tests
- [ ] Fix `vi_mode` segment. This contains old code based on `generators/async.p9k`..
- [ ] Improve code structure. The clarity of the functions in `generator/default.p9k` could be improved..
- [ ] Every segment must report their metadata to the P9K engine. This is important as this example illustrates: If a segment is conditional and that condition is met if we enter a special directory, the cached data of the segment must be reset, if we leave the special directory.
So `p9k::prepare_segment` must always be called (we have to look on conditions enclosing `p9k::prepare_segment` and early `return` statements). This is already done for most of the segments, but not for all:
- - [ ] battery
- - [ ] nvm
- - [ ] rspec_stats
- - [ ] symfony2_tests

### DIFF
--- a/generator/default.p9k
+++ b/generator/default.p9k
@@ -284,20 +284,21 @@ function __p9k_prompt_custom() {
 function __p9k_build_left_prompt() {
   [[ ${P9K_LEFT_PROMPT_ELEMENTS[1]} == "" ]] && return
   local index=1
-  local element joined
-  for element in "${P9K_LEFT_PROMPT_ELEMENTS[@]}"; do
+  local raw_segment joined
+  for raw_segment in "${P9K_LEFT_PROMPT_ELEMENTS[@]}"; do
+    local -a segment_meta
+    # Split by double-colon
+    segment_meta=(${(s.::.)raw_segment})
+    local segment="$segment_meta[1]"
     # Check if segment should be joined
-    [[ "${element[-7,-1]}" == '_joined' ]] && joined=true || joined=false
-
-    # Remove joined information in direct calls
-    element="${element%_joined}"
+    [[ ${segment_meta[(r)joined]} == "joined" ]] && joined=true || joined=false
 
     # Check if it is a custom command, otherwise interpet it as
     # a prompt.
-    if [[ $element[0,7] =~ "custom_" ]]; then
-      "__p9k_prompt_custom" "left" "$index" ${joined} $element[8,-1]
+    if [[ ${segment_meta[(r)custom]} == "custom" ]]; then
+      "__p9k_prompt_custom" "left" "$index" ${joined} $segment
     else
-      "prompt_$element" "left" "$index" $joined
+      "prompt_${segment}" "left" "$index" $joined
     fi
 
     index=$((index + 1))
@@ -316,19 +317,20 @@ function __p9k_build_left_prompt() {
 function __p9k_build_right_prompt() {
   [[ ${P9K_RIGHT_PROMPT_ELEMENTS[1]:-} == "" ]] && return
   local index=1
-  local element joined
-  for element in "${P9K_RIGHT_PROMPT_ELEMENTS[@]}"; do
+  local raw_segment joined
+  for raw_segment in "${P9K_RIGHT_PROMPT_ELEMENTS[@]}"; do
+    local -a segment_meta
+    # Split by double-colon
+    segment_meta=(${(s.::.)raw_segment})
+    local segment="$segment_meta[1]"
     # Check if segment should be joined
-    [[ "${element[-7,-1]}" == '_joined' ]] && joined=true || joined=false
-
-    # Remove joined information in direct calls
-    element="${element%_joined}"
+    [[ ${segment_meta[(r)joined]} == "joined" ]] && joined=true || joined=false
 
     # Check if it is a custom command, otherwise interpet it as a prompt.
-    if [[ $element[0,7] =~ "custom_" ]]; then
-      "__p9k_prompt_custom" "right" "$index" ${joined} $element[8,-1]
+    if [[ ${segment_meta[(r)custom]} == "custom" ]]; then
+      "__p9k_prompt_custom" "right" "$index" ${joined} $segment
     else
-      "prompt_$element" "right" "$index" $joined
+      "prompt_${segment}" "right" "$index" $joined
     fi
 
     index=$((index + 1))
--- a/generator/default.p9k
+++ b/generator/default.p9k
@@ -482,6 +482,8 @@ function __p9k_prepare_prompts() {
 
   # stop any running async jobs
   async_flush_jobs "__p9k_async_worker"
+  # Update the current working directory of the async worker.
+  async_worker_eval "__p9k_async_worker" builtin cd -q $PWD
 
   __p9k_build_segment_cache "left"
   __p9k_build_segment_cache "right"
--- a/generator/default.p9k
+++ b/generator/default.p9k
@@ -291,11 +291,11 @@ function __p9k_build_left_prompt() {
     segment_meta=(${(s.::.)raw_segment})
     local segment="$segment_meta[1]"
     # Check if segment should be joined
-    [[ ${segment_meta[(r)joined]} == "joined" ]] && joined=true || joined=false
+    [[ "${segment_meta[(ie)joined]}" -le "${#segment_meta}" ]] && joined=true || joined=false
 
     # Check if it is a custom command, otherwise interpet it as
     # a prompt.
-    if [[ ${segment_meta[(r)custom]} == "custom" ]]; then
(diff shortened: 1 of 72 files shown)


segments/anaconda/anaconda.spec
```diff
@@ -13,7 +13,7 @@ function setUp() {
 function testAnacondaSegmentPrintsNothingIfNoAnacondaPathIsSet() {
   local P9K_CUSTOM_WORLD='echo world'
```
onaforeignshore: What happens to these `P9K_CUSTOM_function` declarations now that you are using tags?

segments/disk_usage/disk_usage.p9k
```diff
@@ -45,17 +45,18 @@ prompt_disk_usage() {
         current_state='critical'
     fi
   else
+    current_state='normal'
     if [[ "$P9K_DISK_USAGE_ONLY_WARNING" == true ]]; then
         current_state=''
-        return
+        # Avoid printing the prompty, by setting $disk_usage
```
onaforeignshore: Spelling: prompt

dritter: Prompty prompt. 😄 Fixed. Thx.

segments/anaconda/anaconda.spec
dritter: Ideally nothing. The tags is just syntactic sugar. Instead of writing `custom_segment` you now write `segment::custom`. But the function still remains `P9K_CUSTOM_SEGMENT='echo segment'`..
Have to verify that custom segments still work though.. 🙄

Quentin-M:
This breaks `status`. Reverting `stauts.p9k` to its previous commit (def5835e95ef106162a8443232f643e2317b4cf7) fixes it.

dritter:
@Quentin-M Thanks for letting me know! I'll have a look at the `status` segment in the next days.

laggardkernel:
The last prompt may be removed if too many async segments are enabled and newlines are added by

```zsh
P9K_PROMPT_ADD_NEWLINE='true'
P9K_PROMPT_ON_NEWLINE='true'
```

In my test, I enabled as many async segments as possible, and entered a directory supporting many segment features. It turned out the last prompt may be removed/overwritten by the new prompt once you press `<Enter>` continuously.

Is there a competition among prompt redrawing activated by async segments?

generator/default.p9k
```diff
@@ -280,68 +313,202 @@ function __p9k_prompt_custom() {
...
+  # For security $PROMPT is never set directly. This way the prompt render is
+  # forced to evaluate the variable and the contents of $__p9k_unsafe[left]
+  # are never executed. The same applies to $RPROMPT.
+  PROMPT='${__p9k_unsafe[left]}'
 
-  echo -n "${__P9K_DATA[LAST_WHITESPACE]}${last_symbol}"
+  # About .reset-promt see:
+  # https://github.com/sorin-ionescu/prezto/issues/1026
+  # https://github.com/zsh-users/zsh-autosuggestions/issues/107#issuecomment-183824034
+  [[ "${1}" == "true" ]] && zle .reset-prompt
```
laggardkernel: `zle -R` (force redisplay) should be added at the end. Otherwise, the current prompt may be moved one line up and overwrite previous prompt.

```zsh
  [[ "${1}" == "true" ]] && zle .reset-prompt && zle -R
```

dritter:
Good findings @laggardkernel . I'll have a look at that in the next couple of days.

laggardkernel:
It turns out the async worker should be restarted in `precmd` to refresh the shell environment.

https://github.com/robobenklein/zinc/blob/28bea2b33e8fe1fc1319eab3c034f81bec8601b5/zinc_functions/prompt_zinc_setup#L102-L109

```zsh
# restarts just the worker - in order to update worker with current shell values
function zinc_worker_restart () {
  # _ZINC_DBG_OUT "restarting worker..."
  # async_worker_eval zinc_segment_worker "zinc_reload"
  async_stop_worker zinc_segment_worker
  async_start_worker zinc_segment_worker
  async_register_callback zinc_segment_worker _zinc_segment_async_callback
}
```

dritter:
@laggardkernel Thanks for pointing me to this. I changed the code, so that the worker gets restarted in the precmd hook.

It still seems odd to me, that we have to restart the whole worker just to update the variables. Maybe a better approach would be to use `async_worker_eval` to update the necessary variables inside the worker.

laggardkernel:
@dritter I think one of the benefits of this "conditional-async" is to make the segment functions work without any change of the code. Anyone could continue to contribute to the sections without realizing the existence of the async worker. 
Using `async_worker_eval` to update the necessary variables means that the contributors should take care of these variables, update the needed variables, and unset deprecated vars in time. Handing the job to the contributors may contradict the purpose above.

generator/default.p9k
```diff
@@ -280,68 +313,202 @@ function __p9k_prompt_custom() {
...
+  PROMPT=''
+  RPROMPT=''
+  # __p9k_unsafe must be a global variable, because we set
+  # PROMPT='$__p9k_unsafe[left]', so without letting ZSH
+  # expand this value (single quotes). This is a workaround
+  # to avoid double expansion of the contents of the PROMPT.
+  typeset -gAh __p9k_unsafe=()
+
+  # Process Cache
+  for data in "${(Oa@v)__p9k_segment_cache}"; do
```
laggardkernel: Another finding.

The use of expansion flag `(Oa@v)` couldn't ensure an iteration based on the array index.

Here's what I find with setting many segments on the left prompt.

```zsh
# example conf
# 12 segments on the left, and 0 on the right
P9K_LEFT_PROMPT_ELEMENTS=(context dir vcs::async node_version::async nvm::async rbenv::async anaconda::async pyenv::async status command_execution_time background_jobs time)

P9K_RIGHT_PROMPT_ELEMENTS=()
```

```zsh 
# result
for data in "${(Oa@v)__p9k_segment_cache}"; do echo "$data"; done
CONTEXT_DEFAULT·|·left·|·1·|·false·|··|··|·false
DIR_HOME_SUBFOLDER·|·left·|·2·|·false·|·…/dritter---powerlevel9k·|··|·true
VCS_CLEAN·|·left·|·3·|·false·|· conditional_async·|· ·|·true
NODE_VERSION·|·left·|·4·|·false·|··|··|·false
NVM·|·left·|·5·|·false·|··|··|·false
COMMAND_EXECUTION_TIME·|·left·|·10·|·false·|··|··|·false
RBENV·|·left·|·6·|·false·|·2.6.3·|··|·false
BACKGROUND_JOBS·|·left·|·11·|·false·|·0·|··|·false
ANACONDA·|·left·|·7·|·false·|·()·|··|·false
TIME·|·left·|·12·|·false·|·%D{%H:%M:%S}·|··|·true
PYENV·|·left·|·8·|·false·|·3.6.8:2.7.16·|··|·false
STATUS_OK·|·left·|·9·|·false·|··|· ·|·true
```

The sequence of `BACKGROUND_JOBS` is not based on its index from array `P9K_LEFT_PROMPT_ELEMENTS`.

laggardkernel: A pr is opened for this problem in your fork. https://github.com/dritter/powerlevel9k/pull/41

laggardkernel:
The current implementation requires the segment functions pass out an additional conditional string to decide whether render the segment or not. Another function `p9k::segment_no_print` is added to handler early exit.

I have some idea to avoid this and **make the async work with segment functions untouched**. I'd like to make a pr to achieve this once you make a decision on my current pr.

dritter:
> Using async_worker_eval to update the necessary variables means that the contributors should take care of these variables, update the needed variables, and unset deprecated vars in time. Handing the job to the contributors may contradict the purpose above.

Btw. @laggardkernel I was thinking of a general approach to transport the current env into the worker (might use a temp file for transport, like [here](https://github.com/dritter/powerlevel9k/blob/4c0ed08c94e90d1d50026de90b458f607374ff93/powerlevel9k.zsh-theme#L1486-L1507)). This could be done in the precmd hook, so you don't have to worry about that when writing new segments.

PR #1178, opened 2019-03-05
Add Performance Issue Template

dritter:
This adds a new Issue Template specifically for Performance Issues.


Syphdias:
Looks good to me. Do you think it would be helpful to mention `zprof` or is it overkill?

dritter:
Yes, I think `zprof` would be too much.. Thx for the review.

PR #1181, opened 2019-03-06
[Enhancement] Only show rust version status when in a rust project directory or subdirectory

loganintech:
#### Description
 

Similar to how the go information only displays when in a go source directory, rust info should only show up when in a rust project.
This leverages Rust's package manager Cargo. Specifically, it runs [cargo pkgid](https://doc.rust-lang.org/cargo/commands/cargo-pkgid.html) which would normally print the fully qualified package information of the current working directory's project. However, when not in a rust project and none of the parent directories contain `Cargo.toml` it prints `error:`. This simply checks if error was printed by Cargo and assumes we're not in a rust project because of it. 

#### Questions
1. I will need assistance with unit tests.
2. I'm not really sure if this is the right branch to merge onto (master to master) but it's a pretty small change
3. I'd like input on whether or not using Cargo as an underlying driver for this will work for the majority of people. Cargo is one of the best package managers I've used and most if not all rust users I know take advantage of it, meaning it's on their machines. However, this requires an outside tool and I'd understand if that would want to be avoided.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -1321,13 +1321,19 @@ prompt_root_indicator() {
 # Segment to display Rust version number
 prompt_rust_version() {
   local rust_version
+  local is_rust_dir
   rust_version=$(command rustc --version 2>/dev/null)
   # Remove "rustc " (including the whitespace) from the beginning
   # of the version string and remove everything after the next
   # whitespace. This way we'll end up with only the version.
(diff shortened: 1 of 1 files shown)


loganintech:
@dritter what do you think about this?

evanjs:
Excited to see this happen.  While at the moment, it feels like having a date for the installed nightly version would make the status indicator more useful, it'd be great to have it only show up in Rust projects to start.

PR #1184, opened 2019-03-09
Fix: bad math expression

paleite:
When `grep` is set to `--color=always` (e.g. by specifying `alias grep="grep --color=always"`), the script didn't take them into account. This is likely the cause of https://github.com/bhilburn/powerlevel9k/issues/1092.

This PR changes all instances of `grep` (and `egrep`) to use `--color=never`.


sambadevi:
Thanks for your PR @paleite ! I've added your fix to my PR #1198 because I am using `egrep` there as well!

paleite:
That’s awesome, @sambadevi 👍🙂

PR #1188, opened 2019-03-12
[Docs] remove reference to POWERLEVEL9K_RBENV_ALWAYS

ChrisBaker97:
I noticed that the environment variable POWERLEVEL9K_RBENV_ALWAYS doesn't seem to do anything or to be referenced anywhere in the source outside of README.md. Seems like it might have been an anachronism left hanging when the variable was renamed to POWERLEVEL9K_RBENV_PROMPT_ALWAYS_SHOW?


dritter:
Thanks @ChrisBaker97 .

PR #1191, opened 2019-03-13
[New Segment]: pyver

sambadevi:
As already mentioned in #821 I updated my code and extracted the functionality into its own segment.

Similar to pyenv, the pyver segment allows for P9K_PYVER_PROMPT_ALWAYS_SHOW to be set to always display the segment even if the global python version is the same as the python version in use.


sambadevi:
Oh I forgot to mention, I also added `.DS-Store` files too `.gitignore`

PR #1119, opened 2018-12-15
[Bugfix] Attempt to fix broken ctrl-c behavior

Syphdias:
#### Description
This PR tries to fix the bug mentioned in #812 and #1027. If you hold `crtl-c` the prompt build stops before evaluating the functions in `$PROMPT` and leaves the raw strings behind.
I approach this by evaluating the functions inside `$PROMPT`, i.e. `__p9k_build_left_prompt`, in function `__p9k_prepare_prompt` and writing it to a newly introduced global variable `$__p9k_unsafe_PROMPT` (this is probably against the style guide since I wrote part of the varibale in capital letters. I will change it if requested, of course).  
The unsafe variable is then put into `$PROMPT` with _single_ quotes: `PROMPT='$__p9k_unsafe_PROMPT'`. This is done to prevent evil code to be executed if the contents of the unsafe variable is tainted after evaluating __p9k_build_left_prompt. The same is done for `$RPROMPT`

The benefit is, that since the function `__p9k_prepare_promt` is run as a precmd hook, it can be interrupted without leaving `$PROMPT` unchanged/raw. This also means it could be out of date, but this is still better than no prompt.

#### Question/Problem
Currently I'm having trouble running tests since I get lots of errors regarding background_jobs on stderr. Example:
```zsh
function setUp() {
  export TERM="xterm-256color"
  source powerlevel9k.zsh-theme
}
function testFoo() {
  cd /
}
source shunit2/shunit2
```
Running this test yields
```
 ❱ zsh foo.spec                             
testFoo
prompt_background_jobs:2: bad math expression: operand expected at end of string

Ran 1 test.

OK
```
I think this is the important part (run with `set -xv` right before the `cd /`)
```zsh
+__p9k_build_right_prompt:19> index=3 
+__p9k_build_right_prompt:4> element=background_jobs
+__p9k_build_right_prompt:6> [[ nd_jobs == _joined ]]
+__p9k_build_right_prompt:6> joined=false 
+__p9k_build_right_prompt:9> element=background_jobs 
+__p9k_build_right_prompt:13> [[ $element[0,7] -regex-match "custom_" ]]
+__p9k_build_right_prompt:16> prompt_background_jobs right 3 false
+prompt_background_jobs:1> local jobs_print=''
prompt_background_jobs:2: bad math expression: operand expected at end of string
[...]

Issue #1190, opened 2019-03-13
[Feature request] An option to turn `time` segment into realtime clock


romkatv:
It would be nice to have an option that would turn `time` segment into realtime clock. That is, the segment would look the same as before but it would constantly update and always display current time. Users are asking for it. This feature exists in several other themes.

I implemented it in Powerlevel10k and it looks cool. However, I'd very much like to avoid having visual features that don't exist in Powerlevel9k because it'll make it difficult for the users to switch back to Powerlevel9k if/when it becomes faster. Hence I named the option `POWERLEVEL9K_EXPERIMENTAL_TIME_REALTIME` with `EXPERIMENTAL` in the name. If you could add `POWERLEVEL9K_TIME_REALTIME` (or use some other name that isn't as obscure), I'll rename my option to match.

bhilburn:
@romkatv - That's really cool! I would love to upstream this.

I'd like to get this in as a new clock segment. I actually prefer _not_ having the clock update in my personal prompt, as it's easy for me to look at that segment to see when the previous command completed rather than scrolling up to when I executed it and back-calculating based on the time expired segment.

@romkatv - Can you file a PR against `next` that adds this feature as a new `clock` segment, perhaps?

romkatv:
> I'd like to get this in as a new clock segment.

The advantage of a separate segment is that it allows both time segments to be visible simultaneously -- the time of the last command completion and the time of the next command start (which is ticking at the current prompt). Both are useful, but in practice it's better to use one kind of `time` segment plus `command_execution_time`. The other `time` can be computed from these.

When I scroll up in my terminal, I personally prefer to see when each command started (plus its duration) rather than when each has ended. It feels more intuitive. Plus, it makes prompt look cooler, even more eye-candy. But I can see how one might prefer to see end times, mostly due to habbit.

The disadvantage of having two time segments is UX complexity. Note that these segments will have identical set of configuration options (`time` has only one option currently).

> @romkatv - Can you file a PR against next that adds this feature as a new clock segment, perhaps?

I'll defer to your dev team. Feel free to look at my implementation or just roll your own. One word of caution: don't use `TMOUT` to implement this because there can be only one `TMOUT` handler in a shell. It's not nice to claim a singleton resource and deny its use to everyone else. Powerlevel9k already does this with `zle-keymap-select`, and it's also not nice.

Syphdias:
I'd like to take a look at this. A few things.

[...]

PR #1193, opened 2019-03-14
[Enhancement] Adds option to periodically update time segment

Syphdias:
This implements #1190

Hey @romkatv,
porting your code was easy. Thank you! I do have some questions though.
1. You mentioned big git repos which would take time. I don't think this is the case on `next`. Please correct my if you think I'm wrong. `PROMPT` gets evaluated every time you redraw the prompt. On `master` `PROMPT` is something like `%f%b%k$(__p9k_build_left_prompt)`, on `next` it is `__p9k_unsafe_PROMPT`. This variable get filled in the precmd hook which is not called on `prompt-reset`. This was done to fix #1119 another problem but come is very handy.
2. Any ideas for how to test this?
3. I ran into some issues. Which lead to the notice in the README and a default delay of 60 seconds.
The issues are with going through the history and tab completion/suggestions. Example for weirdness:
This is how it should look:
image
This is what happens:
image
~~I even had a segfault once doing this.~~ I can reproduce the segfault by `ls <tab><arrow-keys-especially-up>`
4. Changing keybinds
You mentioned that `zle-keymap-select` was a rude practice in #1190(?) Would you say the same about an option that would do this?
```zsh
renew-prompt-accept-line() {
    zle .reset-prompt
    zle accept-line
}
zle -N renew-prompt-accept-line
bindkey "^M" renew-prompt-accept-line
ZSH_AUTOSUGGEST_CLEAR_WIDGETS+=renew-prompt-accept-line
```

### DIFF
--- a/segments/time/README.md
+++ b/segments/time/README.md
@@ -13,6 +13,8 @@ where you want to show this segment.
 | Variable | Default Value | Description |
 |----------|---------------|-------------|
 |`P9K_TIME_FORMAT`|`'H:M:S'`|ZSH time format to use in this segment.|
+|`P9K_TIME_REALTIME`|`false`|Enabling this option will update your prompt every `P9K_TIME_REALTIME_DELAY` seconds to display the current time. Note: This will trigger a `.reset-prompt` and can lead to side effects like history getting "stuck" or suggestions disappearing. |
+|`P9K_TIME_REALTIME_DELAY`|`60`|This only takes effect if `P9K_TIME_REALTIME` is `true` and is used to set the delay between updates.|
 
 As an example, if you wanted a reversed time format, you would use this:
 ```zsh
--- a/segments/time/time.p9k
+++ b/segments/time/time.p9k
@@ -13,6 +13,50 @@
   #   segment_name  context  background  foreground  Generic  Flat/Awesome-Patched  Awesome-FontConfig  Awesome-Mapped-FontConfig  NerdFont
   #                                                                                                            
   p9k::register_segment "TIME" "" "${DEFAULT_COLOR_INVERTED}" "${DEFAULT_COLOR}"  ''  $'\uE12E'  $'\uF017 '  $'\uF017 '  $'\uF017 '
+
+  ################################################################
+  # Register segment default values
+  p9k::set_default P9K_TIME_FORMAT "%D{%H:%M:%S}"
+  p9k::set_default P9K_TIME_REALTIME false
+  p9k::set_default P9K_TIME_REALTIME_DELAY 60
+
+  ################################################################
+  # Register handler for realtime prompt resets
+  if [[ $P9K_TIME_REALTIME == true ]]; then
+    local fifo
+    fifo=$(mktemp -u "${TMPDIR:-/tmp}"/p9k.$$.pipe.time.XXXXXXXXXX)
+    typeset -giH __P9K_TIMER_FD=0
+    __p9k_start_timer() {
+      emulate -L zsh
+      setopt err_return
+      unsetopt bg_nice
+      mkfifo $fifo
+      exec {__P9K_TIMER_FD}<>$fifo
+      typeset -gfH __p9k_on_timer() {
+        # reset prompt on pipe update
+        emulate -L zsh
+        local _
+        IFS='' read -u $__P9K_TIMER_FD _ && zle && zle .reset-prompt
+      }
+      # register handler for created name pipe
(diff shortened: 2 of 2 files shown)


Syphdias:
@romkatv wtf? How did you comment before I even created the PR...
Well, I think I answer your question in 1

romkatv:
> [...] This was done to fix #1119 another problem but come is very handy.

Interesting. So does it mean `background_jobs` broke? Just checked -- yes, it is broken in `next`. I opened https://github.com/bhilburn/powerlevel9k/issues/1196.

> 2. Any ideas for how to test this?

None.

> 3. I ran into some issues...

Ha-ha, this is funny. I mailed https://www.zsh.org/mla/workers//2019/msg00161.html. ZSH seems kinda buggy. This is the 4th bug I've bumped into and I haven't been using ZSH even for a month.

> 4. Changing keybinds
>    You mentioned that `zle-keymap-select` was a rude practice in #1190(?) Would you say the same about an option that would do this?
> 
> ```shell
> renew-prompt-accept-line() {
>     zle .reset-prompt
>     zle accept-line
> }
> zle -N renew-prompt-accept-line
> bindkey "^M" renew-prompt-accept-line
> ZSH_AUTOSUGGEST_CLEAR_WIDGETS+=renew-prompt-accept-line
> ```

Is this related? I'd say this code is fine in `.zshrc` and not fine anywhere else. If your plugin thinks it's OK to override `renew-prompt-accept-line`, it should be OK for other plugins too. And it obviously won't work.  Great post on this topic, highly recommended: https://devblogs.microsoft.com/oldnewthing/?p=35413.

Syphdias:
> Interesting. So does it mean `background_jobs` broke? Just checked -- yes, it is broken in `next`. I opened #1196.

Good catch! Interestingly I did run into trouble with background jobs but did not notice at the time.

> > 1. I ran into some issues...
> 
> Ha-ha, this is funny. I mailed https://www.zsh.org/mla/workers//2019/msg00161.html. ZSH seems kinda buggy. This is the 4th bug I've bumped into and I haven't been using ZSH even for a month.

😆  


> Is this related? I'd say this code is fine in `.zshrc` and not fine anywhere else. If your plugin thinks it's OK to override `renew-prompt-accept-line`, it should be OK for other plugins too. And it obviously won't work. Great post on this topic, highly recommended: https://devblogs.microsoft.com/oldnewthing/?p=35413.

Actually, this is straight from my `.zshrc` I don't think I have overridden `renew-prompt-accept-line` since I named it. I should have gone with the p9k naming conventions in my example. But it does override the keybind for `^M` (Enter) from `accept-line` to the new function. Better example:
```zsh
__p9k_renew-prompt-accept-line() {
    zle .reset-prompt
    zle accept-line
}
zle -N __p9k_renew-prompt-accept-line
bindkey "^M" __p9k_renew-prompt-accept-line
ZSH_AUTOSUGGEST_CLEAR_WIDGETS+=__p9k_renew-prompt-accept-line
```
So I'm not asking for overriding the function but overriding the keybind.

Syphdias:
> Is this related? I'd say this code is fine in `.zshrc` and not fine anywhere else. If your plugin thinks it's OK to override `renew-prompt-accept-line`, it should be OK for other plugins too. And it obviously won't work. Great post on this topic, highly recommended: https://devblogs.microsoft.com/oldnewthing/?p=35413.

I don't quite get what you are playing at. If I'd change a bindkey setting twice it will simply be overridden there is no conflict or logical problem besides the fact that you might not want to to be overridden.
Just to clarify: I'd offer this as an option, not by default.

romkatv:
I'm not sure where the example with key binding is coming from. My comment was abut TMOUT. The user isn't explicitly asking TMOUT to be hogged, they just want thge clock to tick. Many plugins would like to claim TMOUT for themselves to implemented this and that but it's not cool because there is only one TMOUT. Likewise with zle overrides. The user isn't asking for zle-keymap-select to be overridden, they just want vi mode in their prompt. If you claim zle-keymap-select to implement vi_mode segment, then the user cannot use any other plugin that also wants to hook into zle-keymap-select.

This is a minor point. I suggest we drop this subject to save time for something more important.

Syphdias:
> This is a minor point. I suggest we drop this subject to save time for something more important.

Agreed. We are arguing about different things.

I still like the feature, but since it is a bit hacky and buggy, I'd like to have another option. I'll implement `__p9k_renew-prompt-accept-line` to see if it is feasible to achieve something similar.

PR #1194, opened 2019-03-14
[Enhancement] Updates to aws Segment to include region and delimiter

buzzsurfr:
#### Description

In the `aws` segment, optionally include the region with the profile and specify a delimiter to use between the two. Will pull region from environment variables or awscli configuration.

#### Questions

First PR here. Let me know if I need to do anything else.

I also could use some help to change the segment graphic to show the region.

### DIFF
--- a/segments/aws/README.md
+++ b/segments/aws/README.md
@@ -11,13 +11,17 @@ where you want to show this segment.
 ## Configuration
 
 If you would like to display the [current AWS
-profile](http://docs.aws.amazon.com/cli/latest/userguide/installing.html), add
-the `aws` segment to one of the prompts, and define `AWS_DEFAULT_PROFILE` in
-your `~/.zshrc`:
+profile and region](http://docs.aws.amazon.com/cli/latest/userguide/installing.html), add
+the `aws` segment to one of the prompts, and define `P9K_AWS_DEFAULT_PROFILE` 
+(and optionally `AWS_DEFAULT_REGION`) in your `~/.zshrc`:
 
 | Variable | Default Value | Description |
 |----------|---------------|-------------|
-|`AWS_DEFAULT_PROFILE`|None|Your AWS profile name|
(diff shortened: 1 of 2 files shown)


sambadevi:
Hey @buzzsurfr thanks for your PR!
First off, good work on the `aws` segment update!
My suggestion would be to set a default for `P9K_AWS_DELIMITER` in your segment initialization (between Line 15 and 16 for example)
```
# default delimiter set to "|"  as example
  p9k::set_default P9K_AWS_DELIMITER "|"
```
That way you could also `unset P9K_AWS_DELIMITER` if no region is set or change the complete output string **if** region is set.

Either way, +1 on this PR!

buzzsurfr:
@sambadevi 

I can add that line in a new commit, but does that mean that I don't need to set the delimiter default here:
https://github.com/buzzsurfr/powerlevel9k/blob/d8a0415b7759e804f666ac6d10bc33ed665fae48/segments/aws/aws.p9k#L44
...and instead can just use the variable name? (Swap `${P9K_AWS_DELIMITER:-/}` for `${P9K_AWS_DELIMITER}`)

sambadevi:
That’s correct, since your variable is already set to a default you can swap `${P9K_AWS_DELIMITER:-/}` for `${P9K_AWS_DELIMITER}`.

This also means that you should be able to remove the the two following lines (else, set local aws_delim....). Since the global delimiter variable is set by default to something, the local delimiter variable only gets set if an aws_region is set.

So in case of no Region `aws_delimiter` is unset and your final segment output would be just the profile

buzzsurfr:
I updated the code to use `p9k::set_default`, but during my testing, if I run...
```
unset P9K_AWS_DELIMITER
```
...then it removes the delimiter altogether, instead of using the default. Thus, I left in the `:-|` for now on (now) line 45, but can remove it if you figure out what I'm missing :-)

PR #1198, opened 2019-03-14
[Enhancement]: add p9k::reset_prompt function

### LINKED ISSUE #1172: Simple way of clearing p9k variables
I know of `prompt_powerlevel9k_teardown` in `next` branch, however master doesn't have this function. Also I'm not sure if the teardown is the right way to do this.

What I'm trying to achieve:
I have multiple configuration (theme-files) for p9k which I want to switch to, but if I have a variable set in theme a and I don't overwrite this variable in theme b, unexpected things might happen.

Example
Theme a sets `POWERLEVEL9K_SHORTEN_DIR_LENGTH=5` (for whatever reason) and theme b doesn't touch the variable and assumes the default values from p9k, `POWERLEVEL9K_SHORTEN_DIR_LENGTH` would still be 5 after the switch

sambadevi:
This function clears every variable from p9k (`egrep -i "^p9k_"`, it's case insensitive) and reloads the default p9k settings (taken from a clean useraccount with only p9k installed and no configuration)

closes #1172


onaforeignshore:
The only problem I see with this is if a segment is updated with a change to the defaults and the person forgets to make the same change to the `defaults.zsh` file.

sambadevi:
This should probably be added to the developers guide then?

> If a segment defines defaults for a variable, please add the default definition to `functions/utilities.zsh`

PR #1214, opened 2019-03-17
[Enhancement] Refactor command_execution_time.p9k segment

ChrisBaker97:
I wanted to add an "s" after the seconds display, in the case that the execution time was under a minute, to make it clearer what units were being displayed, when there were no colons. 

In the process of investigating this modification, I noticed that the code for formatting either whole or partial seconds (depending on the `P9K_COMMAND_EXECUTION_TIME_PRECISION` variable) seemed unnecessarily complex. By eliminating the pre-declaration of the number type and replacing it with a `printf` statement to format on the fly, I was able to squash a three-way `if`/`elif`/`else` control structure into a single line.

I also noticed that all this formatting was being done every time, regardless of whether the execution time actually exceeded the threshold, so I moved that logic to the front of the module, so that none of those computations happen if nothing will be displayed anyway.

### DIFF
--- a/segments/command_execution_time/command_execution_time.p9k
+++ b/segments/command_execution_time/command_execution_time.p9k
@@ -30,22 +30,21 @@
 #   $3 boolean Whether the segment should be joined
 ##
 prompt_command_execution_time() {
-  # Print time in human readable format
-  # For that use `strftime` and convert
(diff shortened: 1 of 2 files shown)


dritter:
Thanks @ChrisBaker97 and apologies for the delay.

PR #1217, opened 2019-03-17
[Bugfix] Right-side prompt segments with an icon but no content double-print whitespace

ChrisBaker97:
When generating a right-side prompt segment that has an icon but no other content (like `status` in the `OK` state), whitespace is inserted both before and after this empty content, resulting in an extended prompt segment. When there is content, it looks fine, because there should be whitespace on both sides of the text:

image

(Showing `status` segment on both left and right for comparison.)

This fix modifies the logic in `default.p9k` for right-side prompt generation so that a second whitespace character is not inserted in the event of an empty `${content}` string. However, it leaves the previous bracketing whitespace intact if content exists:

image

### DIFF
--- a/generator/default.p9k
+++ b/generator/default.p9k
@@ -182,8 +182,9 @@ function __p9k_right_prompt_segment() {
       || visual_identifier="${visual_identifier}${middle_ws}"
   fi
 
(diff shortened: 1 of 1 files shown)


dritter:
Thanks @ChrisBaker97

PR #1220, opened 2019-03-19
[Bugfix] Hide kubecontext prompt if there is no current context

tdmalone:
This hides the kubecontext prompt if there is no current context set in `~/.kube/config`. Essentially, this prevents a namespace from being shown all on its own.

**BEFORE this PR - no context set:**
image

**AFTER this PR - no context set:**
image

**BEFORE & AFTER this PR (i.e. no change) - context set:**
image


mvaldesdeleon:
Would be nice if this got merged, as I just implemented the same feature and came here looking for unmerged PRs before submitting my own.

PR #1221, opened 2019-03-19
[Bugfix] Don't unset global git user config after git tests

Syphdias:
Currently running the git unit tests unsets name and email in the global git config.

This PR gets rid of unnecessary variable juggling and simply sets the git config in the new home. The config (`/tmp/powerlevel9k-test/.gitconfig`) will be removed when the test directory is removed. I also added a "test" (which is a bit of a hack) to check if email or name were changed.

PS: Teardowns are currently called an additional time (see https://github.com/kward/shunit2/issues/112 for details) after shunit finished all tests. It should be fine for now. Just a heads up since I stumbled upon this bug while testing.

### DIFF
--- a/segments/vcs/vcs_git.spec
+++ b/segments/vcs/vcs_git.spec
@@ -5,6 +5,21 @@
 setopt shwordsplit
 SHUNIT_PARENT=$0
 
+function oneTimeSetUp() {
+  # must be done before any setup
+  ORIGINAL_GIT_AUTHOR_NAME="$(git config --global user.name)"
+  ORIGINAL_GIT_AUTHOR_EMAIL="$(git config --global user.email)"
+}
+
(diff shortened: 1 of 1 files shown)


Syphdias:
@dritter I'm confused. How is that not what is being done [here](https://github.com/bhilburn/powerlevel9k/pull/1221/files#diff-33d8cc39110005e14f88376a43b0450fR40)? Which is why I deleted most of the setting resets.

I also added the test to ensure the user config modification doesn't happen again. This apparently failed on Mac OS...

dritter:
Ha! I missed that line (folded away by github).

So I adjust my comment: I wouldn't trust `git config --global`, and instead just write the config file directly to our newly created HOME folder.. ;)

dritter:
Thanks @Syphdias .

PR #1227, opened 2019-03-27
Fix dir issues

### LINKED ISSUE #1206: Many bugs in dir segment
I went to optimize `prompt_dir` in Powerlevel10k and noticed that the code is crawling with bugs. Just from the cursory look, here are some.

Tested on `master` with default settings plus whatever is specified in each test case.

1.

```zsh
mkdir ~/'%E%K{red}'
cd ~/'%E%K{red}'
```

Expected content of dir segment: `~/%E%K{red}`.
Actual: color mayhem.

2.

```zsh
mkdir ~/~
cd ~/~
```

Expected content of dir segment: `~/~`.
Actual: `~`.

3.

```zsh
POWERLEVEL9K_DIR_PATH_HIGHLIGHT_BOLD=true POWERLEVEL9K_DIR_OMIT_FIRST_CHARACTER=true
mkdir ~/~
cd ~/~
```

Expected content of dir segment: `/~`.
Actual: `~`.

Note that with these settings we'll get `/foo` when we are in `~/foo`. I'm assuming here this is the expected behavior of `POWERLEVEL9K_DIR_OMIT_FIRST_CHARACTER` although I cannot imagine why anyone would want this. What's clear is that it should be consistent between `~/foo` and `~/~` but currently it isn't.

4.

```zsh
POWERLEVEL9K_HOME_FOLDER_ABBREVIATION=HOME
sudo mkdir /~
cd /~
```

Expected content of dir segment: `/~`.
Actual: `/HOME`.

dritter:
This fixes #1206 . At least on `next` branch. The original Issue aims at `master`.

Caveat: There are probably more bugs in the dir segment. That segment needs a major refactoring.

### DIFF
--- a/segments/dir/dir.p9k
+++ b/segments/dir/dir.p9k
@@ -207,37 +207,34 @@ prompt_dir() {
   [[ ${dir_state_foreground} == "%f" ]] && dir_state_foreground=${__P9K_DATA[DIR_${current_state}_FG]}
 
   local dir_name base_name
-  # use ZSH substitution to get the dirname and basename instead of calling external functions
-  dir_name=${path_opt%/*}
-  base_name=${path_opt##*/}
+  # Split current_path into an array again (*sigh*).
+  local -a dirs=(${(s:/:)path_opt})
+  # First part up to the second last
+  dir_name="${(j:/:)dirs[1, -2]}"
+  # Last part
+  base_name="${dirs[-1]}"
+  # Reset current_path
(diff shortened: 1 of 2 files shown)


Syphdias:
Looks good to me but the segment really could use some love.

dritter:
I'm merging this for now. We should schedule the refactoring for v0.8.0.. Thanks for the review @Syphdias

PR #1229, opened 2019-03-27
Add Gitstatus segment

### LINKED ISSUE #1185: Add gitstatus segment
As discussed in #1170 , we should add a `gitstatus` segment. This is faster than `VCS_INFO`, and could be included in 0.7.0 or 0.8.0. Just wanted to open this Issue that we don't forget.
If anyone wants to implement this, I'd be happy.

Pro/Con List of include this in the `vcs` segment:
:+1: Users don't have to worry about this
:+1:  No duplication of tests. Obviously tests have to be duplicated (or be very similar) if not included.
:-1:  `gitstatus` does not support all use-cases we have in the `vcs` segment (AFAIR; correct me if I'm wrong).
:-1:  The `vcs` segment code is quite lengthy already
:-1:  Cleaner readme: The `gitstatus` install instructions can be addressed cleanly
:-1:  Easier handling: If for some reason (like user forgot to install `gitstatus`), it would just not work, so the feedback loop is cleaner. Otherwise it would fallback to `VCS_INFO`, pretending to work

dritter:
This adds a segment for `gitstatus`. Fixes #1185 .

Caveat: This does not make use of the asynchronous mode that `gitstatus` provides. This will be added as soon as we have a proper basis for that (presumably with #1176 ).

// cc @romkatv


segments/gitstatus/gitstatus.p9k
```diff
@@ -0,0 +1,146 @@
...
+  local localBranch="${VCS_STATUS_LOCAL_BRANCH}"
+  local remoteBranch="${VCS_STATUS_REMOTE_BRANCH}"
+  [[ -n "$localBranch" ]] && segment_content+=("${__P9K_ICONS[GITSTATUS_BRANCH]}${localBranch//\%/%%}")
+  # If no local branch and no remote branch was detected, we are probably in detached mode.
+  [[ -z "$localBranch" && -z "$remoteBranch" ]] && segment_content+=("${__P9K_ICONS[GITSTATUS_BRANCH]}(detached)")
+
+  if [[ -n "${remoteBranch}" \
+      && ( \
+        "${P9K_GITSTATUS_ALWAYS_SHOW_REMOTE_BRANCH}" == 'true' \
+        || "${remoteBranch#*/}" != "${VCS_STATUS_LOCAL_BRANCH#heads/}" \
```
romkatv: things like heads/ are already stripped out from VCS_STATUS_LOCAL_BRANCH and VCS_STATUS_REMOTE_BRANCH

segments/gitstatus/gitstatus.p9k
```diff
@@ -0,0 +1,146 @@
...
+  # Register segment
+  # Parameters:
+  #   segment_name  context  background  foreground  Generic  Flat/Awesome-Patched  Awesome-FontConfig  Awesome-Mapped-FontConfig  NerdFont
+  p9k::register_segment  "GITSTATUS" "CLEAN"     "green"  "${DEFAULT_COLOR}" ''  ''  ''  ''  ''
+  p9k::register_segment  "GITSTATUS" "MODIFIED"  "yellow" "${DEFAULT_COLOR}" ''  ''  ''  ''  ''
+  p9k::register_segment  "GITSTATUS" "UNTRACKED" "green"  "${DEFAULT_COLOR}" '?' $'\uE16C' $'\uF059' '\u'${CODEPOINT_OF_AWESOME_QUESTION_CIRCLE} $'\uF059'
+
+  ################################################################
+  # Register segment default values
+  p9k::set_default P9K_GITSTATUS_MAX_INDEX_SIZE_DIRTY "4096"
```
romkatv: -1 would be a better default. The latest version if gitstatus (not submitted yet) returns data for chromium (300k files) in less than 75 ms. This is 5 times faster than `git status`. It'll feel fast even without async.

segments/gitstatus/gitstatus.p9k
```diff
@@ -0,0 +1,146 @@
...
+
+  if [[ -n "${remoteBranch}" \
+      && ( \
+        "${P9K_GITSTATUS_ALWAYS_SHOW_REMOTE_BRANCH}" == 'true' \
+        || "${remoteBranch#*/}" != "${VCS_STATUS_LOCAL_BRANCH#heads/}" \
+      ) ]]; then
+      segment_content+=("${__P9K_ICONS[GITSTATUS_REMOTE_BRANCH]}${remoteBranch//\%/%%}")
+  fi
+
+  (( VCS_STATUS_HAS_UNTRACKED )) && current_state="UNTRACKED" && segment_content+=("${__P9K_ICONS[GITSTATUS_UNTRACKED]}")
```
romkatv: VCS_STATUS_HAS_UNTRACKED and VCS_STATUS_HAS_UNSTAGED are tribools:

  0: no
  1: yes
  -1: unknown

-1 is returned if git index has more than P9K_GITSTATUS_MAX_INDEX_SIZE_DIRTY files

You probably want to handle 1 and -1 differently.

romkatv:
Looks great! Finally gitstatus will have tests :-D

segments/gitstatus/gitstatus.p9k
```diff
@@ -0,0 +1,146 @@
...
+# @args
+#   $1 string Alignment - left | right
+#   $2 integer Segment index
+#   $3 boolean Whether the segment should be joined
+##
+prompt_gitstatus() {
+  gitstatus_query -d ${${GIT_DIR:a}:-$PWD} -t $P9K_GITSTATUS_MAX_SYNC_LATENCY_SECONDS P9K 2>/dev/null
+
+  # Print nothing, if we are not in a repo, or if the gitstatus daemon sends a timeout
+  [[ -z "${VCS_STATUS_RESULT}" || "${VCS_STATUS_RESULT}" =~ "norepo-" || "${VCS_STATUS_RESULT}" == "tmout" ]] && return
```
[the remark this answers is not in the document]
dritter: You mean a negated check for `ok-*`? This is the early exit here..

dritter:
> Looks great! Finally gitstatus will have tests :-D

I don't want to disappoint you, I've mocked `gitstatus` away, so the tests run on travis.. ;)

segments/gitstatus/gitstatus.p9k
dritter: Good point! Btw. The `VCS_STATUS_REMOTE_BRANCH` looks a bit too much stripped, for my taste. If I track a foreign remote, it would be nice, if `gitstatus` would include the origin..

segments/gitstatus/gitstatus.p9k
dritter: Awesome! I'll change it to `-1`, as you are way faster than me anyway. ;)

segments/gitstatus/gitstatus.p9k
dritter: 👍

romkatv:
> I don't want to disappoint you, I've mocked `gitstatus` away, so the tests run on travis.. ;)

Oh, man! And just like, that my test coverage went back to zero.

segments/gitstatus/gitstatus.p9k
romkatv: How would you prefer it to look like? Could you give an example?

segments/gitstatus/gitstatus.p9k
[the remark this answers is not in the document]
romkatv: Yes, of course.

```zsh
[[ $VCS_STATUS_RESULT =~ ok-* ]] || return
```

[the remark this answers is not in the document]
dritter: Btw. that would work, except if gitstatus returns a timeout. I am a bit undecided, what I should do in that state, TBH. With the current implementation it shows (wrongly) a detached state (clarification: the segment code does this, not gitstatus), which is maybe better than showing nothing. On the other hand I could handle the timeout case differently (showing something like "timeout", or similar..).

segments/gitstatus/gitstatus.p9k
dritter: Well. I have that case that I track `ben/next` as local `next`, and if gitstatus would return the remote as well, the segment could show `next -> ben/next`.

romkatv: `ben` is the name of the remote? `git remove -vv | cut -f1` shows it?

If this is what you mean, we can add `VCS_STATUS_REMOTE_NAME` to gitstatus. This value is already computed [here](https://github.com/romkatv/gitstatus/blob/53b5d5dda0a4eed001c4dee9b393050e27e86256/src/git.cc#L282) in order to strip it off remote branch symbolic reference. It could be exported along with other variables.

Would you like to send a PR? The change itself isn't difficult but figuring out how to [build](https://github.com/romkatv/gitstatus/#compiling) gitstatus can be challenging. On the other hand, it'll let you build expertise that can be valuable in the future given that gitstatus is now your dependency (bus factor and all that). I'll take care of building binaries for all architectures but it would be great if you could verify by yourself that the change works on your machine.

segments/gitstatus/gitstatus.p9k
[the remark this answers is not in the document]
romkatv: There are a few options.

The simplest is to just disable timeout. Since `vcs_info` is invoked without a timeout, it kinda makes sense. On the other hand, gitstatus is way less stable than `vcs_info` (by about the same factor it's faster, which is to say a lot less stable). In addition, unlike `vcs_info`, crashes in `gitstatus` will lead to hangs rather than simple errors. All-in-all, I don't recommend going without a timeout.

You could fall back to `vcs_info`. But given that it's you've implemented gitstatus as a separate segment, it makes little sense.

You could pretend it's not a git repo. This will look smooth but can be misleading.

You could display "error" in the segment, perhaps with a hint to blame gitstatus so that people will file bugs directly against me rather bothering you. So, "gitstatus error". I think this is the best option.

segments/gitstatus/gitstatus.p9k
dritter: > ben is the name of the remote? git remove -vv | cut -f1 shows it?

Exactly, yes.

About sending a PR: Yes, I can try that probably by the weekend. But know that I have no experience with C. So chances are that I fail.. But spreading the knowledge is a good thing for sure.

segments/gitstatus/gitstatus.p9k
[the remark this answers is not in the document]
dritter: Okay. Thanks for your input. I output `gitstatus timeout` now, if there is a timeout. Additionally, there is a section for troubleshooting in the README (with the hint to increase the timeout). Think that should be enough, until we get to the point being able to fall back to async.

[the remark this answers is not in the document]
romkatv: That's perfect.

README.md
```diff
@@ -109,6 +109,7 @@ The segments that are currently available are:
 * [`ssh`](segments/ssh/README.md) - Indicates whether or not you are in an SSH session.
 
 **Development Environment Segments:**
+* [`gitstatus`](segments/gitstatus/README.md) - Information about this `git` (if you are in one).
```
Syphdias:
```suggestion
* [`gitstatus`](segments/gitstatus/README.md) - Information about this `git` repository (if you are in one).
```

segments/gitstatus/gitstatus.p9k
```diff
@@ -0,0 +1,149 @@
...
+  local LC_ALL="" LC_CTYPE="en_US.UTF-8"
+
+  ################################################################
+  # Register segment helper icons
+  # Parameters:
+  #   name_of_icon  Generic  Flat/Awesome-Patched  Awesome-FontConfig  Awesome-Mapped-FontConfig  NerdFont
+  #                                          ●                                                                                  
+  p9k::register_icon "GITSTATUS_UNSTAGED"          $'\u25CF'   $'\uE17C'   $'\uF06A'   '\u'${CODEPOINT_OF_AWESOME_EXCLAMATION_CIRCLE}    $'\uF06A'
+#   #                                          ?                                                                                  
+#   p9k::register_icon "GITSTATUS_UNTRACKED"         '?'         $'\uE16C'   $'\uF059'   '\u'${CODEPOINT_OF_AWESOME_QUESTION_CIRCLE}       $'\uF059'
```
Syphdias: Any reason why this is commented out?

dritter: Yep. The reason is that there is a collision with `p9k::register_segment`. This function can optionally register an icon under a given state name. This name then cannot be used for `p9k::register_icon`. This is the case here: We already have a state `UNTRACKED`.. This bugs me and is on my to refactor list in my head.. I have tripped over that more than once.

Syphdias: Oh, I see.

PR #1231, opened 2019-03-27
[Docs] updates `status` README

Syphdias:
This correctly some documentation for the `status` segment and adds some examples.

### DIFF
--- a/segments/status/README.md
+++ b/segments/status/README.md
@@ -21,7 +21,7 @@ This segment shows the return code of the last command.
 
 | Variable | Default Value | Description |
 |----------|---------------|-------------|
-|`P9K_STATUS_CROSS`|`false`|Set to `true` if you wish not to show the error code when the last command returned an error and optionally hide this segment when the last command completed successfully by setting `P9K_STATUS_OK` to `false`.|
+|`P9K_STATUS_CROSS`|`false`|Set to `true` if you wish not to show the error code when the last command returned an error.|
 |`P9K_STATUS_OK`|`true`|Set to `true` if you wish to show this segment when the last command completed successfully, `false` to hide it.|
 |`P9K_STATUS_SHOW_PIPESTATUS`|`true`|Set to `true` if you wish to show the exit status for all piped commands.|
 |`P9K_STATUS_HIDE_SIGNAME`|`false`|Set to `true` to show the raw exit code (`1-255`).  When set to `false`, values over 128 are shown as `SIGNAME(-n)` (e.g. `KILL(-9)`)|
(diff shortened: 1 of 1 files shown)


segments/status/README.md
```diff
@@ -21,7 +21,7 @@ This segment shows the return code of the last command.
 
 | Variable | Default Value | Description |
 |----------|---------------|-------------|
-|`P9K_STATUS_CROSS`|`false`|Set to `true` if you wish not to show the error code when the last command returned an error and optionally hide this segment when the last command completed successfully by setting `P9K_STATUS_OK` to `false`.|
+|`P9K_STATUS_CROSS`|`false`|Set to `true` if you wish not to show the error code when the last command returned an error.|
 |`P9K_STATUS_OK`|`true`|Set to `true` if you wish to show this segment when the last command completed successfully, `false` to hide it.|
 |`P9K_STATUS_SHOW_PIPESTATUS`|`true`|Set to `true` if you wish to show the exit status for all piped commands.|
 |`P9K_STATUS_HIDE_SIGNAME`|`false`|Set to `true` to show the raw exit code (`1-255`).  When set to `false`, values over 128 are shown as `SIGNAME(-n)` (e.g. `KILL(-9)`)|
```
Syphdias: This is incorrect. E.g. exit code `128+9` yields `SIGKILL(9)` not `KILL(-9)`. The question is if it's a fault in the docs or in the code.

dritter: Hm. AFAIK there is no signal `KILL` (at least, `man signal` does not mention a `KILL` signal). The `kill` program _sends_ `SIGKILL`, if you invoke it with parameter `-9`, which is equivalent with invoking it with `-KILL`. But the docs are at bit confusing here..

Syphdias: I always thought`SIGKILL` Just stands for the Signal `KILL`. After all, all Signals start with "SIG", don't they? (No `man signal.7` at hand)

They point is a different though: The p9k docs are not lining up with what is actually happening. Would you say the docs should be adjusted or the code?

dritter: I think we mean the same thing. :D I got you wrong at first.

TBH, I think the small numbers are more useful to most people. No clue if `137` equals `9` is common sense. So I would prefer to change the code.

Syphdias: Does that make more sense?

dritter: Yep. That makes more sense. And Yes, I would fix the docs to add the `SIG` and remove the negative.

dritter:
Thx @Syphdias

PR #1234, opened 2019-03-31
added virtualenv option to show python version  [Enhancement]

jerome-diver:
I add an option to be able to show python version from virtualenv activate.

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -99,134 +99,129 @@ typeset -AH deprecated_variables
 deprecated_variables=(
   # General icons
   # Due to the changes to `p9k::register_icon`, these now require the '_ICON' suffix to work.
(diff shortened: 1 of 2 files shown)


Syphdias:
Hi, @jerome-diver, I'm a bit confused about all the changes you make. I think you only want to change `prompt_virtualenv`. So all your changes would probably be in `segments/virtualenv`
Could you take another look at your changes?

jerome-diver:
@Syphdias yes sure, i need one or two days, i'm vry busy actually. I will follow the correct proceed also.
And yes, it is a little change (3 lines i think) for add an option to segment/virtual_env to be able to read python version when inside a python virutalenv environment.
I did allready an other one PR, but i have no time this day to check that... i'll be back for that.
Thank you

jerome-diver:
hi @Syphdias, i done this pull request instead: https://github.com/bhilburn/powerlevel9k/pull/1241
and it seems to be ok.
But i can not see the change inside master official repo. How to make this officialy accepted ? It should be fine if at the next update i can hold this change on my config and i see other users asked for this change.
What do you think about that change ?

PR #1236, opened 2019-04-01
Improve show_defined_icons

dritter:
`show_defined_icons` prints out all defined icons. Before, it just printed the internal name, which was not very helpful to users. Now it prints the copy&paste ready variable name plus a info on how to use these variable overrides.


Syphdias:
LGTM

dritter:
Thx for review @Syphdias

PR #1237, opened 2019-04-02
[Enhancement] Adding the ability to strip cluster name from AWS ARN if using EKS

Thutm:
Thank you so much for opening a PR for P9k! Many of our best features and segments have come from the community, and we are excited to see your contribution.

To help you make the best PR, here are some guidelines:

  - The `master` branch is our *stable* branch, and the `next` branch is our development branch. If you are submitting a bug fix, please file your PR against `master`. If it is a new feature, enhancement, segment, or something similar, please submit it against `next`. For more information, please see our [Developer's Guide](https://github.com/bhilburn/powerlevel9k/wiki/Developer's-Guide).
  - We maintain unit tests for segments and features in the `test` directory. Please add unit tests for anything new you have developed! If you aren't sure how to do this, go ahead and file your PR and ask for help!
  - For running manual tests in different environments, we have Vagrant and Docker configurations. Please see the [Test README](https://github.com/bhilburn/powerlevel9k/blob/next/TESTS.md) and make sure your new feature is working as expected!
  - If your PR requires user configuration, please make sure that it includes an update to the README describing this.
  - P9k maintains a lot of useful information in our [Wiki](https://github.com/bhilburn/powerlevel9k/wiki). Depending on the content of your PR, we might ask you to update the Wiki (or provide text for us to use) to document your work. Most PRs don't require this.
  - Please make your commit messages useful! Here is a [great short guide on useful commit messages](https://code.likeagirl.io/useful-tips-for-writing-better-git-commit-messages-808770609503).

Once you have submitted your PR, P9k core contributors will review the code and work with you to get it merged. During this process, we might request changes to your code and discuss different ways of doing things. This is all part of the open source process, and our goal is to help you create the best contribution possible for P9k `=)`.

Please follow this template for creating your PR:

#### Title

Please make the title of your PR descriptive! If appropriate, please prefix the title with one of these tags:

- [Bugfix]
- [New Segment]
- [Docs]
- [Enhancement]

#### Description

Please describe the contribution your PR makes! Screenshots are especially helpful, here, if it's a new segment.

If your PR is addressing an issue, please reference the Issue number here.

#### Questions

Is there something in your PR you're not sure about or need help with? Is there a particular piece of code you would like feedback on? Let us know here!

### DIFF
--- a/segments/kubecontext/kubecontext.p9k
+++ b/segments/kubecontext/kubecontext.p9k
@@ -45,6 +45,11 @@ prompt_kubecontext() {
       k8s_final_text="$cur_ctx/$cur_namespace"
     fi
 
+    # Extract cluster name and namespace from AWS ARN, if enabled.
+    if [[ "$P9K_KUBECONTEXT_STRIPEKS" == true ]]; then
+      k8s_final_text="$(echo $k8s_final_text | cut -d':' -f 6 | sed 's/cluster\/*//')"
+    fi
+
     p9k::prepare_segment "$0" "" $1 "$2" $3 "$k8s_final_text"
   fi
 }
--- a/segments/kubecontext/kubecontext.p9k
+++ b/segments/kubecontext/kubecontext.p9k
@@ -36,6 +36,11 @@ prompt_kubecontext() {
       cur_namespace="default"
     fi
 
+    # Extract cluster name and namespace from AWS ARN, if enabled.
+    if [[ "$P9K_KUBECONTEXT_STRIPEKS" == true ]]; then
+      cur_ctx="$(echo $cur_ctx | cut -d':' -f 6 | sed 's/cluster\/*//')"
+    fi
+
     local k8s_final_text=""
 
     if [[ "$cur_ctx" == "$cur_namespace" ]]; then
@@ -43,11 +48,6 @@ prompt_kubecontext() {
       k8s_final_text="$cur_ctx"
     else
       k8s_final_text="$cur_ctx/$cur_namespace"
-    fi
-
-    # Extract cluster name and namespace from AWS ARN, if enabled.
-    if [[ "$P9K_KUBECONTEXT_STRIPEKS" == true ]]; then
-      k8s_final_text="$(echo $k8s_final_text | cut -d':' -f 6 | sed 's/cluster\/*//')"
     fi
 
     p9k::prepare_segment "$0" "" $1 "$2" $3 "$k8s_final_text"
--- a/segments/kubecontext/README.md
+++ b/segments/kubecontext/README.md
@@ -22,3 +22,10 @@ P9K_KUBECONTEXT_BACKGROUND='blue'
 
 The main Icon can be changed by setting `P9K_KUBECONTEXT_ICON="my_icon"`. To change the
 icon color only, set `P9K_KUBECONTEXT_ICON_COLOR="red"`.
+
+### Strip Cluster and Namespace from AWS ARN (If using EKS)
+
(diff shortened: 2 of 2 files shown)


Thutm:
For testing, I need to add a mock AWS ARN string to be returned on by kubectl. Should I add to the existing tests or just duplicate all the current tests referencing "minikube"? I can't think of any reasons for additional mock strings to test against and didn't want to unnecessarily update all existing tests.

dritter:
Hi @Thutm ,

Thanks for the PR. I am a fan of one test per use-case, so please add some new tests. I think it will be enough to add a positive and a negative test.

segments/kubecontext/kubecontext.p9k
```diff
@@ -36,6 +36,11 @@ prompt_kubecontext() {
       cur_namespace="default"
     fi
 
+    # Extract cluster name and namespace from AWS ARN, if enabled.
+    if [[ "$P9K_KUBECONTEXT_STRIPEKS" == true ]]; then
+      cur_ctx="$(echo $cur_ctx | cut -d':' -f 6 | sed 's/cluster\/*//')"
```
dritter: I am pretty sure that can be done without piping to external programs. Once you added some tests and I know what `$cur_ctx` is, I can help you with that.

Thutm: Sure, i'll add new tests rather than modifying. The $cur_ctx string that comes back when using AWS EKS is something like this:

arn:aws:eks:us-east-1:1234567890:cluster/dev-eks/default

Where 1234567890 is the account ID. It follows the AWS ARN pattern. I put screenshots in the PR description. Do we typically want to avoid using external binaries to retrieve the output we want? I noticed some other segments were using cut/sed so I was trying to follow current practice.

dritter: >  Do we typically want to avoid using external binaries to retrieve the output we want? I noticed some other segments were using cut/sed so I was trying to follow current practice.

Yes. At least to a certain extend. It is usually vastly slower to execute an external binary, than to query a variable. But, on the other hand, if the variable is only valid in certain circumstances, than it might be better to call the external binary. As a rule of thumb, external calls should be avoided as much as possible. Regarding the use of cut/sed: We are currently trying to get rid of all these calls, but we are not finished yet.

dritter: `"${${cur_ctx##*\:}/cluster\//}"` This strips everything before the last colon away, and removes `cluster/`. So this should have the same effect like `| cut -d':' -f 6 | sed 's/cluster\/*//'`, without the subshells.

Syphdias: This should do the same thing:
```suggestion
      cur_ctx="${${(s.:.)cur_ctx}[6]#cluster/}"
```

Syphdias: Please check before use :)

dritter: Or this: `"${${cur_ctx##*\:}/cluster\//}`. 😄

Thutm:
Thanks all, I actually ended up using:
`
      cur_ctx=${cur_ctx#*:*:*:*:*:};
      cur_ctx=${cur_ctx:8}
`

and then updated it to dritters suggestion. Let me know if that works.

segments/kubecontext/kubecontext.p9k
Syphdias: I hope there is no `:` in the last part. Because then all suggestions fail :laughing:
If there are not additional `/`s then you could even go for `${cur_ctx:t}`

Thutm:
Screen Shot 2019-04-16 at 2 00 52 PM

Attached segment screenshot. No terraform specific icon that I could find. Used a globe instead for now.

PR #1238, opened 2019-04-03
Fix background_jobs segment

### LINKED ISSUE #1196: [Regression] background_jobs segment doesn't update in next branch when jobs finish
In `release` branch, when a background job completes, `background_jobs` segment updates. In `next` it doesn't.

Here's `master`:

```zsh
  git clone -b master https://github.com/bhilburn/powerlevel9k.git ~/powerlevel9k
  echo "
    POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=()
    POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(background_jobs)
    source ~/powerlevel9k/powerlevel9k.zsh-theme
    sleep 3&" >~/.zshrc
  zsh -i'
```

job-complete-master

Replace `master` with `next` to test with `next` branch.

job-complete-next

Notice how `next` is still saying there is a background job after the job completes.

dritter:
This commit replaces the precmd hook with a call to `jobs -l`.
It is not ideal, but gets the job done.

Performance: This is not nearly as fast as using `%j`, but that yields the total count of background jobs only. And this is probably slower than the former precmd solution.

So, this is a possible Plan B, if we don't find a better solution.

Fixes #1196

### DIFF
--- a/segments/background_jobs/README.md
+++ b/segments/background_jobs/README.md
@@ -13,7 +13,8 @@ where you want to show this segment.
 | Variable | Default Value | Description |
 |----------|---------------|-------------|
 |`P9K_BACKGROUND_JOBS_VERBOSE`|`true`|If there is more than one background job, this segment will show the number of jobs. Set this to `false` to turn this feature off.|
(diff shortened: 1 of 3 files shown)


dritter:
Closed in favor of #1254

Issue #1209, opened 2019-03-15
vcs segment shows incorrect remote branch when tracking remote isn't called "origin"


romkatv:
Tested on master with the default config.

1. Clone a repo from github.
2. `cd` into it.
3. `git remote rename origin source`

Expected: There is no remote branch in the vcs prompt (because remote is the same as local).
Actual: vcs prompt shows remote branch "source/master".

dritter:
Hey @romkatv .

I've been thinking about this, since I implemented the remote name in `gitstatus` and used that to prepend the branch name if the remote name differs from `origin`. And that leads to the same error in the gitstatus segment.
And, to be honest, I don't know how if we can fix this. I don't see a proper way to see if a remote is "owned" by the user or not. As far as I know, git makes not difference between a renamed remote or a remote that was just added. All remotes are equal (I might be wrong). But if this is the case, the only thing we could do about that is to check for `origin` and show the remote name otherwise.

Syphdias:
Maybe I'm reading this wrong, but what exactly is wrong?
![image](https://user-images.githubusercontent.com/16988672/55658655-b4e92a00-57fe-11e9-8d4d-b9d31c92941a.png)

romkatv:
> Maybe I'm reading this wrong, but what exactly is wrong?

Oops, sorry about that. I should've tried my own reproduction instructions before posting them. This should demonstrate the issue:

```zsh
git remote rename origin my/source
```

Full setup:

```zsh
docker run -e LANG=C.UTF-8 -e LC_ALL=C.UTF-8 -e TERM=$TERM -it --rm ubuntu bash -uexc '
  apt update && apt install -y zsh git
  git clone https://github.com/bhilburn/powerlevel9k.git ~/powerlevel9k
  cd ~/powerlevel9k
  git remote rename origin my/source
  echo "
    POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(vcs)
    POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=()
    source ~/powerlevel9k/powerlevel9k.zsh-theme" >~/.zshrc
  zsh -i'
```

![wrong-remote](https://user-images.githubusercontent.com/1282067/55667291-9a449e80-585a-11e9-9fa0-f4ead2264e61.png)

[...]

PR #1248, opened 2019-04-12
Clarify remote name <-> remote branch

dritter:
This fixes the original issue from #1209 . The `vcs` segment now shows `<remote_branch>@<remote_name>`, if there is a remote. The delimiter is configurable by setting `P9K_VCS_REMOTE_DELIMITER`.

Bildschirmfoto 2019-04-13 um 01 28 16

I hope that clarifies what the remote name is and what the remote branch is. Performance might got a bit worse, if a remote was detected (one subshell more)..

Additionally, this PR fixes some documentation issues and moves truncating the branch name in a better named hook (`git-branch`, instead of `git-remote-branch`).

// cc @Syphdias @romkatv

### DIFF
--- a/segments/vcs/README.md
+++ b/segments/vcs/README.md
@@ -32,6 +32,7 @@ customization is provided via:
 |`P9K_VCS_SVN_HOOKS`|`(vcs-detect-changes svn-detect-changes)`|Layout of the segment for SVN repositories.|
 |`P9K_VCS_ACTIONFORMAT_FOREGROUND`|`red`|The color of the foreground font during actions (e.g., `REBASE`).|
 |`P9K_VCS_GIT_ALWAYS_SHOW_REMOTE_BRANCH`|`false`|Set to true If you would to always see the remote branch.|
+|`P9K_VCS_REMOTE_DELIMITER`|`@`|This delimits the remote name from the remote branch. E.g. `master@origin`.|
 
 ### vcs symbols
 
--- a/segments/vcs/README.md
+++ b/segments/vcs/README.md
@@ -67,8 +67,8 @@ Customizations available are:
 |----------|---------------|-------------|
 |`P9K_VCS_SHORTEN_LENGTH`|None|This field determines how many characters to show.|
 |`P9K_VCS_SHORTEN_MIN_LENGTH`|None|This field determines minimum branch length. Branch name will be truncated if its length greater than this field.|
-|`P9K_VCS_DIR_SHORTEN_STRATEGY`|None|This field determines how branch name should be truncated. See the table below for more information.|
-|`P9K_DIR_SHORTEN_DELIMITER`|`...`|Delimiter to use in truncated strings. This can be any string you choose, including an empty string if you wish to have no delimiter.|
+|`P9K_VCS_SHORTEN_STRATEGY`|None|This field determines how branch name should be truncated. See the table below for more information.|
+|`P9K_VCS_SHORTEN_DELIMITER`|`...`|Delimiter to use in truncated strings. This can be any string you choose, including an empty string if you wish to have no delimiter.|
 
 | Strategy Name | Description |
 |---------------|-------------|
@@ -79,8 +79,8 @@ For example, if you want to truncate `1234-super_super_long_branch_name` to `123
 ```zsh
 P9K_VCS_SHORTEN_LENGTH=4
 P9K_VCS_SHORTEN_MIN_LENGTH=11
-P9K_VCS_DIR_SHORTEN_STRATEGY="truncate_from_right"
-P9K_VCS_DIR_SHORTEN_DELIMITER=".."
+P9K_VCS_SHORTEN_STRATEGY="truncate_from_right"
+P9K_VCS_SHORTEN_DELIMITER=".."
 ```
 
 ### Advanced features
--- a/segments/vcs/README.md
+++ b/segments/vcs/README.md
@@ -27,7 +27,7 @@ customization is provided via:
 |`P9K_VCS_CHANGESET_HASH_LENGTH`|`12`|How many characters of the hash / changeset to display in the segment.|
 |`P9K_VCS_SHOW_SUBMODULE_DIRTY`|`true`|Set to `false` to not reflect submodule status in the top-level repository prompt.|
 |`P9K_VCS_HIDE_TAGS`|`false`|Set to `true` to stop tags being displayed in the segment.|
-|`P9K_VCS_GIT_HOOKS`|`(vcs-detect-changes vcs-icon git-untracked git-aheadbehind git-stash git-remotebranch git-gitdir git-tagname)`|Layout of the segment for git repositories.|
+|`P9K_VCS_GIT_HOOKS`|`(vcs-detect-changes vcs-icon git-gather-data git-untracked git-aheadbehind git-stash git-branch git-remotebranch git-gitdir git-tagname)`|Layout of the segment for git repositories.|
 |`P9K_VCS_HG_HOOKS`|`(vcs-detect-changes)`|Layout of the segment for Mercurial repositories.|
 |`P9K_VCS_SVN_HOOKS`|`(vcs-detect-changes svn-detect-changes)`|Layout of the segment for SVN repositories.|
 |`P9K_VCS_ACTIONFORMAT_FOREGROUND`|`red`|The color of the foreground font during actions (e.g., `REBASE`).|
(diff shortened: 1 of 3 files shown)


Syphdias:
A few observations:
- The first time I tried it, I ended up with `master→origin/master@` since I had set some git hooks. I then added `git-gather-data`, that led to loosing the branch entirely. This will definatly confuse people if the modify (or have modified) their hooks
- The \@-syntax is understandable but is not really git-syntax that can be copied. Just something to consider. Also setting the delimiter will not make that true since it's `<branch><remote>`.
- I would consider making the test a little bit more solid (one if these asserts actually fails):
```diff
diff --git a/segments/vcs/vcs_git.spec b/segments/vcs/vcs_git.spec
index b963109b..859b5aac 100755
--- a/segments/vcs/vcs_git.spec
+++ b/segments/vcs/vcs_git.spec
@@ -425,11 +425,15 @@ function testAlwaysShowRemoteBranch() {
 
   git clone . ../vcs-test2 1>/dev/null 2>&1
   cd ../vcs-test2
+  git remote rename origin some/remote
 
-  assertEquals "%K{002} %F{000} master→master@origin %k%F{002}%f " "$(__p9k_build_left_prompt)"
+  assertEquals "%K{002} %F{000} master→master@some/remote %k%F{002}%f " "$(__p9k_build_left_prompt)"
 
   local P9K_VCS_GIT_ALWAYS_SHOW_REMOTE_BRANCH='false'
   assertEquals "%K{002} %F{000} master %k%F{002}%f " "$(__p9k_build_left_prompt)"
+
+  git branch -m master new/ref/master
+  assertEquals "%K{002} %F{000} new/ref/master→master@some/remote %k%F{002}%f " "$(__p9k_build_left_prompt)"
 }
 
 function testGitDirClobber() {
```

dritter:
Thanks for the review @Syphdias . Yep, how we merge that configuration array is still unsolved. I'm open to suggestions ;) The obvious one - merge that automatically - won't work, as we cannot know if a user configured that hook away for any reason, or if that hook was added by us..

I deliberately went away from the git syntax, because the git syntax is ambiguous as well (`<remote>/<branch>`; at least if using `git rev-parse`), if your branch contains slashes. A whitespace could be another solution. Do you like that better?

Agreed with making the tests more robust. Will do.

Syphdias:
> Thanks for the review @Syphdias . Yep, how we merge that configuration array is still unsolved. I'm open to suggestions ;) The obvious one - merge that automatically - won't work, as we cannot know if a user configured that hook away for any reason, or if that hook was added by us..

Two things:
1. I don't entirely get why you split it into two hooks. I suppose you need `git-gather-data` because you do something irreversible in a later hook and need the data. Could you make it a normal function that is called in the hook(s) where you need it, so there are no hooks depending on another hook? I'm not sure if there are other "hook dependencies" but I would like to avoid them firstly from a UX perspective and second to avoid github issues (aka work for us :wink:) which will be caused from people using it "wrongly".
2. The `@` at the end I described should be easily avoidable by something like this in `+vi-git-remotebranch` (ugly but functional):
```zsh
[[ -n ${remote_name} ]] \
  && hook_com[branch]+="${__P9K_ICONS[VCS_REMOTE_BRANCH]}${remote_branch}${P9K_VCS_REMOTE_DELIMITER}${remote_name}" \
  || hook_com[branch]+="${__P9K_ICONS[VCS_REMOTE_BRANCH]}${remote_branch}"
```


> I deliberately went away from the git syntax, because the git syntax is ambiguous as well (`<remote>/<branch>`; at least if using `git rev-parse`), if your branch contains slashes. A whitespace could be another solution. Do you like that better?

Just to be clear: I'm not totally disagreeing with the `@`. It's just usually used differently in git references. Not sure if whitespace would work.  
I also thought of `//` since you can neither start a branch with `/` nor end a remote on it; but also non-standard.  
Github uses `:` in PRs which cannot be used in either branch or remote name but is used in pushing afaik.

I would opt for `:` but I'll leave it up to you.

dritter:
> [...] I suppose you need `git-gather-data` because you do something irreversible in a later hook and need the data.

Correct. It is possible to truncate long branch names. That is why I saved the original branch name as early as possible.

> [...] Could you make it a normal function [...]?

Well. The problem with that approach would be that we either need to cache the value on the first call (and there is no guarantee that the first call already truncated the branch name), or that we have to query `git` all the time to get the branch name..

> Github uses : in PRs which cannot be used in either branch or remote name but is used in pushing afaik.

Fair point. The main problem is that our approach is still not very flexible. We are bound to the initial structure (branch first, remote second information). Sure, we could change that, but not the user (I'd love to have some simple templating for such things).
GitLab uses a slash, GitHub a colon, but both write first remote, then branch name. I'll change to colon as well and remote first.

dritter:
Hmm. I even introduced a new hook `git-branch-icon`, that shows the icon. With this hook, users can decide, where the branch icon should show up. This could be done as function as well, but then we could not decide, if the icon was already show. This is necessary, to show the icon for remote (`git-remote-branch`) and local branches (`git-branch`). Of course, if we do not want to show an branch icon for the remote branch, we could just include the icon in the normal `git-branch` hook..

Syphdias:
Friendly reminder, this is still an issue:

image
Additional info this even happens with `P9K_VCS_GIT_ALWAYS_SHOW_REMOTE_BRANCH=false`. Maybe this could work?
```diff
diff --git a/segments/vcs/vcs.p9k b/segments/vcs/vcs.p9k
index e3581168..3cc48e51 100644
--- a/segments/vcs/vcs.p9k
+++ b/segments/vcs/vcs.p9k
@@ -217,7 +217,7 @@ function +vi-git-remotebranch() {
   local remote=${$(command git rev-parse --verify HEAD@{upstream} --symbolic-full-name 2>/dev/null)/refs\/(remotes|heads)\/}
 
   # Only show the remote if it differs from the local
-  if [[ -n ${remote} \
+  if [[ -n ${remote} && -n ${__P9K_VCS_DATA[branch]} \
       && ( \
         "${P9K_VCS_GIT_ALWAYS_SHOW_REMOTE_BRANCH}" == 'true' \
         || "${remote#*/}" != "${__P9K_VCS_DATA[branch]#heads/}" \
```


Also, I think we discussed this before. Is this wanted behaviour, showing the remote if it's not names "origin"?
image
My idea to checking this in the tests:
```diff
diff --git a/segments/vcs/vcs_git.spec b/segments/vcs/vcs_git.spec
index 662af683..efc165d6 100755
--- a/segments/vcs/vcs_git.spec
+++ b/segments/vcs/vcs_git.spec
@@ -429,6 +429,7 @@ function testAlwaysShowRemoteBranch() {
   assertEquals "%K{002} %F{000} master %k%F{002}%f " "$(__p9k_build_left_prompt)"
 
   git remote rename origin some/remote
+  assertEquals "%K{002} %F{000} master %k%F{002}%f " "$(__p9k_build_left_prompt)"
   local P9K_VCS_GIT_ALWAYS_SHOW_REMOTE_BRANCH='true'
   assertEquals "%K{002} %F{000} master→some/remote:master %k%F{002}%f " "$(__p9k_build_left_prompt)"
 ```
And my approach to fixing it (this includes the bug fix from above):
```diff
diff --git a/segments/vcs/vcs.p9k b/segments/vcs/vcs.p9k
index e3581168..dd544527 100644
--- a/segments/vcs/vcs.p9k
+++ b/segments/vcs/vcs.p9k
@@ -217,12 +217,12 @@ function +vi-git-remotebranch() {
   local remote=${$(command git rev-parse --verify HEAD@{upstream} --symbolic-full-name 2>/dev/null)/refs\/(remotes|heads)\/}
 
   # Only show the remote if it differs from the local
-  if [[ -n ${remote} \
+  local remote_name=$(command git config branch.${__P9K_VCS_DATA[branch]#heads/}.remote)
+  if [[ -n ${remote} && -n ${__P9K_VCS_DATA[branch]} \
       && ( \
         "${P9K_VCS_GIT_ALWAYS_SHOW_REMOTE_BRANCH}" == 'true' \
-        || "${remote#*/}" != "${__P9K_VCS_DATA[branch]#heads/}" \
+        || "${remote#${remote_name}/}" != "${__P9K_VCS_DATA[branch]#heads/}" \
       ) ]]; then
-    local remote_name=$(command git config branch.${__P9K_VCS_DATA[branch]#heads/}.remote)
     # Remote Branch is the remote branch name, without the remote name.
     # Additionally, remove trailing whitespace and escape percent signs.
     local remote_branch=${${${remote#${remote_name}/}// /}//\%/%%}
```

Sorry that I'm turning you into the reviewer now :sweat_smile: 

I still think this might case some issues with people who do not use the default hooks. We might need to inform users above the change, but I'm not sure if that causes more or less confusion.

PS: I find this exceptionally hard to test with all the possibilities and chances of weird edge cases. I think it is very hard to get this segment perfect. Kudos to you for trying!

PR #1250, opened 2019-04-15
Fix prompt status error

MichaelAquilina:
#### Description

I pull this plugin from the `next` branch (for the new features it provides). Recently I've noticed that my local machine shows the following error on startup:

```
prompt_status:16: bad math expression: operand expected at end of string
```

Looking into it, it seems like both the `ec` and the `ec_sum` variable are not being initialised. This results in an incorrect arithmetic expression of `$(( + ))` being evaluated.

The actual solution for this might be to actually fix the underlying cause but I though this PR would be appropriate to start a discussion about where the issue might be and why it is showing up for me specifically.

### DIFF
--- a/segments/status/status.p9k
+++ b/segments/status/status.p9k
@@ -62,7 +62,7 @@ prompt_status() {
 
     for ec in "${(@)RETVALS[2,-1]}"; do
       ec_text="${ec_text}|$(exit_code_or_status "$ec")"
-      ec_sum=$(( $ec_sum + $ec ))
+      ec_sum=$(( ${ec_sum:-0} + ${ec:-0} ))
     done
   else
     # We use RETVAL instead of the right-most RETVALS item because


dritter:
Thanks for the PR @MichaelAquilina . The code looks solid to me. `RETVAL` and `RETVALS` are initialised in `__p9k_save_retvals`, which is defined in `generators/default.p9k`.
In my tests the exit code was always set correctly. Do you have some special ZSH options set?

I'm merging this fix any way, as it is nice and small. Thank you very much.

MichaelAquilina:
Would be interesting to see why my specific case is failing.

The settings I have set for powerlevel are the following

```
# Theme
zplug "bhilburn/powerlevel9k", as:theme, at:next
P9K_LEFT_PROMPT_ELEMENTS=(status dir vcs dir_writable)
P9K_RIGHT_PROMPT_ELEMENTS=()
P9K_DIR_SHORTEN_STRATEGY="dir"
P9K_DIR_SHORTEN_LENGTH=2
P9K_VIRTUALENV_BACKGROUND="cyan"
```

Syphdias:
@MichaelAquilina Could you try putting the zplug line to the end?

MichaelAquilina:
@Syphdias still the same issue. The plugin is not loaded at the `zplug` command. Its loaded at a later step when `zplug load` is run.

PR #1264, opened 2019-04-30
[Enhancement] Bold Icons

### LINKED ISSUE #1223: [Feature request] Bold prompt segments
Add `FOO_BOLD` option (defaults to `false`, can be set to `true`) for every `FOO` for which there is currently `FOO_FOREGROUND` or `FOO_COLOR`.

For example, `POWERLEVEL9K_TIME_BOLD=true` to make `time` segment bold. `POWERLEVEL9K_TIME_VISUAL_IDENTIFIER_BOLD=true` to make the icon in `time` bold.

Syphdias:
Instead of `_BACKGROUND`, `_FOREGROUND` or `_ICON` you can append:
- `_BOLD` to the stateful name to get bold text
- `_ICON_BOLD` to the stateful name to get bold visual identifier/"icon"
  (this only works if the font allows for it to be bold)

The `_BOLD` feature seems to be undocumented which should be changed. And
also document `_ICON_BOLD` next to it.

aims to resolves #1223

### DIFF
--- a/functions/icons.zsh
+++ b/functions/icons.zsh
@@ -159,6 +159,9 @@ function p9k::register_segment() {
   local BOLD_USER_VARIABLE="P9K_${STATEFUL_NAME}_BOLD"
   local BOLD="${(P)BOLD_USER_VARIABLE}"
   [[ -z "${BOLD}" ]] || __P9K_DATA[${STATEFUL_NAME}_BD]=true
+  local BOLD_ICON_USER_VARIABLE="P9K_${STATEFUL_NAME}_ICON_BOLD"
+  local BOLD_ICON="${(P)BOLD_ICON_USER_VARIABLE}"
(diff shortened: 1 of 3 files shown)


bhilburn:
This is a much better way to have done this, thanks @Syphdias.

And yes, great point regarding the missing documentation - this should really be added to the README. Were you planning on pushing a separate PR for that?

Syphdias:
> Were you planning on pushing a separate PR for that?

Like #1265? 😉

Syphdias:
I just checked.This has the same issue as described in #1302 in ZSH 5.2

dritter:
> I just checked.This has the same issue as described in #1302 in ZSH 5.2

Thanks for checking. I'm merging anyway, as this is on the path to v0.7.0.

Thanks for the PR @Syphdias .

PR #1268, opened 2019-05-11
Fix completion menu

### LINKED ISSUE #1267: POWERLEVEL9K_RPROMPT_ON_NEWLINE=false breaks completion menu
This is on `master`.

To reproduce:

1. Enable completion menu.
2. Resize your terminal so that it has fewer than 20 lines (confirm with `echo $LINES`).
3. Set `POWERLEVEL9K_PROMPT_ON_NEWLINE=true` and `POWERLEVEL9K_RPROMPT_ON_NEWLINE=false`.
4. Type `grep --` and press `TAB` twice.

Steps 1-3 more precisely (you still need to type `grep --` and press `TAB` twice once ZSH prompt appears):

```zsh
  cd
  git clone https://github.com/bhilburn/powerlevel9k.git
  echo "
    zstyle :completion:\* menu yes select
    autoload -Uz compinit && compinit
    POWERLEVEL9K_PROMPT_ON_NEWLINE=true
    POWERLEVEL9K_RPROMPT_ON_NEWLINE=false
    source ~/powerlevel9k/powerlevel9k.zsh-theme" >~/.zshrc
  exec zsh'
```

Actual behavior:

actual

Expected behavior:

expected

dritter:
This fixes #1267 .

The problem was that we move the cursor around to place the right hand prompt. So the fix is to move the cursor back up in the right completion menu.

### DIFF
--- a/generator/default.p9k
+++ b/generator/default.p9k
@@ -407,6 +407,8 @@ ${__P9K_ICONS[MULTILINE_LAST_PROMPT_PREFIX]}"
     else
       RPROMPT_PREFIX='%{'${P9K_RPROMPT_PREFIX:-}'%}'
       RPROMPT_SUFFIX='%{'${P9K_RPROMPT_SUFFIX:-}'%}'
+      # Move right completion menu one line up
+      RPROMPT2='%{'$'\e[1A''%}'"${RPROMPT2}"
     fi
   else
     __p9k_unsafe_PROMPT="%f%b%k$(__p9k_build_left_prompt)"


Syphdias:
I can still reproduce the described bug like this. Remember to shrink your terminal to 20-30 lines in height. Also I think your changes are not even run in the case described by the issue.

```zsh
  cd
  apt update && apt install -y zsh git vim
  git clone https://github.com/dritter/powerlevel9k.git -b fix_completion_menu
  echo "
    zstyle :completion:\* menu yes select
    autoload -Uz compinit && compinit
    P9K_PROMPT_ON_NEWLINE=true
    P9K_RPROMPT_ON_NEWLINE=false
    source ~/powerlevel9k/powerlevel9k.zsh-theme" >~/.zshrc
  exec zsh'
```

dritter:
Ha! You are right. I tricked myself by setting `P9K_RPROMPT_ON_NEWLINE=true` accidentally. 🙄

PR #1271, opened 2019-05-14
[Bugfix] Fix extra tilde when using truncate_with_folder_marker under $HOME

mr-salty:
truncate_with_folder_marker adds a tilde when it finds HOME, but
POWERLEVEL9K_HOME_FOLDER_ABBREVIATION is also prepended, so if you're
in a subdirectory of $HOME you get an extra tilde (although this does not
occur if you're actually in $HOME)

Without the change, the new test I added fails:
```
testHomeWithFolderMarkerWorks
ASSERT:expected:<%K{004} %F{000}~/powerlevel9k-test %k%F{004}%f > but was:<%K{004} %F{000}~~/powerlevel9k-test %k%F{004}%f >
```

### DIFF
--- a/powerlevel9k.zsh-theme
+++ b/powerlevel9k.zsh-theme
@@ -874,7 +874,7 @@ prompt_dir() {
               trunc_path="/"
             elif [[ "$marked_folder" == "$HOME" ]]; then
               # If we reached home folder, stop upsearch.
-              trunc_path="~"
+              trunc_path=""
             elif [[ "${marked_folder%/*}" == $last_marked_folder ]]; then
               trunc_path="${trunc_path%/}/${marked_folder##*/}"
             else
--- a/test/segments/dir.spec
+++ b/test/segments/dir.spec
@@ -213,6 +213,28 @@ function testTruncateWithFolderMarkerWorks() {
   rm -fr $BASEFOLDER
 }
 
+function testHomeWithFolderMarkerWorks() {
+  local -a POWERLEVEL9K_LEFT_PROMPT_ELEMENTS
+  POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(dir)
+  local POWERLEVEL9K_SHORTEN_STRATEGY="truncate_with_folder_marker"
+  local POWERLEVEL9K_HOME_FOLDER_ABBREVIATION='~'
+  local dir=$PWD
+
+  mkdir -p ~/powerlevel9k-test
+
+  # Load Powerlevel9k
+  source ${P9K_HOME}/powerlevel9k.zsh-theme
+
+  cd ~
+  assertEquals "%K{004} %F{000}~ %k%F{004}%f " "$(build_left_prompt)"
+
+  cd ~/powerlevel9k-test
+  assertEquals "%K{004} %F{000}~/powerlevel9k-test %k%F{004}%f " "$(build_left_prompt)"
+
+  cd $dir
+  rm -fr ~/powerlevel9k-test
+}
+
 function testTruncateWithFolderMarkerWithChangedFolderMarker() {
   local -a POWERLEVEL9K_LEFT_PROMPT_ELEMENTS
   POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(dir)


romkatv:
`truncate_with_folder_marker` might be more broken than this fix implies. This PR makes it work correctly as long as there are no markers, but in this case there is also no truncation. Try creating some markers and see what you get. Given how broken it is, I doubt anyone is using this truncation strategy other than by mistake.

mr-salty:
ah, I see what you mean. I can take a shot at fixing that as well.  I filed #1272 to track it, if you have anything to add.

FWIW, I'm using it for a specific purpose where I didn't like any of the other strategies, but the directories where I have folder markers aren't under $HOME so I didn't notice.

romkatv:
> FWIW, I'm using it for a specific purpose where I didn't like any of the other strategies, but the directories where I have folder markers aren't under $HOME so I didn't notice.

Can you share extra details? What are the directories outside of your `$HOME` where you add markers? Are you committing markers to piper so that your citc paths are shortened?

In my opinion this shortening strategy shouldn't have existed. Instead, Powerlevel9k should respect ZSH named directories, both [static](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Static-named-directories) and [dynamic](http://zsh.sourceforge.net/Doc/Release/Expansion.html#Dynamic-named-directories). If you can shed some light on how you are using `truncate_with_folder_marker`, it would help me understand whether my hunch is right. I could then add support for named directories, which not only would make paths look nicer for you, but those paths would be real in a sense that you can `cd` into them or use them as an argument to `ls` or anything else.

mr-salty:
Yeah, that's right (ah, didn't realize you were an xoogler), my directory is i.e.: `/google/src/cloud/salty/client-name/google3`; helpfully each client has `.citc` in its root, so I use that as a marker and get `/…/client-name/google3` which saves real estate. I'd be happy to have some alias like `citc` show up there instead.

Semi-unrelated, but ideally I'd like to use the perforce vcs stuff and have that interact with the directory name. When I was using tcsh, my prompt was just `<client-name>/google3`. I hacked the zsh perforce vcs stuff to work but it was so slow I turned it back off until I have some time to speed it up (maybe I'll try it with 10k)... but it also didn't do anything for the directory name.

romkatv:
Thanks for the explanation. I'll leave the discussion of p4/g4 integration for another time. As far as directory is concerned, here's what you can do in ZSH without any plugins or themes.

First, start zsh like so:

```zsh
PROMPT='%~ ' zsh -df
```

This will give you a simple prompt that contains just your current directory. Notice how your home directory is abbreviated as `~`.

You can define a static named directory for citc.

```zsh
hash -d citc=/google/src/cloud/salty
```

Now, if you are in `/google/src/cloud/salty/client-name/google3`, your prompt will show `~citc/client-name/google3`. Moreover, you can type `cd ~citc/client-name` and it'll work. You can even type `cd ~citc/<TAB>` and the completion system will work, too.

This is _static named directory_. It's static because for every named directory of this kind we need to run `hash` (you usually put these calls in `~/.zshrc`).

There are also dynamic named directories, which allow you to have unlimited number of named directories whose names and targets depend on external data. They look like this: `~[name]`. For example, you could make `~[client-name]` map to `/google/src/cloud/salty/client-name`. Once again, this mapping will work everywhere the same way `~ <=> $HOME` works. http://zsh.sourceforge.net/Doc/Release/Expansion.html#Dynamic-named-directories explains how to define dynamic named directories.

Static and dynamic named directories give you a lot of flexibility that doesn't require builtin support from Powerlevel9k. Plus, they integrate with ZSH seamlessly.

Ideally, Powerlevel9k should show named directories in the prompt just like native ZSH prompt does it. Unfortunately, it doesn't work. I think making it work would be valuable, and it's also not difficult.

FWIW, `simple_dir` in Powerlevel10k does support named directories. If you just want to try them, you can replace `dir` with `simple_dir` in your config, define a bunch of named directories and see if you like it. Note, however, that `simple_dir` doesn't support truncation and a few other configuration options. On the positive side it's super fast. Keep in mind that it's an undocumented segment that I threw together as a demo and thus it can disappear at any time.

romkatv:
@mr-salty FYI: [Powerlevel10k](https://github.com/romkatv/powerlevel10k) now supports named directories, both static and dynamic. All shortening strategies and other dir prompt options work properly when you are under a named directory.

You don't need to do anything special to enable it. Just update Powerlevel10k, add `hash -d citc=/google/src/cloud/salty` to your `~/.zshrc` and you'll get a nice prompt.

PR #1275, opened 2019-05-17
removes truncate_absolute_chars since it is redundant

Syphdias:
The `P9K_DIR_SHORTEN_STRATEGY` `truncate_absolute` and `truncate_absolute_chars` are identical in function. I removed `truncate_absolute_chars` from the docs but kept it functional for compatibility reasons.
I picked the explanation from the docs that seemed easier to understand and expanded on it.


bhilburn:
Solid 👍

PR #1278, opened 2019-05-19
[Bugfix] fixes hidden segment on last line

### LINKED ISSUE #1266: Inconsistent format of line endings in left prompt
This is in `master`.

Config:

```zsh
POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(dir newline background_jobs newline dir_writable)
POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=()
POWERLEVEL9K_PROMPT_ADD_NEWLINE=true
```

Actual behavior:

actual

Unexpected behavior:

1. In the first prompt the second line doesn't have a triangle at the end, while the third line does. Note that both of these lines are otherwise empty.
2. In the third prompt the second line has an unfilled triangle at the end, while the third line has it filled.

Expected behavior: All lines have consistent formatting of the trailing triangle.

expected

FYI: The expected screenshot is from [Powerlevel10k](https://github.com/romkatv/powerlevel10k/).

Syphdias:
This fixed a bug where the last line is not properly rendered if there is a hidden segment at the end:  
`P9K_LEFT_PROMPT_ELEMENTS=(dir newline background_jobs newline dir_writable)`
image

This PR initially was created to fix #1266. What this does not change is the displaying of the "triangle" (`LEFT_SEGMENT_SEPARATOR`) at the end on newline. Currently the triangle is not shown if there is no segment on this line. This [could be considered a bug](https://github.com/bhilburn/powerlevel9k/issues/1266#issuecomment-493737977) as @romkatv notes, depending on if want that or not. So the question is: Should the triangle be displayed on empty lines? If no, this PR is finished (unless the review brings something up), if so, I could expand on this PR.

### DIFF
--- a/generator/default.p9k
+++ b/generator/default.p9k
@@ -108,9 +108,7 @@ function __p9k_left_prompt_segment() {
 # @noargs
 ##
 function __p9k_left_prompt_end() {
-  if [[ ${(U)CURRENT_BG} == 'NONE' ]]; then
-    echo -n "%k%f${__P9K_ICONS[LEFT_SEGMENT_SEPARATOR]}"
-  elif [[ -n ${CURRENT_BG}  ]]; then
+  if [[ -n ${CURRENT_BG} && ${(U)CURRENT_BG} != 'NONE' ]]; then
     echo -n "%k%F${CURRENT_BG#%K}${__P9K_ICONS[LEFT_SEGMENT_SEPARATOR]}"
   else
     echo -n "%k"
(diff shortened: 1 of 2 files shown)


bhilburn:
Nice fix, @Syphdias, and thanks to linking to the comment from Roman.

I think he has a good point - there should be consistent indication of a rendered line, otherwise it appears as though the line above it contained an `\n` when it didn't, which can really matter for things like text parsing / REPL. 

Can you add the triangle to the empty line? Aside from that, this looks good! =)

dritter:
Thx @Syphdias .

Syphdias:
> I think he has a good point - there should be consistent indication of a rendered line, otherwise it appears as though the line above it contained an `\n` when it didn't, which can really matter for things like text parsing / REPL.

@dritter are you aware that I didn't come around to implementing the changes @bhilburn suggested?

Issue #1281, opened 2019-05-22
pyenv shows incorrect version


romkatv:
This is on `master`.

To reproduce:

1. Install pyenv.
2. Use `master` branch of Powerlevel9k with `pyenv` as one of the prompt segments.
3. Type these commands:

```zsh
pyenv install 3.7.3
export PYENV_VERSION=python-3.7.3
```

Expected behavior: pyenv segment shows `3.7.3` (the same as `pyenv version-name`).

Actual behavior: pyenv segment shows `python-3.7.3`.

Syphdias:
`rbenv` looks very similar to `pyenv`, I suppose it suffers the same issue...

laggardkernel:
It's because `PYENV_VERSION` is detected firstly and used by powerlevel9k. And powerlevel9k just reflects what it get from `PYENV_VERSION`.

Besides, this may not be a problem at all, cause you should try `export PYENV_VERSION=3.7.3`. From output `pyenv versions`, there's no prefix for python version like `3.7.3`.

https://github.com/bhilburn/powerlevel9k/blob/3057e8fc07374e9c580790a023cb85998ca91fc8/powerlevel9k.zsh-theme#L1645-L1659

@Syphdias Your simplification slows down the virtual environment detection dramatically. Get python version from `PYENV_VERSION` is much **faster** than `pyenv version-name`

```zsh
❯ time (pyenv version-name)
2.7.16
0.05s user 0.04s system 96% cpu 0.097 total

❯ time (echo $PYENV_VERSION)
2.7.16
0.00s user 0.00s system 59% cpu 0.001 total
```

`pyenv version-name` is just an implementation of [the version detection logic](https://github.com/pyenv/pyenv#choosing-the-python-version):
1. `PYENV_VERSION`
2. `.python-version`
3. `${PYENV_ROOT}/version`

There's some overhead of `pyenv version-name` compared with checking `PYENV_VERSION` directly.

laggardkernel:
Reading from the [manual of pyenv](https://github.com/pyenv/pyenv/blob/master/COMMANDS.md#pyenv-shell-advanced), version name without `python-` prefix is the canonical way to use it.

`PYENV_VERSION` is exported by `pyenv shell <venv-name>` and `pyenv activate <venv-name>`, you'd better not touch it directly.

There's some usage examples in the test.

[...]

PR #1283, opened 2019-05-22
[Bugfix] pyenv weirdness

### LINKED ISSUE #1280: pyenv doesn't respect POWERLEVEL9K_PYENV_PROMPT_ALWAYS_SHOW=false
This is on `master`.

To reproduce:

1. Install pyenv.
2. Use `master` branch of Powerlevel9k with `pyenv` as one of the prompt segments.
3. Set `POWERLEVEL9K_PYENV_PROMPT_ALWAYS_SHOW=false` (this is also the default).
4. Type these commands:

```zsh
pyenv global system
pyenv shell system
```

Expected behavior: pyenv segment not shown because current pyenv version is equal to the global pyenv version (both are `system`).

Actual behavior: pyenv segment shows `system`.

Syphdias:
I was a bit confused to find `__p9k_$1_prompt_segment` here...

This produced the wrong segment content, e.g. `systemblack`. Also resolves #1280 by not displaying the version if it is "system" and `P9K_PYENV_PROMPT_ALWAYS_SHOW=false`

### DIFF
--- a/segments/pyenv/pyenv.p9k
+++ b/segments/pyenv/pyenv.p9k
@@ -32,8 +32,8 @@
 #   [Choosing the Python Version](https://github.com/pyenv/pyenv#choosing-the-python-version)
 ##
 prompt_pyenv() {
-  if [[ -n "$PYENV_VERSION" ]]; then
-    "__p9k_$1_prompt_segment" "$0" "$2" "blue" "$DEFAULT_COLOR" "$PYENV_VERSION" 'PYTHON_ICON'
+  if [[ -n "$PYENV_VERSION" && ("$PYENV_VERSION" != "system" || "${P9K_PYENV_PROMPT_ALWAYS_SHOW}" == "true") ]]; then
+    p9k::prepare_segment "$0" "" $1 "$2" $3 "$PYENV_VERSION"
   elif [ $commands[pyenv] ]; then
     local pyenv_version_name="$(pyenv version-name)"
     local pyenv_global="system"
--- a/segments/pyenv/pyenv.p9k
+++ b/segments/pyenv/pyenv.p9k
@@ -32,15 +32,12 @@
 #   [Choosing the Python Version](https://github.com/pyenv/pyenv#choosing-the-python-version)
 ##
 prompt_pyenv() {
-  if [[ -n "$PYENV_VERSION" && ("$PYENV_VERSION" != "system" || "${P9K_PYENV_PROMPT_ALWAYS_SHOW}" == "true") ]]; then
+  local pyenv_root="$(pyenv root)"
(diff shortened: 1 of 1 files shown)


Syphdias:
Question: Why not use `pyenv global`?

laggardkernel:
@Syphdias Your simplification slows down the virtual environment detection dramatically. Get python version from `PYENV_VERSION` is much **faster** than `pyenv version-name`. The similar thing happens with `rbenv` and `nodenv`.

```zsh
❯ time (pyenv version-name)
2.7.16
0.05s user 0.04s system 96% cpu 0.097 total

❯ time (echo $PYENV_VERSION)
2.7.16
0.00s user 0.00s system 59% cpu 0.001 total
```

`pyenv version-name` is just an implementation of [the version detection logic](https://github.com/pyenv/pyenv#choosing-the-python-version):
1. `PYENV_VERSION`
2. `.python-version`
3. `${PYENV_ROOT}/version`

There's some overhead of `pyenv version-name` compared with checking `PYENV_VERSION` directly.

Syphdias:
> There's some overhead of `pyenv version-name` compared with checking `PYENV_VERSION` directly.

The expected behaviour is, that the segment spits out the same output as the command to check the version. Otherwise, false information is provided by the segment that might confuse a user. To get exactly that behaviour you need every logic handled by `pyenv version-name`. How do we get this exact behaviour now?

1. Use the (slow) command itself
2. reimplement every logic, including future changes

If you choose 2 you can improve performance if there is unnecessary overhead.
It's not true that performance improvements cannot be done with 1. You could utilize techniques like caching or prefetching to get the output faster.

The old implementation did not reimplement all the logic that is used by `pyenv version-name` thus proving false output. That's why I decided to remove the shortcut. Of course, you could go with the current version as a compromise of speed and wrong output.

laggardkernel:
Obviously, I'm not that insane to implement every logic again by myself. I'm only indicating you should check `PYENV_VERSION` first, then fallback to `pyenv version-name`.

Syphdias:
> you should check `PYENV_VERSION` first

As mentioned in #1281 `PYENV_VERSION` cannot be trusted. And it is a compromise. There are pros and cons to trusting `PYENV_VERSION` anyways.
I'm not a user. I cannot decide which one is the better version. The current version just has less unexpected behaviour.

Syphdias:
Closing this PR since #1285 makes this one redundant.

@laggardkernel, I'll leave it to @bhilburn which one he thinks is best for users, since the current version does have its drawbacks as well.

laggardkernel:
Sorry to bother you. I didn't realize there was another related pr being merged. It seems my words is too late.

PR #1284, opened 2019-05-22
[Bugfix] rbenv doesn't respect POWERLEVEL9K_RBENV_PROMPT_ALWAYS_SHOW=false

### LINKED ISSUE #1282: rbenv doesn't respect POWERLEVEL9K_RBENV_PROMPT_ALWAYS_SHOW=false
This is on `master`.

To reproduce:

1. Install rbenv.
2. Use `master` branch of Powerlevel9k with `rbenv` as one of the prompt segments.
3. Set `POWERLEVEL9K_RBENV_PROMPT_ALWAYS_SHOW=false` (this is also the default).
4. Type these commands:

```zsh
rbenv install 2.6.3
rbenv global 2.6.3
rbenv shell 2.6.3
```

Expected behavior: rbenv segment not shown because the current rbenv version is equal to the global rbenv version (both are `2.6.3`).

Actual behavior: rbenv segment shows `2.6.3`.

Syphdias:
Similar to #1283

resolves #1282

### DIFF
--- a/segments/rbenv/rbenv.p9k
+++ b/segments/rbenv/rbenv.p9k
@@ -29,11 +29,11 @@
 #   $3 boolean Whether the segment should be joined
 ##
 prompt_rbenv() {
-  if [[ -n "${RBENV_VERSION}" ]]; then
+  local rbenv_global="$(rbenv global)"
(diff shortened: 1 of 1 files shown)


bhilburn:
It's funny - the Ruby segments were some of the first that p9k ever had, and aside from VCS segments, I think they require the most maintenance, hah.

Thanks so much, @Syphdias :)

PR #1285, opened 2019-05-22
[Bugfix] Simplifing pyenv and rbenv to get rid of unexpected behaviour

Syphdias:
I removed some of the complexity and manual checking for versions. `pyenv` and `rbenv` should:tm: take care of that themselves. This seems to me like the most straight forward solution to not introducing unexpected behaviour. This would also close 3 issues (and 2 PRs I opened without simplifying code).

I did only very little testing since I know very little about `pyenv` and `rbenv` and these segments have no unit test.

### DIFF
--- a/segments/pyenv/pyenv.p9k
+++ b/segments/pyenv/pyenv.p9k
@@ -33,7 +33,7 @@
 ##
 prompt_pyenv() {
   if [ $commands[pyenv] ]; then
-    local pyenv_version_name="$(pyenv version-name)"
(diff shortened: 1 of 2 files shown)


bhilburn:
Ugh, still waiting for Travis to finish >:(

bhilburn:
FINALLY! Tests complete =)

This branch had a conflict, which I resolved.

I really like this PR. In addition to fixing those bugs, this is dramatically simpler code. Thanks, @Syphdias, and I think you should be good to close those other bugs! =)

PR #1286, opened 2019-05-23
[Bugfix] Removes whitespace from virtualenv icon if Awesome-Mapped-FontConfig is uses (for consistency)

Syphdias:
see title


bhilburn:
Hah, yes - thanks.

Consistency in our spacing has never been a strong point. Good to fix these =)

PR #1319, opened 2019-06-27
[Bugfix] fix #1013 TERM="xterm-256color" not recognized anymore on FreeBSD

### LINKED ISSUE #1013: TERM="xterm-256color" not recognized anymore on FreeBSD
#### Describe Your Issue
What is happening?

After updating the theme from 0e3c192 to b16c0d5 I get the warning message that my terminal does not support 256 colors and that I should set `export TERM="xterm-256color" ` at the top of my `~/.zshrc`.
I did also update to b16c0d5 on my nearly identical Ubuntu (WSL), macOS and Termux setups and they don't show this message. Is happens as soon as I ssh into my FreeBSD server and the used terminal emulator does not seem to matter here.

I already added `export TERM="xterm-256color" ` to my `~/.zshrc` and it is set according to `echo $TERM`. Even running something like `fortune | lolcat` shows that 256 colors are working. The presence of the warning message seems to be an error. 🤔

#### Environment Information
This information will help us understand your configuration. 

  - What version of ZSH are you using? You can use `zsh --version` to see this.
    - zsh 5.6.2 (amd64-portbld-freebsd11.2)
  - Do you use a ZSH framework (e.g., Oh-My-ZSH, Antigen)?
    - Oh My Zsh 
  - How did you install P9k (cloning the repo, by tarball, a package from your OS, etc.,)?
    - cloning the repo into  ~/.oh-my-zsh/custom/themes
  - What version of P9k are you using?
    - b16c0d5 
  - Which terminal emulator do you use?
    - iTerm2, mintty, Termux

thawk:
#### Description

In FreeBSD (My version is ``FreeBSD 11.2``), ``tput colors`` is the same as ``tput co``, it returns the number of columns instead of the number of colors.  ``tput Co`` returns the number of colors.

But in other OS, like my MacBook, ``tput Co`` failed to be executed.  So some kind OS based logic has to be introduced to use difference terminfo capability at difference OS.

This PR use ``tput Co`` for and only for FreeBSD, use original ``tput colors`` otherwise.

### DIFF
--- a/functions/colors.zsh
+++ b/functions/colors.zsh
@@ -278,7 +278,11 @@ function termColors() {
   local term_colors
 
   if which tput &>/dev/null; then
-	term_colors=$(tput colors)
+    if [[ "$OSTYPE" == "freebsd"* ]]; then
+      term_colors=$(tput Co)
+    else
+      term_colors=$(tput colors)
+    fi
   else
 	term_colors=$(echotc Co)
   fi


romkatv:
I've fixed this bug in Powerlevel10k in a portable way: https://github.com/romkatv/powerlevel10k/blob/b404c1b54e5c0e0ce978e4839c828171178ee824/internal/p10k.zsh#L2673. Feel free to backport the fix to Powerlevel9k.

(I usually open an issue against Powerlevel9k for every bug that I fix in Powerlevel10k but this time apparently I forgot to do so.)

thawk:
@romkatv has a better, OS independent solution.
Thx.

PR #1329, opened 2019-07-08
[Bugfix] Escape control characters

dritter:
This is a backport of 5446c1f51e01af8d73e93f4fa6b7b9ba1df7ce8f . I removed the `__p9k_unsafe_PROMPT`, as this is not necessary any more and Romans solution is more flexible. This fixes #1310

Issue #1306, opened 2019-06-11
Right prompt overlaps left prompt


romkatv:
To reproduce:

```zsh
docker run -e LANG=C.UTF-8 -e LC_ALL=C.UTF-8 -e TERM=$TERM -it --rm ubuntu bash -uexc '
  cd
  apt update && apt install -y zsh git
  git clone https://github.com/bhilburn/powerlevel9k.git
  echo "
    POWERLEVEL9K_PROMPT_ON_NEWLINE=true
    POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(custom_left)
    POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(custom_right)
    typeset POWERLEVEL9K_CUSTOM_{LEFT,RIGHT}=\"echo \${(pl.\$((COLUMNS/2))..-.)}\"
    POWERLEVEL9K_CUSTOM_LEFT_BACKGROUND=red
    POWERLEVEL9K_CUSTOM_RIGHT_BACKGROUND=blue
    source ~/powerlevel9k/powerlevel9k.zsh-theme" >~/.zshrc
  exec zsh'
```

Actual:

![actual](https://user-images.githubusercontent.com/1282067/59246190-e249be00-8c1b-11e9-8b83-0005339a3739.png)

Expected:

![expected](https://user-images.githubusercontent.com/1282067/59246197-e70e7200-8c1b-11e9-9f16-0560cb63c7d6.png)

Normally, ZSH doesn't display right prompt if it would overlap left prompt. If you type the following command, you won't see right prompt.

```zsh
PROMPT="${(pl.$((COLUMNS/2))..-.)}> " RPROMPT="<${(pl.$((COLUMNS/2))..-.)}" zsh -df
```

Now, if you make your terminal wider, right prompt will appear.

If you remove `POWERLEVEL9K_PROMPT_ON_NEWLINE=true` from the p9k config above, right prompt won't be displayed, as expected.

```zsh
docker run -e LANG=C.UTF-8 -e LC_ALL=C.UTF-8 -e TERM=$TERM -it --rm ubuntu bash -uexc '
  cd
  apt update && apt install -y zsh git
  git clone https://github.com/bhilburn/powerlevel9k.git
  echo "
    POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(custom_left)
    POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(custom_right)
[...]

Issue #1291, opened 2019-05-27
TRANSIENT_RPROMPT is not respected


romkatv:
To reproduce:

```zsh
docker run -e LANG=C.UTF-8 -e LC_ALL=C.UTF-8 -e TERM=$TERM -it --rm ubuntu bash -uexc '
  cd
  apt update && apt install -y zsh git
  git clone https://github.com/bhilburn/powerlevel9k.git
  echo "
    setopt TRANSIENT_RPROMPT
    POWERLEVEL9K_PROMPT_ON_NEWLINE=true
    source ~/powerlevel9k/powerlevel9k.zsh-theme" >~/.zshrc
  exec zsh'
```

Then press `<enter>` a few times.

Expected: right prompt is displayed only for the last (active) prompt.
Actual: right prompt is displayed even for the past prompts.

Note: Adding `POWERLEVEL9K_RPROMPT_ON_NEWLINE=true` to the config fixes the issue.

dritter:
Fixed on `next` with #1335

PR #1335, opened 2019-07-15
[Bugfix] Fix completion menu

### LINKED ISSUE #1267: POWERLEVEL9K_RPROMPT_ON_NEWLINE=false breaks completion menu

dritter:
This PR is a backport from P10K and fixes several problems around multi line prompts.

Fixes #1267 , #1291 and #1306 .

Test are not green right now, because I am waiting for #1329 to be merged..

### DIFF
--- a/generator/default.p9k
+++ b/generator/default.p9k
@@ -393,32 +393,6 @@ function __p9k_prepare_prompts() {
 
   local LC_ALL="" LC_CTYPE="en_US.UTF-8" # Set the right locale to protect special characters
 
-  if [[ "${P9K_PROMPT_ON_NEWLINE:-}" == true ]]; then
-    __p9k_unsafe_PROMPT="${__P9K_ICONS[MULTILINE_FIRST_PROMPT_PREFIX]}%f%b%k$(__p9k_build_left_prompt)
-${__P9K_ICONS[MULTILINE_LAST_PROMPT_PREFIX]}"
-    if [[ "$P9K_RPROMPT_ON_NEWLINE" != true ]]; then
-      # The right prompt should be on the same line as the first line of the left
-      # prompt. To do so, there is just a quite ugly workaround: Before zsh draws
-      # the RPROMPT, we advise it, to go one line up. At the end of RPROMPT, we
-      # advise it to go one line down. See:
-      # http://superuser.com/questions/357107/zsh-right-justify-in-ps1
-      RPROMPT_PREFIX='%{'$'\e[1A'${P9K_RPROMPT_PREFIX:-}'%}' # one line up
-      RPROMPT_SUFFIX='%{'${P9K_RPROMPT_SUFFIX:-}$'\e[1B''%}' # one line down
-    else
-      RPROMPT_PREFIX='%{'${P9K_RPROMPT_PREFIX:-}'%}'
-      RPROMPT_SUFFIX='%{'${P9K_RPROMPT_SUFFIX:-}'%}'
-    fi
-  else
-    __p9k_unsafe_PROMPT="%f%b%k$(__p9k_build_left_prompt)"
-    RPROMPT_PREFIX='%{'${P9K_RPROMPT_PREFIX:-}'%}'
-    RPROMPT_SUFFIX='%{'${P9K_RPROMPT_SUFFIX:-}'%}'
-  fi
-
-  if [[ "${P9K_DISABLE_RPROMPT:-}" != true ]]; then
-    __p9k_unsafe_RPROMPT="${RPROMPT_PREFIX}%f%b%k$(__p9k_build_right_prompt)%{${reset_color}%}${RPROMPT_SUFFIX}"
-    RPROMPT='$__p9k_unsafe_RPROMPT'
-  fi
-
   # Allow iTerm integration to work
   [[ "${ITERM_SHELL_INTEGRATION_INSTALLED:-}" == "Yes" ]] \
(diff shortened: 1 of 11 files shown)


romkatv:
Are you sure it fixes #1267, #1291 and #1306? Can you reproduce these bugs without this PR and then verify that bugs no longer trigger with the PR?

Syphdias:
> Are you sure it fixes #1267, #1291 and #1306? Can you reproduce these bugs without this PR and then verify that bugs no longer trigger with the PR?

| issue | broken on `next` | fixed on `fix-completion-menu` |
| --- | --- | --- |
| #1267 | yes | yes |
| #1291 | yes | yes |
| #1306 | yes | kinda |

image

romkatv:
> > Are you sure it fixes #1267, #1291 and #1306? Can you reproduce these bugs without this PR and then verify that bugs no longer trigger with the PR?
> 
> issue	broken on `next`	fixed on `fix-completion-menu`
> #1267	yes	yes
> #1291	yes	yes
> #1306	yes	kinda
> image

I may not be reading the code right but doesn't it simply remove support for `P9K_RPROMPT_ON_NEWLINE=false`? If this is the goal, there is no need to look at `transient_prompt` and to hook zle widgets -- zsh will do the right thing automatically.

dritter:
> I may not be reading the code right but doesn't it simply remove support for P9K_RPROMPT_ON_NEWLINE=false?

D0h! Fixed three Issues and brought back another one.. 🙄 Checking for a fix. Thanks for the hint!

test/powerlevel9k.spec
```diff
@@ -43,9 +43,14 @@ function oneTimeTearDown() {
...
 
 function testUsingUnsetVariables() {
-  setopt local_options
-  set -u
-  __p9k_prepare_prompts
+  startSkipping # Skip test
+
+  # setopt local_options
+  # set -u
+  # __p9k_prepare_prompts
```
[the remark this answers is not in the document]
dritter: Yes, at some point. AFAIR, the old code succeeded at this point, because `__p9k_prepare_prompts` set the `unsafe` variables, that were not evaluated at that time.. Fixing this test would require add a default to every unset variable (which is probably a good thing).

romkatv:
Moving in the right direction :+1: 

You might want to link to https://stackoverflow.com/a/57141646/1095235.

generator/default.p9k
```diff
@@ -546,11 +546,7 @@ function prompt_powerlevel9k_setup() {
           _P9K_RPROMPT_OVERRIDE=
           zle && zle .reset-prompt && zle -R
         }
-        autoload -Uz add-zle-hook-widget
-        zle -N _p9k_zle_line_finish
-        add-zle-hook-widget line-finish _p9k_zle_line_finish
```
romkatv: By the way, the reason for this change is that `add-zle-hook-widget` can lead to infinite loops if there is at least one plugin that wraps zle widgets the old school way instead of using `add-zle-hook-widget`. Until `add-zle-hook-widget` is fixed, it's must be avoided by everyone as a single user can break everything. The fix isn't difficult but the benefits are also low because wrapping widgets works fine. So the summary is that `add-zle-hook-widget` should be treated as if it doesn't exist and all users should be told to stop using it.

dritter: Good to know. Thanks for the insights. 👍

dritter:
Btw. This is still not finished, as transient_rprompt with multiline prompt eats up one line.. Still investigating this.

romkatv:
> Btw. This is still not finished, as transient_rprompt with multiline prompt eats up one line.. Still investigating this.

Does the same thing work in p10k? If it doesn't, I'll join in debugging.

dritter:
Nah. It works in p10k. It's just a bit hard to read the generated code.. ;)

romkatv:
> Nah. It works in p10k. It's just a bit hard to read the generated code.. ;)

Tell me about it :grin:

It's also a good idea to check that `reset-prompt` doesn't eat (or create) a line. Hit `esc-x` and type `reset-prompt`. Many kinds of bugs can do this sort of badness.

Another test is to hit `ctrl-r` and to verify that your cursor doesn't move. Same thing -- it's easy to break this.

dritter:
Thanks for the tips. I'll double check that, once I am this far..

dritter:
I think this is now done. @romkatv The checks you suggested pass on my machine.

Additionally this resolves issues with `ZLE_RPROMPT_INDENT`, which we handle as 1 when set to zero.

@Syphdias I even reactivated `testUsingUnsetVariables`. :) It would be super nice if you could retest this.

PR #1338, opened 2019-07-16
[Bugfix] Make icons appear properly.

ZeGentzy:
This was originally only happening to the blue bar, but when I rebased this commit against the most recent commit on `next` it had actually regressed!

Anyways, pictures are worth a thousand words.

```
P9K_MULTILINE_LAST_PROMPT_PREFIX_ICON="%K{004}%F{000} %D{\uF017 %H:%M \uF073 %d.%m.%y} %f%k%F{004} %f"
P9K_TIME_FORMAT="%D{\uF017 %H:%M \uF073 %d/%m/%y}"
```

`next` from yesterday:

`next` from today:

`next` with this PR:

Edit: Note: I normally only have the blue one, not the white one. I just added it for reference.

### DIFF
--- a/generator/default.p9k
+++ b/generator/default.p9k
@@ -450,6 +450,9 @@ local NEWLINE='
     repeat ${P9K_PROMPT_ADD_NEWLINE_COUNT:-1} { NEWLINES+=${NEWLINE} }
     PROMPT="${NEWLINES}${PROMPT}"
   fi
+
+  # Make icons appear properly
+  PROMPT="$(echo -n "${PROMPT}")"
 }
 
 p9k::set_default P9K_IGNORE_TERM_COLORS false


ZeGentzy:
cc @dritter

dritter:
Hmm. The workaround for this would be to use the decomposed UTF-8 Chars directly.. This would avoid the subshell:
`P9K_TIME_FORMAT="%D{ %H:%M  %d/%m/%y}"`

Does this work for you?
And sorry for the inconvenience.

ZeGentzy:
Good enough for me. I think a lot of https://github.com/bhilburn/powerlevel9k/wiki/Show-Off-Your-Config will need updating, as this was previously functional on `master`.

dritter:
Yeah. There is maybe a better solution if we try to decompose chars on initialization. I'll think about this in the next days. Thanks for letting me know.

PR #1341, opened 2019-07-22
[Bugfix] Fix decomposing utf8

dritter:
This was originally discovered by @ZeGentzy in #1338 .

Here we now decompose UTF-8 chars that might be set by the user.

@ZeGentzy could you confirm this works for you?

### DIFF
--- a/functions/utilities.zsh
+++ b/functions/utilities.zsh
@@ -236,6 +236,32 @@ function p9k::set_default() {
 
 ###############################################################
 # @description
+#   This function expands the given parameter. This has to be
+#   done, as some people write P9K_DIR_PATH_SEPARATOR='\uNNNN'
+#   instead of P9K_DIR_PATH_SEPARATOR=$'\uNNNN'.
+##
+# @args
+#   $1 string The value to be expanded
+##
+# @returns
+#   Nothing.
+##
+function p9k::expand() {
+  (( $+parameters[$1] )) || return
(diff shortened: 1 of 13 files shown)


ZeGentzy:
It does not.

```
	P9K_MULTILINE_LAST_PROMPT_PREFIX_ICON="%K{004}%F{000} %D{\uF017 \uF073  %H:%M  %d.%m.%y} %f%k%F{004} %f"
	P9K_TIME_FORMAT="%D{\uF017 \uF073  %H:%M  %d.%m.%y}"
```

dritter:
Ha! Could you try again? I forgot to expand user icons..

romkatv:
These are very important fixes 👍

functions/utilities.zsh
```diff
@@ -203,20 +203,59 @@ function p9k::defined() {
...
+#   Nothing.
+##
+function p9k::expand() {
+  (( $+parameters[$1] )) || return
+  local -a ts=("${=$(typeset -p $1)}")
+  shift ts
+  local x
+  for x in "${ts[@]}"; do
+    [[ $x == -* ]] || break
+    # Don't change readonly variables. Ideally, we shouldn't modify any variables at all,
```
romkatv: Sorry, I've missed this when I skimmed the changes the first time. I don't recommend copying this hack. It has bad failure mode when it doesn't work and some unsavory side effects when it does.

PR #1347, opened 2019-07-31
 [Bugfix]  Pid fix

ZeGentzy:
When started in the tty, powerline9k tries to open the stat file of the
`login` process. Because the `login` process belongs to root, it's
existence is hidden from non-privileged users. This will result in an
error like
`__p9k_detect_terminal:27: no such file or directory: /proc/3434/stat`.

Like #1336, this error appears to be ultimately harmless, but never the less annoying.


ZeGentzy:
ping @dritter

PR #1192, opened 2019-03-13
[New Segment]: Microsoft Azure segment

sambadevi:
This segment adds your currently active Microsoft Azure subscription to your prompt

Currently the Pictures in `.svg`and `.png` format are missing. Could someone please tell me how to generate them?

I'll add a screenshot of the segment here for the time being:
image

PR #1396, opened 2020-02-13
[New Segment]: Microsoft Azure Subscription segment (async)

Iridias:
This segment adds your currently active Microsoft Azure subscription to your prompt.

Fetching is async, as the command can take ~ 2 seconds.


sambadevi:
Duplicate, or rather enhancement of #1192 (which was not async)

sambadevi:
Would you please add an entry to README.md under Segments? We collect available segments in a list

